PR #52, opened 2020-08-29
[WIP] Sdl packages

### LINKED ISSUE #51: Add all SDL libraries
### Describe the solution you'd like
I would like to add the SDL libraries to the repo.
+ SDL_image
+ SDL_ttf
+ SDL_mixer
+ SDL_net
+ SDL_gfx 

### Describe alternatives you've considered
I am currently working on a pull request

PucklaMotzer09:
Closes #51 
This pull request adds all SDL libraries to the repo.
+ SDL_image
+ SDL_ttf
+ SDL_mixer
+ SDL_net
+ SDL_gfx

Currently **image**, **mixer** and **net** work on both Windows and Linux. I cannot test MacOSX, but since the install process is the same as of Linux, I think that it works too. **ttf** only works on Windows, since Linux needs **freetype** to compile it. **gfx** has to be compiled for both operating systems, so this is more complicated. **gfx** needs **SDL2** for the compilation, so currently this library doesn't work at all.

To make the compilation of **ttf** and **gfx** work, I need to somehow tell **autoconf** where to find the required libraries. Is there a way to do this? Is there an API for getting the include and lib directories of another package?

I tried setting **freetype** as a dependency of **ttf**, but **freetype** fails to compile on linux. But even if I get **freetype** to work, I don't think that **autoconf** will find the required includes and libs, or does it?


PucklaMotzer09:
I found the **sdl2-config** file inside the package directory of **libsdl**. To make **gfx** compile I just need to set the **SDL_CONFIG** environment variable to point to this path. Is there a way in the API to get this path? For me it's *~/.xmake/packages/l/libsdl/2.0.8/76b7c3070d364167afbd2190b4d3d703/bin*

waruqi:
Can you split into a pull request corresponding to a package? This makes it easy to test the installation of each package on ci. It is also convenient for review and merge.

waruqi:
> To make the compilation of ttf and gfx work, I need to somehow tell autoconf where to find the required libraries. Is there a way to do this? Is there an API for getting the include and lib directories of another package?

You can use add_deps() to add deps. autoconf  will detect their pkg-config infos automatically.

PucklaMotzer09:
> You can use add_deps() to add deps. autoconf will detect their pkg-config infos automatically.

I am already using add_deps to add **libsdl** as a dependency:

```lua
package("libsdl_gfx")
    add_deps("libsdl")
```
xmake detects the dependency, but autoconf seems to ignore it. Should I call it somewhere else?

packages/l/libsdl_gfx/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+        if package:config("shared") then
+            table.insert(configs, "--enable-shared=yes")
+        else
+            table.insert(configs, "--enable-shared=no")
+        end
+        import("package.tools.autoconf").install(package, configs)
+        local file_name = path.join(package:installdir("include"), "SDL2", "SDL2_framerate.h")
+        local f = io.open(file_name)
+        local content = f:read("*all")
+        f:close()
```
waruqi, line 25: You can write

local content = io.readfile(file_name)

packages/l/libsdl_gfx/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+        local file_name = path.join(package:installdir("include"), "SDL2", "SDL2_gfxPrimitives.h")
+        local f = io.open(file_name)
+        local content = f:read("*all")
+        f:close()
+
+        content = string.gsub(content, "\"SDL.h\"", "<SDL2/SDL.h>")
+
+        local f = io.open(file_name, "w")
+        f:write(content)
+        f:close()
```
waruqi, line 42: io.writefile(file_name, content)

packages/l/libsdl_gfx/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+
+        local file_name = path.join(package:installdir("include"), "SDL2", "SDL2_gfxPrimitives.h")
+        local f = io.open(file_name)
+        local content = f:read("*all")
+        f:close()
+
+        content = string.gsub(content, "\"SDL.h\"", "<SDL2/SDL.h>")
```
waruqi, line 38: content = content:gsub()

waruqi: or 

use io.gsub(file_name, "", "") to replace content of file.

waruqi:
> I found the sdl2-config file inside the package directory of libsdl. To make gfx compile I just need to set the SDL_CONFIG environment variable to point to this path. Is there a way in the API to get this path? For me it's ~/.xmake/packages/l/libsdl/2.0.8/76b7c3070d364167afbd2190b4d3d703/bin

package:installdir("bin")

package:dep("libsdl"):installdir("bin")

PucklaMotzer09:
I tried to get the installdir of **libsdl** and give it to **autoconf** via ```--with-sdl-prefx```, but **find_package** doesn't find **libsdl** or any other package when called inside **on_install**.

PucklaMotzer09:
> Can you split into a pull request corresponding to a package? This makes it easy to test the installation of each package on ci. It is also convenient for review and merge.

I will close this pull request and add seperate pull requests for every package once I am done.

waruqi:
You can manually execute ./configure first, and try to add libsdl's pkgconfig path or sdlconfig to autoconf. If libsdl can be detected successfully, we can go to on_install to try it.

In fact, by default, xmake will already pass the pkgconfig path of libsdl to autoconf

https://github.com/xmake-io/xmake/blob/f4deeedb389e6d5f7ed3cd6594be27da67d8b19b/xmake/modules/package/tools/autoconf.lua#L128

packages/l/libsdl_gfx/xmake.lua
```diff
@@ -0,0 +1,45 @@
...
+        content = content:gsub("\"SDL.h\"", "<SDL2/SDL.h>")
+
+        io.writefile(file_name, content)
+
+        local file_name = path.join(package:installdir("include"), "SDL2", "SDL2_rotozoom.h")
+        local content = io.readfile(file_name)
+
+        content = content:gsub("\"SDL.h\"", "<SDL2/SDL.h>")
+
+        io.writefile(file_name, content)
```
waruqi, line 44: If you just do only once file content replacement, you can use the following code.

```lua
local sdl2_includedir = path.join(package:installdir("include"), "SDL2")
io.gsub(path.join(sdl2_includedir, "SDL2_gfxPrimitives.h"), ""\"SDL.h\"", "<SDL2/SDL.h>"")
io.gsub(path.join(sdl2_includedir, "SDL2_rotozoom.h"), ""\"SDL.h\"", "<SDL2/SDL.h>"")
```

waruqi, line 44: If you just do only once file content replacement, you can use the following code.

```lua
local sdl2_includedir = path.join(package:installdir("include"), "SDL2")
io.gsub(path.join(sdl2_includedir, "SDL2_gfxPrimitives.h"), ""\"SDL.h\"", "<SDL2/SDL.h>"")
io.gsub(path.join(sdl2_includedir, "SDL2_rotozoom.h"), ""\"SDL.h\"", "<SDL2/SDL.h>"")
```

waruqi: In addition, why do you want to replace the include path inside? You can call `add_includedirs` to add extra include directories.

```lua
package()
    on_load(function (package)
        package:add("includedirs", package:installdir("include"))
        package:add("includedirs", package:dep("libsdl"):installdir("include", "SDL2")) -- add extra include directories
    end)
```

PucklaMotzer09: > In addition, why do you want to replace the include path inside? You can call add_includedirs to add extra include directories.

Yes, you are right. I haven't thought about that.

waruqi:
I have added libsdl_gfx to the dev branch, https://github.com/xmake-io/xmake-repo/commit/af94d136dedad49ed35a70e7474cda0b075b8d4e

You can refer to it and add other libraries.

PucklaMotzer09:
Now I have added support for windows in **libsdl_gfx** using **msbuild**. But currently it just works for **x86**, since there is no **x64** configuration in the Visual Studio project file.

PucklaMotzer09:
> In addition, why do you want to replace the include path inside? You can call add_includedirs to add extra include directories.

I couldn't get this to work, because calling
```lua
package:add("includedirs", package:dep("libsdl"):installdir("include", "SDL2"))
```
just combines the directory I specify in the second argument with the ```installdir``` of ```package```, which results in paths that do not exist.

waruqi:
> I couldn't get this to work, because calling

I have added includedirs to libsdl. see https://github.com/xmake-io/xmake-repo/commit/af94d136dedad49ed35a70e7474cda0b075b8d4e

PucklaMotzer09:
> I have added includedirs to libsdl. see af94d13

Now I saw that. I removed the code that replaces the includes.

waruqi:
I think you should modify libsdl_gfx and submit pr to the dev branch first, and then try other packages.

PucklaMotzer09:
> I think you should modify libsdl_gfx and submit pr to the dev branch first, and then try other packages.

I will split up all packages into seperate PRs once I am done.

PucklaMotzer09:
I will now split up the PR.

PR #58, opened 2020-09-01
Add assimp

PucklaMotzer09:
Adds the assimp library.

### DIFF
--- a/packages/a/assimp/xmake.lua
+++ b/packages/a/assimp/xmake.lua
@@ -1,0 +1,101 @@
+package("assimp")
+
+    set_homepage("https://assimp.org")
+    set_description("Portable Open-Source library to import various well-known 3D model formats in a uniform manner")
+
+    if is_plat("windows", "linux", "macosx") then
+        set_urls("https://github.com/assimp/assimp/archive/v$(version).zip")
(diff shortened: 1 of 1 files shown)


packages/a/assimp/xmake.lua
```diff
@@ -0,0 +1,98 @@
...
+            add_config_arg("build_assimp_tools", "ASSIMP_BUILD_ASSIMP_TOOLS")
+        end
+
+        if is_plat("android", "iphoneos") then
+            table.insert(config, "-DASSIMP_BUILD_ASSIMP_TOOLS=OFF")
+        end
+
+        import("package.tools.cmake").install(package, config)
+        if is_plat("linux") then
+            package:add("links", "assimp", "IrrXML")
```
waruqi, line 95: you can uses add_syslinks("irrXML") to ensure link order.

waruqi (commented):
Are you sure all linux systems have built-in irrXML library?

waruqi:
In order to ensure that xmake-repo does not have a long commit history, please merge all the commits into single commit.

Thanks.

packages/a/assimp/xmake.lua
PucklaMotzer09: The irrXML library is built into assimp and gets compiled with it when ```system_irrxml``` is false. Which is the default behaviour.

waruqi:
I have merged into the master branch.

PR #59, opened 2020-09-01
Add irrXML

PucklaMotzer09:
Adds the irrXML library which is necessary for assimp (#58)

### DIFF
--- a/packages/i/irrXML/xmake.lua
+++ b/packages/i/irrXML/xmake.lua
@@ -1,0 +1,23 @@
+package("irrxml")
+
+    set_homepage("https://sourceforge.net/projects/irrlicht/")
+    set_description("High speed and easy-to-use XML Parser for C++")
+
+    set_urls("https://deac-fra.dl.sourceforge.net/project/irrlicht/irrXML%20SDK/$(version)/irrxml-$(version).zip")
+    add_versions("1.2", "9b4f80639b2dee3caddbf75862389de684747df27bea7d25f96c7330606d7079")
+
+    on_install(function (package)
+        io.writefile("xmake.lua", [[
+        add_rules("mode.debug", "mode.release")
+        target("irrxml")
+            set_kind("static")
+            add_headerfiles("src/*.h")
+            add_files("src/*.cpp")
+        ]])
+
+        import("package.tools.xmake").install(package)
+    end)
+
+    on_test(function (package)
+        assert(package:has_cxxfuncs("irr::io::createIrrXMLReader(\"example.xml\")", {includes = "irrXML.h"}))
+    end)


packages/i/irrXML/patches/1.2/add-xmake.patch
```diff
@@ -0,0 +1,37 @@
...
++    else
++        set_symbols("none")
++        set_strip("all")
++        set_optimize("fastest")
++    end
++
++    add_headerfiles("src/*.h")
++    add_files("src/*.cpp")
+\ No newline at end of file
+--
```
waruqi, line 35: You need only write:

```lua
target("irrXML")
    set_kind("static")
    add_headerfiles("src/*.h")
    add_files("src/*.cpp")
```

packages/i/irrXML/xmake.lua
```diff
@@ -0,0 +1,16 @@
+package("irrXML")
+
+    set_homepage("https://sourceforge.net/projects/irrlicht/")
+    set_description("High speed and easy-to-use XML Parser for C++")
+
+    set_urls("https://deac-fra.dl.sourceforge.net/project/irrlicht/irrXML%20SDK/$(version)/irrxml-$(version).zip")
+    add_versions("1.2", "9b4f80639b2dee3caddbf75862389de684747df27bea7d25f96c7330606d7079")
+    add_patches("1.2", path.join(os.scriptdir(), "patches", "1.2", "add-xmake.patch"), "a7eb6abd96a9c4b98824c2101378022cb905d8ed64b9a7fc5badc13c3188438c")
+
+    on_install(function (package)
```
waruqi, line 10: If xmake.lua is small and not much different from the version, it is recommended to write it directly to the current source directory.

```lua
io.writefile("xmake.lua", [[
target("irrXML")
    set_kind("static")
    add_headerfiles("src/*.h")
    add_files("src/*.cpp")]]

Comment (waruqi) [reply packages/i/irrXML/xmake.lua:10]: In this case, writing directly to xmake.lua will be more convenient for maintenance than adding patches.

**Comment by PucklaMotzer09** [comment]
The test of the package fails.
Am I missing something here?
```lua
on_test(function (package)
        assert(package:has_cxxfuncs("irr::io::createIrrXMLReader", {includes = "irrXML.h"}))
end)
```

waruqi:
> The test of the package fails.
> Am I missing something here?
> 
> ```lua
> on_test(function (package)
>         assert(package:has_cxxfuncs("irr::io::createIrrXMLReader", {includes = "irrXML.h"}))
> end)
> ```
> 
> `on_test` seems to have problems with namespaces.
> 
> ```
> checkinfo: ...amdir/core/sandbox/modules/import/core/tool/compiler.lua:88: @programdir/modules/core/tools/gcc.lua:458: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200901/_BA5A9E5F76034F008EB290B59EB0BB20.cpp:9:20: error: use of undeclared identifier 'pirr'; did you mean 'irr'?
>     volatile void* pirr::io::createIrrXMLReader = (void*)&irr::io::createIrrXMLReader;;
>                    ^~~~
>                    irr
> ```

Please try:

```lua
        assert(package:has_cxxfuncs("irr::io::createIrrXMLReader(0, 0, pass some default args)", {includes = "irrXML.h"}))
```

waruqi:
Or you can add c++ snippet.

https://github.com/xmake-io/xmake-repo/blob/1b31502390c7abe2c0193adefe82beba663e369a/packages/c/co/xmake.lua#L12-L20

PucklaMotzer09:
Why does it fail to fetch the package?
```
finding irrXML from xmake ..
finding irrXML from brew ..
finding irrXML from vcpkg ..
finding irrXML from conan ..

error: @programdir/actions/require/impl/action/install.lua:191: fetch irrXML-1.2 failed!
```

waruqi:
> Why does it fail to fetch the package?
> 
> ```
> finding irrXML from xmake ..
> finding irrXML from brew ..
> finding irrXML from vcpkg ..
> finding irrXML from conan ..
> 
> error: @programdir/actions/require/impl/action/install.lua:191: fetch irrXML-1.2 failed!
> ```

This is the bug of xmake/find_package. xmake cannot find a package with an upper name on a case-sensitive system (linux). I have now fixed this problem on the dev branch of xmake. But before the release, you can only change irrXML to irrxml.

packages/i/irrXML/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+        io.writefile("xmake.lua", [[
+        set_project("irrXML")
+        add_rules("mode.debug", "mode.release")
+        target("irrXML")
+            set_kind("static")
+            add_headerfiles("src/*.h")
+            add_files("src/*.cpp")
+        ]])
+
```
waruqi, line 18:
```lua
        io.writefile("xmake.lua", [[
        target("irrxml")
            set_kind("static")
            add_headerfiles("src/*.h")
            add_files("src/*.cpp")
        ]])
```

packages/i/irrXML/xmake.lua
```diff
@@ -0,0 +1,24 @@
+package("irrXML")
+
```
waruqi, line 2: package("irrxml")

packages/i/irrXML/xmake.lua
```diff
@@ -8,9 +8,9 @@ package("irrXML")
 
     on_install(function (package)
         io.writefile("xmake.lua", [[
-        set_project("irrXML")
+        set_project("irrxml")
```
waruqi, line 11: we can remove this line, it is optional for package.

PucklaMotzer09:
Apparently irrxml is not required for assimp (#58). The order of the links was just wrong. But it is required when the user decides to use ```system_irrxml```

waruqi:
In order to ensure that xmake-repo does not have a long commit history, please merge all the commits into single commit.

Thanks.

PucklaMotzer09:
Here you go

waruqi:
thanks.

PR #60, opened 2020-09-01
Add box2d

PucklaMotzer09:
Adds the box2d physics library

### DIFF
--- a/packages/b/box2d/xmake.lua
+++ b/packages/b/box2d/xmake.lua
@@ -1,0 +1,79 @@
+package("box2d")
+
+    set_homepage("https://box2d.org")
+    set_description("A 2D Physics Engine for Games")
+
+    if is_plat("windows", "linux", "macosx") then
+        set_urls("https://github.com/erincatto/box2d/archive/v$(version).zip")
+        add_versions("2.4.0", "6aebbc54c93e367c97e382a57ba12546731dcde51526964c2ab97dec2050f8b9")
+    end
+
+    add_deps("cmake")
+
+    on_install("windows", "linux", "macosx", function (package)
+        local config = {}
+
+        table.insert(config, "-DBOX2D_BUILD_UNIT_TESTS=OFF")
+        table.insert(config, "-DBOX2D_BUILD_TESTBED=OFF")
+        table.insert(config, "-DBOX2D_BUILD_DOCS=OFF")
+
+        if xmake.version():ge("2.3.7") then
+            import("package.tools.cmake").build(package, config, {buildir = "build"})
+        else
+            local vs_runtime = package:config("vs_runtime")
+            if vs_runtime then
+                table.insert(config, '-DCMAKE_CXX_FLAGS_DEBUG="/' .. vs_runtime .. 'd"')
+                table.insert(config, '-DCMAKE_CXX_FLAGS_RELEASE="/' .. vs_runtime .. '"')
+                table.insert(config, '-DCMAKE_C_FLAGS_DEBUG="/' .. vs_runtime .. 'd"')
+                table.insert(config, '-DCMAKE_C_FLAGS_RELEASE="/' .. vs_runtime .. '"')
+            end
+            table.insert(config, "-S")
+            table.insert(config, ".")
+            table.insert(config, "-B")
+            table.insert(config, "build")
+
+            os.mkdir("build")
+            os.vrunv("cmake", config)
+
+            if is_plat("windows") then
+                local old_dir = os.cd("build")
+
+                local configs = {}
+                local arch = package:is_arch("x86") and "Win32" or "x64"
+                local mode = package:debug() and "Debug" or "Release"
+
+                table.insert(configs, "/property:Configuration=" .. mode)
+                table.insert(configs, "/property:Platform=" .. arch)
+                table.insert(configs, "box2d.sln")
+
+                import("package.tools.msbuild").build(package, configs)
+
+                local build_dir = path.join(arch, mode)
+                os.cp(path.join(build_dir, "*.lib"), package:installdir("lib"))
+
+                os.cd(old_dir)
+            else
+                os.vrunv("make", {"-j4"}, {curdir = "build"})
+            end
+
+            package:add("linkdirs", "lib")
+            package:add("links", "box2d")
+        end
+        local mode = package:debug() and "Debug" or "Release"
+
+        if is_plat("linux", "macosx") then
+            os.trycp("build/src/*.a", package:installdir("lib"))
+        else
+            os.trycp("build/src/" .. mode .. "/*.lib", package:installdir("lib"))
+        end
+        os.cp("include", package:installdir())
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+                void test(int argc, char** argv) {
+                    b2World world(b2Vec2(0.0f, -10.0f));
+                }
+            ]]}, {configs = {languages = "c++11"},
+                  includes = "box2d/box2d.h"}))
+    end)


PucklaMotzer09:
I think the checks fail because cmake on linux, macosx and mingw tries to always copy the install/include folder, but box2d doesn't create one.
https://github.com/xmake-io/xmake/blob/master/xmake/modules/package/tools/cmake.lua#L150

waruqi:
> I think the checks fail because cmake on linux, macosx and mingw tries to always copy the install/include folder, but box2d doesn't create one.
> https://github.com/xmake-io/xmake/blob/master/xmake/modules/package/tools/cmake.lua#L150

There is no `make install` command in the makefile generated by cmake

waruqi:
Or you can run `os.runv("cmake", {...})` command or call `build.sh/build.bat` script to build it.

waruqi:
I have updated xmake/dev to add cmake.build interface. you can try

```lua
        if xmake.version():ge("2.3.7") then
            import("package.tools.cmake").build(package, config, {buildir = "build"})
            os.vrun("./build.bat")
        else
            os.vrun("./build.sh")
        end
        os.trycp("build/src/*.a", package:installdir("lib"))
        os.trycp("build/src/*.lib", package:installdir("lib"))
        os.cp("include", package:installdir())

```

packages/b/box2d/patches/2.4.0/fix-build-scripts.patch
```diff
@@ -0,0 +1,38 @@
...
+index 3433595..482d376 100644
+--- a/build.sh
++++ b/build.sh
+@@ -4,5 +4,5 @@
+ rm -rf build
+ mkdir build
+ cd build
+-cmake -DBOX2D_BUILD_DOCS=OFF ..
++cmake $@ ..
+ cmake --build .
```
waruqi, line 35: These two scripts are only a few lines, we can directly write them in on_install().

```lua
os.mkdir("build")
os.vrunv("cmake", {"--build", "."}, {curdir = "build"}) 
```

This is more common, there is no need to write a separate patch file.

packages/b/box2d/xmake.lua
```diff
@@ -0,0 +1,76 @@
...
+        else
+            local contents = io.readfile("build.bat")
+            print(contents)
+            local contents = io.readfile("build.sh")
+            print(contents)
+            if is_plat("windows") then
+                os.vrun("build.bat", config)
+            else
+                os.vrun("./build.sh", config)
+            end
```
waruqi, line 45:
```lua
if xx then
else
    os.mkdir("build")
    os.vrunv("cmake", {"--build", "."}, {curdir = "build"})
end
```

packages/b/box2d/xmake.lua
```diff
@@ -0,0 +1,96 @@
...
+            table.insert(config, "-S")
+            table.insert(config, ".")
+            table.insert(config, "-B")
+            table.insert(config, "build")
+
+            os.mkdir("build")
+            os.vrunv("cmake", config)
+
+            if is_plat("windows") then
+                os.cd("build")
```
waruqi, line 48: local oldir = os.cd("builld")


os.cd(oldir)

PucklaMotzer09:
linux doesn't have any links or linkdirs.
```
finding box2d from xmake ..
{ 
  version = "2.4.0",
  includedirs = { 
    "/home/runner/.xmake/packages/b/box2d/2.4.0/77c75344736c4bf88a4590c6c3fa9639/include",
  },
}
```

waruqi:
> linux doesn't have any links or linkdirs.
> 
> ```
> finding box2d from xmake ..
> { 
>   version = "2.4.0",
>   includedirs = { 
>     "/home/runner/.xmake/packages/b/box2d/2.4.0/77c75344736c4bf88a4590c6c3fa9639/include",
>   },
> }
> ```

The .a library was not found in the installdir/lib directory.

packages/b/box2d/xmake.lua
```diff
@@ -0,0 +1,113 @@
...
+                local old_dir = os.cd("build")
+
+                local configs = {}
+                local arch = package:is_arch("x86") and "Win32" or "x64"
+                local mode = package:debug() and "Debug" or "Release"
+
+                table.insert(configs, "/property:Configuration=" .. mode)
+                table.insert(configs, "/property:Platform=" .. arch)
+                table.insert(configs, "box2d.sln")
+
```
waruqi, line 54: we need set "/MT" or "/MD" for vs runtime.

```lua
        local vs_runtime = package:config("vs_runtime")
        if vs_runtime then
            table.insert(configs, '-DCMAKE_CXX_FLAGS_DEBUG="/' .. vs_runtime .. 'd"')
            table.insert(configs, '-DCMAKE_CXX_FLAGS_RELEASE="/' .. vs_runtime .. '"')
            table.insert(configs, '-DCMAKE_C_FLAGS_DEBUG="/' .. vs_runtime .. 'd"')
            table.insert(configs, '-DCMAKE_C_FLAGS_RELEASE="/' .. vs_runtime .. '"')
        end
```

see https://github.com/xmake-io/xmake/blob/e49ef767d42e4e4cfd1ecf659e5fd1ea4b8fa9f9/xmake/modules/package/tools/cmake.lua#L30-L37

PucklaMotzer09:
MinGW has got undefined references even though he finds the library and links to it.
```
> /usr/local/opt/mingw-w64/bin/x86_64-w64-mingw32-gcc -c -std=c++11 -I/Users/runner/.xmake/packages/b/box2d/2.4.0/76c2909639c24ff4861d24045fb57525/include -m64 -o /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_8979D576C76D43108C49F08F59B69120.o /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_98586F2EBB95431086C0AAEA237AD320.cpp
> /usr/local/opt/mingw-w64/bin/x86_64-w64-mingw32-g++ -o /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_2DC128E8AB25421084163782E4600820.b /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_8979D576C76D43108C49F08F59B69120.o -L/Users/runner/.xmake/packages/b/box2d/2.4.0/76c2909639c24ff4861d24045fb57525/lib -lbox2d -m64
> checking for the c++ includes(box2d/box2d.h)
> checking for the c++ links(box2d)
> checking for the c++ snippet(test)
checkinfo: ...gramdir/core/sandbox/modules/import/core/tool/linker.lua:91: @programdir/core/sandbox/modules/os.lua:258: /usr/local/Cellar/mingw-w64/7.0.0_2/toolchain-x86_64/bin/x86_64-w64-mingw32-ld: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_8979D576C76D43108C49F08F59B69120.o:_98586F2EBB95431086C0AAEA237AD320.cpp:(.text+0x4d): undefined reference to `b2World::b2World(b2Vec2 const&)'
/usr/local/Cellar/mingw-w64/7.0.0_2/toolchain-x86_64/bin/x86_64-w64-mingw32-ld: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.xmake501/200903/_8979D576C76D43108C49F08F59B69120.o:_98586F2EBB95431086C0AAEA237AD320.cpp:(.text+0x59): undefined reference to `b2World::~b2World()'
collect2: error: ld returned 1 exit status
```

packages/b/box2d/xmake.lua
```diff
@@ -0,0 +1,125 @@
...
+    on_install("windows", "linux", "macosx", function (package)
+        local config = {}
+        local function add_config_arg(config_name, cmake_name)
+            table.insert(config, "-D" .. cmake_name .. "=" .. (package:config(config_name) and "ON" or "OFF"))
+        end
+
+        add_config_arg("build_unit_tests", "BOX2D_BUILD_UNIT_TESTS")
+        add_config_arg("build_testbed",    "BOX2D_BUILD_TESTBED")
+        add_config_arg("build_docs",       "BOX2D_BUILD_DOCS")
```
waruqi, line 31: I don't think there is a need to build documentation and tests, they will never be used. Because these packages are not installed in the system environment, they are only used to integrate xmake projects

PucklaMotzer09: I thought I would just add these configs since these are options of cmake. But I could just set those to false on default. ```build_docs``` is already false on default. Or should I remove them completely?

waruqi: I think these configurations are not needed, because these configuration options are never enabled in add_requires. 

Please remove these unnecessary options and try to keep `package()` simple.

waruqi:
On the master version, I have passed mingw/iphoneos/android toolchain information to cmake. Therefore, in the next version, we can call cmake through xmake in on_install to support the compilation of mingw/android/iphoneos library

PR #70, opened 2020-09-18
Add nana 1.7.4

PucklaMotzer09:
Adds version 1.7.4 of **nana** to the repository. Every version below 1.7.3 is not compatible with windows so I added this version. I had to write this code
```lua
if is_plat("windows") then
            local file_name = path.join(os.curdir(), "source", "system", "split_string.cpp")
            local contents = io.readfile(file_name)
            contents = contents:gsub(" and ", " && ")
            io.writefile(file_name, contents)
        end
```

because a patch somehow didn't work.


packages/n/nana/xmake.lua
```diff
@@ -7,23 +7,36 @@ package("nana")
...
     elseif is_plat("linux") then
         add_syslinks("pthread", "X11", "Xft", "fontconfig")
     end
 
     on_install("linux", "windows", function (package)
+        if is_plat("windows") then
+            local file_name = path.join(os.curdir(), "source", "system", "split_string.cpp")
+            local contents = io.readfile(file_name)
+            contents = contents:gsub(" and ", " && ")
+            io.writefile(file_name, contents)
```
waruqi, line 28: Please use `io.gsub(filename, " and ", " && ")` instead of these three lines.

PR #71, opened 2020-09-22
libuv: fix shared lib name on windows

paul-reilly:
`uv` is the .lib that is associated with the DLL for libuv.

### DIFF
--- a/packages/l/libuv/xmake.lua
+++ b/packages/l/libuv/xmake.lua
@@ -28,6 +28,10 @@ package("libuv")
         package:add("syslinks", "advapi32", "iphlpapi", "psapi", "user32", "userenv", "ws2_32", "kernel32", "gdi32", "winspool", "shell32", "ole32", "oleaut32", "uuid", "comdlg32")
     end)
 
+    on_load("linux", function (package)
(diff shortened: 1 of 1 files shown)


waruqi:
>  /unix/thread.c:399: undefined reference to `pthread_once

maybe we need add syslinks(pyhread) for linux.

paul-reilly:
Yes, that was the one!

packages/l/libuv/xmake.lua
```diff
@@ -28,6 +28,10 @@ package("libuv")
         package:add("syslinks", "advapi32", "iphlpapi", "psapi", "user32", "userenv", "ws2_32", "kernel32", "gdi32", "winspool", "shell32", "ole32", "oleaut32", "uuid", "comdlg32")
     end)
 
+    on_load("linux", function (package)
+        package:add("syslinks", "pthread")
+    end)
```
waruqi, line 33: we need only use add_syslinks()

```lua
if is_plat("linux") then
    add_syslinks("pthread")
end
```

waruqi:
In addition, in order to reduce the commit history, please merge into a single commit. Thanks.

PR #82, opened 2020-10-08
Fix links and add test to glew

PucklaMotzer09:
This PR fixes the following
+ Adds test
+ Correct links when static linking and dynamic linking
  + glew32s and glew32
+ Correct Preprocessor Macros when static linking and dynamic linking
+ Correct syslinks

### DIFF
--- a/packages/g/glew/xmake.lua
+++ b/packages/g/glew/xmake.lua
@@ -11,8 +11,29 @@ package("glew")
         add_versions("2.1.0", "2700383d4de2455f06114fbaf872684f15529d4bdc5cdea69b5fb0e9aa7763f1")
     end
 
-    on_load("windows", "mingw", function (package)
-        package:add("links", is_plat("windows") and "glew32s" or "glew32")
+    if is_plat("windows", "mingw") then
+        add_syslinks("glu32", "opengl32")
+    else
+        add_links("GLEW")
+        if is_plat("linux") then
+            add_syslinks("GLU", "GL")
+        elseif is_plat("macosx") then
+            add_frameworks("OpenGL")
+        end
+    end
+
+    on_load(function (package)
+        if package:config("shared") then
+            package:add("defines", "GLEW_BUILD")
+            if package:is_plat("windows", "mingw") then
+                package:add("links", "glew32")
+            end
+        else
+            package:add("defines", "GLEW_STATIC")
+            if package:is_plat("windows", "mingw") then
+                package:add("links", "glew32s")
+            end
+        end
     end)
 
     if is_plat("mingw") then
@@ -31,7 +52,7 @@ package("glew")
     end)
 
     on_install("linux", "macosx", function (package)
-        os.vrun("make")
+        os.vrun("make", {"glew.lib." .. (package:config("shared") and "shared" or "static")})
         os.cp("lib", package:installdir())
         os.cp("include", package:installdir())
     end)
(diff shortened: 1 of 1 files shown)


PucklaMotzer09:
The linux build is failing because ```libGL``` cannot be found. This can be fixed by installing a OpenGL library:
```
sudo apt-get install libgl1-mesa-dev
```

waruqi:
> The linux build is failing because `libGL` cannot be found. This can be fixed by installing a OpenGL library:
> 
> ```
> sudo apt-get install libgl1-mesa-dev
> ```

In order to pass the test, you can temporarily add it in linux.yml

PucklaMotzer09:
It seems that the MinGW build is building for MacOSX. The arguments are the exact same. There even is ```-framework OpenGL``` which only gets added when the platform is ```macosx```

waruqi:
> It seems that the MinGW build is building for MacOSX. The arguments are the exact same. There even is `-framework OpenGL` which only gets added when the platform is `macosx`

? Yes, what's the problem now?

PucklaMotzer09:
So this is expected behaviour? I thought that it would compile more like the windows version like it used to in all the other libraries.

waruqi:
> So this is expected behaviour? I thought that it would compile more like the windows version like it used to in all the other libraries.

On macos, we can also use
mingw sdk to compile windows libraries.

PucklaMotzer09:
Should the Installation step of linux stay or what are we going to do about it?

waruqi:
And we can alse add msys2/mingw on windows for ci.

like https://github.com/xmake-io/xmake/blob/master/.github/workflows/msys2_mingw.yml

waruqi:
> Should the Installation step of linux stay or what are we going to do about it?

If the tests have all passed, you can revert linux.yml and I will merge it

PucklaMotzer09:
I removed the Installation step. You would have to add the following command
```
sudo apt-get install -y libgl1-mesa-dev libglu1-mesa-dev
```

packages/g/glew/xmake.lua
```diff
@@ -11,8 +11,29 @@ package("glew")
...
+    on_load(function (package)
+        if package:config("shared") then
+            package:add("defines", "GLEW_BUILD")
+            if package:is_plat("windows", "mingw") then
+                package:add("links", "glew32")
+            end
+        else
+            package:add("defines", "GLEW_STATIC")
+            if package:is_plat("windows", "mingw") then
+                package:add("links", "glew32s")
```
waruqi, line 34: and here. syslinks

PucklaMotzer09: The syslinks are always the same. So I added them above using ```add_syslinks```

waruqi: You can use add_syslinks to instead of `package:add("links")` or use `package:add("syslinks")`

PucklaMotzer09: But glew32 is not a syslink.

waruqi: Oh, glew32 is not. my mistake.

waruqi:
I have merged it.

waruqi:
> It seems that the MinGW build is building for MacOSX. The arguments are the exact same. There even is `-framework OpenGL` which only gets added when the platform is `macosx`

I have added msys/mingw ci and updated xmake/dev to support for using cmake/autotools on msys/mingw.

PR #83, opened 2020-10-08
Fix links and add test to glfw

PucklaMotzer09:
Fixes the following
+ Adds test
+ Correct libraries and defines for shared and static
+ Disables compilation of unneccesary options in linux

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,6 +19,10 @@ jobs:
         with:
           xmake-version: branch@dev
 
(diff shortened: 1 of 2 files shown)


waruqi:
In order to better support vs_runtime (/MT and /MD) and more flexibly support various configuration options, we switched glfw to source code compilation and installation.

Maybe you need to pay attention to this change to confirm whether your project is affected.

see patch: https://github.com/xmake-io/xmake-repo/pull/103

PR #91, opened 2020-10-30
add package blosc

xq114:
add package blosc

### DIFF
--- a/packages/b/blosc/xmake.lua
+++ b/packages/b/blosc/xmake.lua
@@ -9,16 +9,19 @@ package("blosc")
     add_versions("1.20.1", "42c4d3fcce52af9f8e2078f8f57681bfc711706a3330cb72b9b39e05ae18a413")
 
     add_deps("cmake")
+    if is_plat("linux") then
+        add_syslinks("pthread")
+    end
 
     on_install("macosx", "linux", "windows", "mingw", function (package)
         local configs = {"-DBUILD_TESTS=OFF", "-DBUILD_BENCHMARKS=OFF"}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         if package:config("shared") then
-                table.insert(configs, "-DBUILD_SHARED=ON")
-                table.insert(configs, "-DBUILD_STATIC=OFF")
+            table.insert(configs, "-DBUILD_SHARED=ON")
+            table.insert(configs, "-DBUILD_STATIC=OFF")
         else
-                table.insert(configs, "-DBUILD_SHARED=OFF")
-                table.insert(configs, "-DBUILD_STATIC=ON")
+            table.insert(configs, "-DBUILD_SHARED=OFF")
+            table.insert(configs, "-DBUILD_STATIC=ON")
         end
         import("package.tools.cmake").install(package, configs)
     end)
--- a/packages/b/blosc/xmake.lua
+++ b/packages/b/blosc/xmake.lua
@@ -11,7 +11,7 @@ package("blosc")
     add_deps("cmake")
 
     on_install("macosx", "linux", "windows", "mingw", function (package)
-        configs = {}
+        local configs = {}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         import("package.tools.cmake").install(package, configs)
         if is_plat("windows") then
--- a/packages/b/blosc/xmake.lua
+++ b/packages/b/blosc/xmake.lua
@@ -1,0 +1,44 @@
+package("blosc")
+
+    set_homepage("https://www.blosc.org/")
+    set_description("A blocking, shuffling and loss-less compression library")
+
+    add_urls("https://github.com/Blosc/c-blosc/archive/v$(version).tar.gz",
+             "https://github.com/Blosc/c-blosc")
+
+    add_versions("1.20.1", "42c4d3fcce52af9f8e2078f8f57681bfc711706a3330cb72b9b39e05ae18a413")
+
+    add_deps("cmake")
+
+    on_install("macosx", "linux", "windows", "mingw", function (package)
+        configs = {}
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        import("package.tools.cmake").install(package, configs)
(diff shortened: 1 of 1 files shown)


packages/b/blosc/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+
+    add_urls("https://github.com/Blosc/c-blosc/archive/v$(version).tar.gz",
+             "https://github.com/Blosc/c-blosc")
+
+    add_versions("1.20.1", "42c4d3fcce52af9f8e2078f8f57681bfc711706a3330cb72b9b39e05ae18a413")
+
+    add_deps("cmake")
+
+    on_install("macosx", "linux", "windows", "mingw", function (package)
+        configs = {}
```
waruqi, line 14: local configs = {}

packages/b/blosc/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        import("package.tools.cmake").install(package, configs)
+        if is_plat("windows") then
+            os.rm(path.join(package:installdir("lib"), "blosc.lib"))
+        elseif is_plat("mingw") then
+            os.rm(path.join(package:installdir("lib"), "libblosc.dll.a"))
+        end
```
waruqi, line 21: we need not remove these files. xmake will automatically clean up these cache files.

xq114: These files provide "address to dynamic link library", and will be useful only if the `.dll` files is in the executable search directories or we use something like gcc -rpath flag during link stage, which is unfortunately disabled on Windows. Otherwise the executable can be compiled normally but not run. Tests on my local environment have shown, since compiler flag `-lblosc` prefer these files, I can't link to the true static libraries which are named 'libblosc.lib' and 'libblosc.a'. That's why I removed these files manually, and I wonder if there is a better solution.

waruqi: You should also deal with package:config("shared") and disable tests

https://github.com/Blosc/c-blosc/blob/5352508a420bec865c57cda116a5e5303df898d2/CMakeLists.txt#L93-L98

packages/b/blosc/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+
+    add_versions("1.20.1", "42c4d3fcce52af9f8e2078f8f57681bfc711706a3330cb72b9b39e05ae18a413")
+
+    add_deps("cmake")
+
+    on_install("macosx", "linux", "windows", "mingw", function (package)
+        local configs = {}
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        import("package.tools.cmake").install(package, configs)
+        if is_plat("windows") then
```
waruqi, line 17: package:is_plat("windows")

packages/b/blosc/xmake.lua
waruqi:
```lua
local configs = {"-DBUILD_TESTS=OFF"}
table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
        table.insert(configs, "-DBUILD_SHARED=ON")
        table.insert(configs, "-DBUILD_STATIC=OFF")
else
        table.insert(configs, "-DBUILD_SHARED=OFF")
        table.insert(configs, "-DBUILD_STATIC=ON")
end
```

xq114: It works. Thank you!

packages/b/blosc/xmake.lua
```diff
@@ -0,0 +1,46 @@
...
+
+    add_versions("1.20.1", "42c4d3fcce52af9f8e2078f8f57681bfc711706a3330cb72b9b39e05ae18a413")
+
+    add_deps("cmake")
+
+    on_install("macosx", "linux", "windows", "mingw", function (package)
+        local configs = {"-DBUILD_TESTS=OFF", "-DBUILD_BENCHMARKS=OFF"}
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        if package:config("shared") then
+                table.insert(configs, "-DBUILD_SHARED=ON")
```
waruqi, line 17: 4 spaces, not 8 spaces

waruqi:
linux ci fails.

```
blosc.c:(.text+0x50e): undefined reference to `pthread_atfork'
/home/runner/.xmake/packages/b/blosc/1.20.1/531f88cff3a64da0aa3fb05ee281ab7d/lib/libblosc.a(blosc.c.o): In function `t_blosc':
blosc.c:(.text+0xa7b): undefined reference to `pthread_barrier_wait'
```

we need add pthread links.

```lua
package("")
    if is_plat("linux") then
        add_syslinks("pthread")
    end
```

xq114:
> linux ci fails.
> 
> ```
> blosc.c:(.text+0x50e): undefined reference to `pthread_atfork'
> /home/runner/.xmake/packages/b/blosc/1.20.1/531f88cff3a64da0aa3fb05ee281ab7d/lib/libblosc.a(blosc.c.o): In function `t_blosc':
> blosc.c:(.text+0xa7b): undefined reference to `pthread_barrier_wait'
> ```
> 
> we need add pthread links for linux.
> 
> ```lua
> package("blosc")
>     if is_plat("linux") then
>         add_syslinks("pthread")
>     end
> ```

Do we need `package:is_plat`? And where should we put the `add_syslinks`, just after the first line `package`, or before `on_install`?

waruqi:
> > linux ci fails.
> > ```
> > blosc.c:(.text+0x50e): undefined reference to `pthread_atfork'
> > /home/runner/.xmake/packages/b/blosc/1.20.1/531f88cff3a64da0aa3fb05ee281ab7d/lib/libblosc.a(blosc.c.o): In function `t_blosc':
> > blosc.c:(.text+0xa7b): undefined reference to `pthread_barrier_wait'
> > ```
> > 
> > 
> > we need add pthread links for linux.
> > ```lua
> > package("blosc")
> >     if is_plat("linux") then
> >         add_syslinks("pthread")
> >     end
> > ```
> 
> Do we need `package:is_plat`? And where should we put the `add_syslinks`, just after the first line `package`, or before `on_install`?

only need package:is_plat in on_install

PR #94, opened 2020-10-31
update gtest to 1.10.0 and add gmock into gtest

45degree:
gmock is a part of googletest, but the repository did not include it.


waruqi:
Thanks. :+1:

PR #104, opened 2020-11-05
Update curl version list

SirLynix:
See https://github.com/xmake-io/xmake-repo/pull/101


SirLynix:
Merged with base branch.

I'm wondering, should we disable test building? It doesn't seem really useful here (and is causing compilation issues with msvc and libcurl 7.64.0 because of an empty file)

waruqi:
If it is a test build of libcurl itself, as long as cmakelist provides an option to disable them, we usually set it to disable.

SirLynix:
Yes it is, I'll make a separate PR for this.

packages/l/libcurl/versions.lua
```diff
@@ -0,0 +1,45 @@
+add_versions("7.73.0", "cf34fe0b07b800f1c01a499a6e8b2af548f6d0e044dca4a29d88a4bee146d131")
```
waruqi, line 1: Please modify as I said, you can only use this way to includes it.

```
includes("versions.lua")
package("curl")
    add_urls()
    add_versions_list()
versions.lua

function add_versions_list()
end
```

SirLynix:
Following #101 

I made the changes, except for two things: 
- I used includes directly instead of declaring a function to prevent declaring a global `add_versions_list` variable which would then pollute the global namespace.
- I removed the patched versions (example: instead of keeping libcurl 7.65.0 I kept 7.65.3), it seems to make more sense this way.

Sorry about all the failing checks from the previous commit.

packages/l/libcurl/versions.lua
SirLynix: I did this to prevent polluting the global namespace (I was writing my comment) and letting a `add_versions_list` function exist / being overridden or worse.

waruqi:
> I used includes directly instead of declaring a function to prevent declaring a global add_versions_list variable which would then pollute the global namespace.

No, xmake.lua of each package is completely isolated and will not affect

SirLynix:
Oh alrighty then. Another thing I didn't know. I'm making the changes as requested.

SirLynix:
Well this time I don't understand why it's failing, maybe it's not including the good versions.lua file?

waruqi:
Oh, it's the first time I use include in package/xmake.lua, I need to look at it, please wait. : (

SirLynix:
No problem, I'm trying some stuff on my side too. It seems the versions.lua file is not executed (adding `print` statements inside this file has no effect), where is includes defined?

waruqi:
You should write: 

```lua
includes(path.join(os.scriptdir(), "versions.lua"))
```

SirLynix:
Perfect! Thank you.

waruqi:
ci fails, maybe we need add `--without-zstd` for macos

```lua
        if is_plat("macosx") then
            -- ...
            table.insert(configs, "--without-zstd")
        end
```

SirLynix:
Should I make another PR?

waruqi:
No, please modify it directly on the current pr

waruqi:
Thank you very much for your contribution. :+1:

SirLynix:
No problem 😸 I think you can expect a few more soon!

PR #105, opened 2020-11-05
libcurl: Disable testing build

SirLynix:
Currently libcurl is built with tests, which can be long and can break installation (7.64.1 doesn't build with msvc because of an empty test file).

This PR disables libcurl test building.

### DIFF
--- a/packages/l/libcurl/xmake.lua
+++ b/packages/l/libcurl/xmake.lua
@@ -22,7 +22,8 @@ package("libcurl")
     end
 
     on_install("windows", function (package)
-        import("package.tools.cmake").install(package)
+        local configs = { "-DBUILD_TESTING=OFF" }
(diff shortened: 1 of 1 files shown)


packages/l/libcurl/xmake.lua
```diff
@@ -22,7 +22,8 @@ package("libcurl")
     end
 
     on_install("windows", function (package)
-        import("package.tools.cmake").install(package)
+        local configs = { "-DBUILD_TESTING=OFF" }
```
waruqi, line 25: please remove spaces `{"-DBUILD_TESTING=OFF"}`

SirLynix: Sure, fix'd!

waruqi:
Thanks! :+1:

PR #110, opened 2020-11-06
Add hopscotch-map

yamashi:

### DIFF
--- a/packages/h/hopscotch-map/xmake.lua
+++ b/packages/h/hopscotch-map/xmake.lua
@@ -1,0 +1,28 @@
+package("hopscotch-map")
+
+    set_homepage("https://github.com/Tessil/hopscotch-map")
+    set_description("A C++ implementation of a fast hash map and hash set using hopscotch hashing")
+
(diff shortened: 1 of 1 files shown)


packages/h/hopscotch-map/xmake.lua
```diff
@@ -0,0 +1,28 @@
+package("hopscotch-map")
+
+    set_homepage("https://github.com/Tessil/hopscotch-map")
+    set_description("A C++ implementation of a fast hash map and hash set using hopscotch hashing")
+
+    set_urls("https://github.com/Tessil/hopscotch-map/archive/v$(version).zip",
+             "https://github.com/Tessil/hopscotch-map.git")
+
+    add_versions("2.3.0", "56ce4ff67215656065ee1a08948533baf9447c4440196ea5133c024856006938")
```
waruqi, line 9: I saw git tag is v2.3.0. we can use `add_versions("v2.3.0")` and `/$(version).zip`. 

so we can also select tag version from git url automatically.

PR #114, opened 2020-11-07
add package libuuid

xq114:


packages/l/libuuid/xmake.lua
```diff
@@ -0,0 +1,26 @@
+package("jsoncpp")
```
SirLynix, line 1: Wrong package name

waruqi: Oh, I fixed it. Thanks!

PR #140, opened 2020-11-16
Sol2: Add includes_lua option to be able to ship a custom Lua library

SirLynix:
In the case of Lua, it's not uncommon to ship a custom Lua library (to update the langage a bit for example), but sol2 forces the lua dependency, this adds an option to disable the default lua shipping.


waruqi:
Thanks!

PR #141, opened 2020-11-16
Lua: add versions 5.4.1 and 5.3.6

SirLynix:
While making #140 I saw that the lua package version list was not up to date, this PR fixes it.


SirLynix:
I don't understand why it doesn't work, or even why it worked before. It seems like no lua.lib is generated although symbols are exported.

SirLynix:
It looks like it's the `is_kind` function which returns false, even when the kind is `shared`

packages/l/lua/xmake.lua
```diff
@@ -21,8 +22,10 @@ package("lua")
 
     on_install("linux", "macosx", "windows", "android", "bsd", function (package)
         io.writefile("xmake.lua", format([[
+            local kind = "%s"
+
             target("lualib")
-                set_kind("%s")
+                set_kind(kind)
```
waruqi, line 28: please revert this line

packages/l/lua/xmake.lua
```diff
@@ -34,8 +37,10 @@ package("lua")
                     add_defines("LUA_USE_MACOSX")
                     add_defines("LUA_DL_DYLD")
                 elseif is_plat("windows") then
-                    add_defines("LUA_USE_WINDOWS")
-                    add_defines("LUA_DL_DLL")
+                    -- Lua already detects Windows and sets according defines
+                    if kind == "shared" then
```
waruqi, line 41: we can use `if is_kind("shared") then`

waruqi:
> It looks like it's the `is_kind` function which returns false, even when the kind is `shared`
> Edit: Confirmed, is_kind only returns true when the kind is set from the command-line, maybe I misunderstood something about this function.

```lua
import("package.tools.xmake").install(package, {kind = package:config("shared") and "shared" or "static"})
```

PR #142, opened 2020-11-16
libcurl: Disable LDAP support on Windows too

SirLynix:
LDAP support requires a link to wldap32.lib, as its support is disabled on Linux this disables it on Windows too.

Initiality I added wldap32 as a syslink but it failed the compilation, for some reasons I can't explain.
I also added advapi32 and winmm dependencies (from [here](https://github.com/curl/curl/blob/master/CMakeLists.txt#L310))

### DIFF
--- a/packages/l/libcurl/xmake.lua
+++ b/packages/l/libcurl/xmake.lua
@@ -22,7 +22,7 @@ package("libcurl")
     elseif is_plat("linux") then
         add_syslinks("pthread")
     elseif is_plat("windows", "mingw") then
-        add_syslinks("ws2_32")
+        add_syslinks("winmm", "wldap32", "ws2_32")
     end
 
     on_load("windows", function (package)
--- a/packages/l/libcurl/xmake.lua
+++ b/packages/l/libcurl/xmake.lua
@@ -22,7 +22,7 @@ package("libcurl")
     elseif is_plat("linux") then
         add_syslinks("pthread")
     elseif is_plat("windows", "mingw") then
-        add_syslinks("wldap32", "ws2_32")
+        add_syslinks("ws2_32")
     end
 
     on_load("windows", function (package)
--- a/packages/l/libcurl/xmake.lua
+++ b/packages/l/libcurl/xmake.lua
@@ -22,7 +22,7 @@ package("libcurl")
     elseif is_plat("linux") then
         add_syslinks("pthread")
     elseif is_plat("windows", "mingw") then
-        add_syslinks("winmm", "wldap32", "ws2_32")
+        add_syslinks("wldap32", "ws2_32")
     end
 
     on_load("windows", function (package)
--- a/packages/l/libcurl/xmake.lua
+++ b/packages/l/libcurl/xmake.lua
@@ -35,6 +35,7 @@ package("libcurl")
         local configs = {"-DBUILD_TESTING=OFF"}
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.includes(configs, "-DCURL_DISABLE_LDAP=ON")
         import("package.tools.cmake").install(package, configs)
     end)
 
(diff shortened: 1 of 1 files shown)


SirLynix:
Check are failing because of xmake internal errors, maybe you should take a look @waruqi.

SirLynix:
Yet another fail from 
```
MSVCRT.lib(chandler4gs.obj) : error LNK2019: unresolved external symbol __except_handler4_common referenced in function __except_handler4```

I really don't understand how I'm causing this error

waruqi:
> Yet another fail from
> 
> ```
> MSVCRT.lib(chandler4gs.obj) : error LNK2019: unresolved external symbol __except_handler4_common referenced in function __except_handler4
> ```
> 
> I really don't understand how I'm causing this error

It's not that you caused the problem, it's because I later fixed the ci for mingw on macOS that triggered the previously hidden problem.

We need to add CURL_STATICLIB defines for `mingw@macos,linux`

https://github.com/xmake-io/xmake-repo/blob/1aa743cabd04bc1c5773de974c0bb24cce800f3a/packages/l/libcurl/xmake.lua#L28-L35

SirLynix:
It's also failing for Windows in static mode/x86: https://github.com/xmake-io/xmake-repo/pull/142/checks?check_run_id=1407124585

Shouldn't CURL_STATICLIB be set for every platform when building it statically? Why only for Windows?

waruqi:
> It's also failing for Windows in static mode/x86: https://github.com/xmake-io/xmake-repo/pull/142/checks?check_run_id=1407124585
>

This is another issue of vs runtime conflict. Windows x86 is also a new one I added to github ci recently. I didn’t test win x86 on ci before.

I will look at this question tomorrow, now I am going to sleep, sorry


> Shouldn't CURL_STATICLIB be set for every platform when building it statically? Why only for Windows?

This is the naming convention problem of dll export symbols only for windows/mingw, you can see 

https://github.com/curl/curl/blob/24f909cebd8b79abf7b7368b649f5dace1a2449f/include/curl/curl.h#L115

SirLynix:
> > It's also failing for Windows in static mode/x86: https://github.com/xmake-io/xmake-repo/pull/142/checks?check_run_id=1407124585
> 
> This is another issue of vs runtime conflict. Windows x86 is also a new one I added to github ci recently. I didn’t test win x86 on ci before.
> 

Oh okay, any tips for fixing this?

> I will look at this question tomorrow, now I am going to sleep, sorry

No problem, good night sir!

> > Shouldn't CURL_STATICLIB be set for every platform when building it statically? Why only for Windows?
> 
> This is the naming convention problem of dll export symbols only for windows/mingw, you can see
> 
> https://github.com/curl/curl/blob/24f909cebd8b79abf7b7368b649f5dace1a2449f/include/curl/curl.h#L115-L129

I don't get why it would be a problem to set it on other platforms then, as it will define CURL_EXTERN to nothing in both cases.

waruqi:
> It's also failing for Windows in static mode/x86: https://github.com/xmake-io/xmake-repo/pull/142/checks?check_run_id=1407124585
> Oh okay, any tips for fixing this?

This is because when on_test is doing link testing, for MT vs runtime, there is no link conflict caused by `-nodefaultlib:msvcrt.lib`

I have improved xmake on dev branch.

https://github.com/xmake-io/xmake/blob/a16354dba704b4a7613eae406812db72cfcf2bd5/xmake/core/package/package.lua#L1087-L1089

You can update to dev version to avoid this issue when you install libcurl with x86/static.

> I don't get why it would be a problem to set it on other platforms then, as it will define CURL_EXTERN to nothing in both cases.

yeah, Using cmake to compile libcurl/static in on_install will automatically define CURL_STATICLIB and define CURL_EXTERN to nothing.

However, when the user actually uses the libcurl static library through `add_packages("libcurl")`, the user does not define CURL_STATICLIB, so at this time, CURL_EXTERN will be __declspec(dllimport), so that when the user links the libcurl static library, it cannot be found libcurl symbol.

And on_test is equivalent to simulating the user's actual link to use libcurl, so this problem will also occur.

Therefore, we need to use `package:add("defines", "CURL_STATICLIB")` to let users automatically add -DCURL_STATICLIB when actually using the libcurl library

https://github.com/xmake-io/xmake-repo/blob/db3b22b6112ba75a8e0dfdca477cbf9d7f26a70a/packages/l/libcurl/xmake.lua#L28-L32

waruqi:
I have merge your patch and fix these issues. see https://github.com/xmake-io/xmake-repo/pull/144

SirLynix:
Thank you!

PR #171, opened 2020-11-30
Update Vulkan-Headers version and regular version name

juteman:
Now the Vulkan-Headers version name is vx.x.xxx, update to it.
And the newer version is v1.2.162

log: change pacages/v/vulkan-headers


waruqi:
Thanks!

PR #181, opened 2020-12-01
add bison, cpu-features and openjpeg

xq114:


xq114:
Needing rerun due to https://github.com/xmake-io/xmake/issues/1095

PR #183, opened 2020-12-02
fix: volk xmake file get two on_install funciton

juteman:
volk xmake get two on_install function, delete unused function.

log: fix packages/v/volk/xmake.lua


juteman:
It seems some building exception again.
But I exec `xmake l ./scripts/test.lua`, it pass on my Linux machine

waruqi:
I am fixing ci. please wait some time.

waruqi:
I have fixed ci. It works.

PR #214, opened 2020-12-15
Fix test of libsdl

PucklaMotzer09:
This fixes the test of **libsdl**. ```SDL_MAIN_HANDLED``` has been added as a define. Previously there was the following line
```lua
local configs = package:is_plat("windows") and {ldflags = "/SUBSYSTEM:CONSOLE"}
```
Is this required in some way? On my Windows PC it runs fine without it.

### DIFF
--- a/packages/l/libsdl/xmake.lua
+++ b/packages/l/libsdl/xmake.lua
@@ -45,6 +45,5 @@ package("libsdl")
     end)
 
     on_test(function (package)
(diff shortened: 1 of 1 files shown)


waruqi:
> This fixes the test of **libsdl**. `SDL_MAIN_HANDLED` has been added as a define. Previously there was the following line
> 
> ```lua
> local configs = package:is_plat("windows") and {ldflags = "/SUBSYSTEM:CONSOLE"}
> ```
> 
> Is this required in some way? On my Windows PC it runs fine without it.

As long as ci is passed, there is no problem.

PR #227, opened 2020-12-26
Add rltgenrandom option to mimalloc

yamashi:


waruqi:
I have merged this patch and improved it. see https://github.com/xmake-io/xmake-repo/pull/229

PR #243, opened 2021-01-16
Add SDL2 latest version (2.0.14)

AntoineJT:
This pull request adds the latest SDL2 version at the time, which is v2.0.14

I don't know why the MacOS CI fails, maybe it needs a bit of investigation, or maybe this version of SDL2 doesn't work on MacOS (but it seems pretty unlikely) 🤷


waruqi:
> This pull request adds the latest SDL2 version at the time, which is v2.0.14
> 
> I don't know why the MacOS CI fails, maybe it needs a bit of investigation, or maybe this version of SDL2 doesn't work on MacOS (but it seems pretty unlikely) 🤷

maybe you need add GameController framework for new sdl version

```lua
package("")
    on_load(function (package)
        if package:is_plat("macosx") and package:version():ge("2.0.14") then
            package:add("frameworks", "GameController")
        end
    end)
```

AntoineJT:
You were right, adding the GameController framework on MacOS to SDL2 has fixed the CI 👍

PR #256, opened 2021-01-25
Add steam's GameNetworkingSockets

yamashi:


packages/g/gamenetworkingsockets/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+
+    on_install("windows", "linux", function (package)
+        local configs = {}
+        if package:config("shared") then
+            configs.kind = "shared"
+        end
+        os.cp(path.join(package:scriptdir(), "port", "xmake.lua"), "xmake.lua")
+        import("package.tools.xmake").install(package, configs)
+
+        os.cp("include/*", package:installdir("include"))
```
yamashi, line 32: @waruqi I am not sure if there is a better way of doing this? Maybe in the port/xmake.lua, my attempts to use add_headerfiles were unsuccessful

waruqi: try `add_headerfiles("include/(stream/*.h)")` in port/xmake.lua

we use `()` to reserve path structure.

packages/g/gamenetworkingsockets/xmake.lua
```diff
@@ -9,16 +9,36 @@ package("gamenetworkingsockets")
...
 
-    add_includedirs("include/")
+    on_load("windows", "linux", function(package)
+        if not package:config("shared") then
+            package:add("defines", "STEAMNETWORKINGSOCKETS_STATIC_LINK")
+            if is_plat("windows") then
+                package:add("deps", "libsodium", "protobuf-cpp")
+                package:add("syslinks", "ws2_32")
+            else
+                package:add("deps", "openssl", "protobuf-cpp", {configs = {cxflags = "-fpic"}})
```
waruqi, line 20: we can use `package:config("pic")` to add/fix pic in openssl/protobuf-cpp packages. 

```lua
package("protobuf-cpp")
    on_install(function (package)
        if not package:config("shared") and package:config("pic") then
            -- add -fpic flags
        end
    end)
```

packages/g/gamenetworkingsockets/xmake.lua
```diff
@@ -9,16 +9,36 @@ package("gamenetworkingsockets")
...
         end
         os.cp(path.join(package:scriptdir(), "port", "xmake.lua"), "xmake.lua")
         import("package.tools.xmake").install(package, configs)
+
+        os.cp("include/*", package:installdir("include"))
+        os.cp("src/public", package:installdir("include"))
     end)
 
     on_test(function (package)
-        assert(package:has_cfuncs("GameNetworkingSockets_Init", {includes = "steam/steamnetworkingsockets.h"}))
```
waruqi: try 
```lua
        assert(package:has_cfuncs("GameNetworkingSockets_Kill()", {includes = "steam/steamnetworkingsockets.h"}))
```

packages/g/gamenetworkingsockets/xmake.lua
yamashi: As far as I understand fPIC it has to be set on all .so libraries otherwise they cannot be loaded, maybe we should make this automatic?

waruqi:
>C:\Users\runneradmin\AppData\Local\.xmake\cache\packages\2101\g\gamenetworkingsockets\v1.2.0\source\GameNetworkingSockets-1.2.0\build\.gens\gamenetworkingsockets\windows\x86\release\rules\protobuf\src\common\steamnetworkingsockets_messages_certs.pb.cc : fatal error C1083: Cannot open compiler generated file: '': Invalid argument

oh, I found the cause of the problem. Because the target name and source file path of this project are too long, the absolute path exceeds the Windows path length limit.

waruqi:
I have fixed this patch and merged it, see https://github.com/xmake-io/xmake-repo/pull/258

PR #264, opened 2021-01-31
Add qdcae

PucklaMotzer09:
This adds the library [qd-cae-python](https://github.com/qd-cae/qd-cae-python). This generally is a python library, but because of its C++ backend it can be used as a C++ library.


waruqi:
Thanks!

PR #273, opened 2021-02-03
Update xmake.lua

yamashi:


yamashi:
m4 package fails on linux, no idea what it is

waruqi:
I have fixed it. see https://github.com/xmake-io/xmake-repo/pull/274

we should use libsodium-$(version)-msvc.zip instead of libsodium-$(version)-stable-msvc.zip

`xxx-stable-xxx` will be updated frequently

PR #275, opened 2021-02-05
libusb: Add support for windows, version 1.0.24 and fix includes

SirLynix:
I need help with this:
- I need to select the right libusb_*.sln depending on the most recent VS installed.
- What was the purpose of the add_includedirs directive? From what I saw in debian packages you should include libusb-1.0/libusb.h to use it.
- I couldn't find a way to build only the shared or static library so for now I build both and then copy the right files.

### DIFF
--- a/packages/l/libusb/xmake.lua
+++ b/packages/l/libusb/xmake.lua
@@ -12,6 +12,8 @@ package("libusb")
         add_deps("autoconf", "automake", "libtool", "pkg-config")
     end
 
+    add_includedirs("include", "include/libusb-1.0")
+
     on_install("windows", function (package)
         os.cd("msvc")
 
--- a/packages/l/libusb/xmake.lua
+++ b/packages/l/libusb/xmake.lua
@@ -21,7 +21,7 @@ package("libusb")
         local mode = package:debug() and "Debug" or "Release"
 
         import("core.tool.toolchain")
-        local vs = toolchain.load("msvc"):config("vs") or "2019"
+        local vs = string.match(toolchain.load("msvc"):config("vs") or "2019", "(vs%d+)")
         local configs = {"libusb_" .. vs .. " .sln"}
         table.insert(configs, "/property:Configuration=" .. mode)
         table.insert(configs, "/property:Platform=" .. arch)
--- a/packages/l/libusb/xmake.lua
+++ b/packages/l/libusb/xmake.lua
@@ -5,18 +5,41 @@ package("libusb")
 
     set_urls("https://github.com/libusb/libusb/archive/$(version).tar.gz",
              "https://github.com/libusb/libusb.git")
+    add_versions("v1.0.24", "b7724c272dfc5713dce88ff717efd60f021ca5b7c8e30f08ebb2c42d2eea08ae")
     add_versions("v1.0.23", "02620708c4eea7e736240a623b0b156650c39bfa93a14bcfa5f3e05270313eba")
 
     if not is_host("windows") then
         add_deps("autoconf", "automake", "libtool", "pkg-config")
     end
 
-    add_includedirs("include/libusb-1.0")
+    on_install("windows", function (package)
+        os.cd("msvc")
+
+        local configs = {"libusb_2019.sln"}
+        local arch = package:is_arch("x86") and "Win32" or "x64"
+        local mode = package:debug() and "Debug" or "Release"
+
+        table.insert(configs, "/property:Configuration=" .. mode)
+        table.insert(configs, "/property:Platform=" .. arch)
+        --table.insert(configs, "-t:" .. (package:config("shared") and "libusb-1_0%20_dll_" or "libusb-1_0%20_static_"))
+
+        import("package.tools.msbuild").build(package, configs)
+
+        os.cd("..")
+
+        os.vcp("libusb/*.h", package:installdir("include/libusb-1.0"))
+        if (package:config("shared")) then
+            os.vcp(path.join(arch, mode, "dll/libusb-1.0.dll"), package:installdir("lib"))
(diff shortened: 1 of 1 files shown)


packages/l/libusb/xmake.lua
```diff
@@ -5,18 +5,41 @@ package("libusb")
...
     set_urls("https://github.com/libusb/libusb/archive/$(version).tar.gz",
              "https://github.com/libusb/libusb.git")
+    add_versions("v1.0.24", "b7724c272dfc5713dce88ff717efd60f021ca5b7c8e30f08ebb2c42d2eea08ae")
     add_versions("v1.0.23", "02620708c4eea7e736240a623b0b156650c39bfa93a14bcfa5f3e05270313eba")
 
     if not is_host("windows") then
         add_deps("autoconf", "automake", "libtool", "pkg-config")
     end
 
-    add_includedirs("include/libusb-1.0")
```
waruqi, line 14: > What was the purpose of the add_includedirs directive? From what I saw in debian packages you should include libusb-1.0/libusb.h to use it.

export includedirs to user projects, user may will use `#include <libusb.h>`

we can use this:

```lua
    add_includedirs("include", "include/libusb-1.0")
```

packages/l/libusb/xmake.lua
```diff
@@ -5,18 +5,41 @@ package("libusb")
...
+    on_install("windows", function (package)
+        os.cd("msvc")
+
+        local configs = {"libusb_2019.sln"}
```
waruqi, line 18:
```lua
import("core.tool.toolchain")
local vs = toolchain.load("msvc"):config("vs") or "2019"
local configs = {"libusb_" .. vs .. " .sln"}
```

packages/l/libusb/xmake.lua
```diff
@@ -5,18 +5,41 @@ package("libusb")
...
+        local arch = package:is_arch("x86") and "Win32" or "x64"
+        local mode = package:debug() and "Debug" or "Release"
+
+        table.insert(configs, "/property:Configuration=" .. mode)
+        table.insert(configs, "/property:Platform=" .. arch)
+        --table.insert(configs, "-t:" .. (package:config("shared") and "libusb-1_0%20_dll_" or "libusb-1_0%20_static_"))
```
waruqi, line 24: Does it not work? `libusb_dll_2019.vcxproj` `libusb_static_2019.vcxproj`

libusb_dll_2019? libusb_static_2019?

packages/l/libusb/xmake.lua
```diff
@@ -5,18 +5,41 @@ package("libusb")
...
+
+        import("package.tools.msbuild").build(package, configs)
+
+        os.cd("..")
+
+        os.vcp("libusb/*.h", package:installdir("include/libusb-1.0"))
+        if (package:config("shared")) then
```
waruqi, line 31: please remove `(..)`

packages/l/libusb/xmake.lua
SirLynix: I don't recall the error but it didn't work

packages/l/libusb/xmake.lua
SirLynix: Thanks!

packages/l/libusb/xmake.lua
SirLynix: Why do we do it for this library and not other?
From what I see in the debian package, it's installed in a libusb-1.0 folder, which means #include <libusb.h> isn't valid.
```
:~$ dpkg -L libusb-1.0-0-dev
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/libusb-1.0-0-dev
/usr/share/doc/libusb-1.0-0-dev/copyright
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libusb-1.0.a
/usr/lib/x86_64-linux-gnu/pkgconfig
/usr/lib/x86_64-linux-gnu/pkgconfig/libusb-1.0.pc
/usr/include
/usr/include/libusb-1.0
/usr/include/libusb-1.0/libusb.h
/usr/share/doc/libusb-1.0-0-dev/README
/usr/share/doc/libusb-1.0-0-dev/changelog.Debian.gz
```

waruqi: It's official examples: https://github.com/libusb/libusb/blob/a108bb8e185cc5c41b296d35dfa83e5e5b13b2c0/examples/listdevs.c#L22

and homebrew 

https://github.com/Homebrew/linuxbrew-core/blob/27ce9737e78d9144910f14d6b15d2ff4055915a4/Formula/libusb.rb#L43

Many users may use `<libusb.h>`

`add_includedirs("include", "include/libusb-1.0")` can support `<libusb.h> and <libusb-1.0/libusb.h>`

SirLynix: Alrighty then

packages/l/libusb/xmake.lua
```diff
@@ -21,7 +21,7 @@ package("libusb")
         local mode = package:debug() and "Debug" or "Release"
 
         import("core.tool.toolchain")
-        local vs = string.match(toolchain.load("msvc"):config("vs") or "2019", "(vs%d+)")
+        local vs = string.match(toolchain.load("msvc"):config("vs") or "", "^(%d+)") or "2019"
```
waruqi, line 24: we need not use `match`, it is always number, "201x"

packages/l/libusb/xmake.lua
SirLynix: Are you sure about that? It doesn't find files now

waruqi:
see https://github.com/xmake-io/xmake-repo/pull/278

PR #276, opened 2021-02-05
CEF3 support for windows

yamashi:
I don't have the need for linux, mac etc so I haven't bothered implementing it.


yamashi:
Again paths are too long for windows...

waruqi:
I added it, https://github.com/xmake-io/xmake-repo/pull/279

But it only support MT and static library now, because Release/cef_sandbox.lib use MT.

PR #281, opened 2021-02-06
Add libuv version 1.40.0

yamashi:


waruqi:
I have fixed and merged this patch. https://github.com/xmake-io/xmake-repo/pull/282

PR #284, opened 2021-02-11
Add libsoundio

SirLynix:


SirLynix:
It seems to error on Windows because the Visual Studio version used by the CI doesn't support C11.
Maybe it would be possible to compile it as C++, but I'm not sure it would be a good idea.

waruqi:
Which version of vs can support it? Can it be compiled normally on your machine? Maybe the high version vs. can be set on ci, but I am not sure

SirLynix:
Yes it does compile on my machine.

C11 support with MSVC is quite recent (only a few months):
https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

There's an issue on libsoundio where it says "just compile it as C++ when using MSVC":
https://github.com/andrewrk/libsoundio/issues/179

waruqi:
> andrewrk/libsoundio#179

xmake has done similar work. If msvc does not support c99/c11, `set_languages("c99")` will automatically set `/TP` to compile the code as c++11.

waruqi:
> C11 support with MSVC is quite recent (only a few months):

Although I can consider to enable the windows version of this package directly, I think for other users, this package will still be interrupted.

packages/l/libsoundio/xmake.lua
```diff
@@ -5,12 +5,16 @@ package("libsoundio")
...
              "https://github.com/andrewrk/libsoundio.git")
+
     add_versions("2.0.0", "67a8fc1c9bef2b3704381bfb3fb3ce99e3952bc4fea2817729a7180fddf4a71e")
+    add_patches("2.0.0", path.join(os.scriptdir(), "patches", "2.0.0", "extern.patch"), "c1c9084012cf39b203cdf7c59d46a8e6362b6a1e7203700d77faef529d49d401")
 
     add_deps("cmake")
 
     add_includedirs("include", "include/soundio")
 
+    set_languages("c11")
```
waruqi, line 16: It only for xmake.lua. If it is a cmake package, you need to pass `/TP` and `/std:c++11` to cmake to force it to be compiled as c++11

```lua
import("package.tools.cmake").install(package, configs, {cxflags = {"/TP", "/std:c++11"}})

SirLynix: Thanks. But it seems to ignore /TP as it's still compiling in C mode.
I tried /TP locally (to test my patch) and it did work.

waruqi: But stdatomic.h is missing in older vs

waruqi: may be we need patch a stdatomic.h

SirLynix: stdatomic.h shouldn't be required in C++ mode

waruqi: I have used xmake.lua to port it and it works on windows. 

But it only support to enable coreaudio backend now. 

If you want to support other backends, you may need to make some improvements. But I'm going to sleep, you can try to continue to improve it on the libsoundio branch. https://github.com/xmake-io/xmake-repo/blob/libsoundio/packages/l/libsoundio/xmake.lua

waruqi:
I have used xmake.lua to port it and it works on windows.

But it only support to enable coreaudio backend now.

If you want to support other backends, you may need to make some improvements. But I'm going to sleep, you can try to continue to improve it on the libsoundio branch. https://github.com/xmake-io/xmake-repo/blob/libsoundio/packages/l/libsoundio/xmake.lua

waruqi:
I merged it. https://github.com/xmake-io/xmake-repo/pull/285

SirLynix:
Thank you.

PR #286, opened 2021-02-12
Bullet3 - add `msvc_runtime_dll` configuration

SoapyMan:
Update for the existing package.
Adding runtime change for Bullet3 library.
As there is still no `is_runtimes()` or such, it's been made as manual switch.


SoapyMan:
Apparently, this switch does not work. Waiting for this issue to be fixed and package updated: https://github.com/bulletphysics/bullet3/issues/3268

PR #293, opened 2021-02-21
Make snappy position independant

yamashi:


packages/s/snappy/xmake.lua
```diff
@@ -18,6 +18,7 @@ package("snappy")
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DSNAPPY_REQUIRE_AVX=" .. (package:config("avx") and "ON" or "OFF"))
         table.insert(configs, "-DSNAPPY_REQUIRE_AVX2=" .. (package:config("avx2") and "ON" or "OFF"))
+        table.insert(configs, "-DCMAKE_POSITION_INDEPENDENT_CODE=ON")
```
waruqi, line 21: We should judge the pic configuration and provide users with the opportunity to disable the pic.

like this.

```lua
        if package:config("pic") ~= false then
            table.insert(configs, "-DCMAKE_POSITION_INDEPENDENT_CODE=ON")
        end
```

I have modified it.

PR #298, opened 2021-02-23
Add entt

yamashi:


packages/e/entt/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+    set_homepage("https://github.com/skypjack/entt")
+    set_description("Gaming meets modern C++ - a fast and reliable entity component system (ECS) and much more.")
+
+    set_urls("https://github.com/skypjack/entt/archive/$(version).tar.gz",
+             "https://github.com/skypjack/entt.git")
+
+    add_versions("v3.6.0", "94b7dc874acd0961cfc28cf6b0342eeb0b7c58250ddde8bdc6c101e84b74c190")
+    add_deps("cmake")
+
+    on_install("*", function (package)
```
waruqi, line 12: please remove "*"

packages/e/entt/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+        local configs = {}
```
waruqi, line 13: and shared/static library 

```lua
```

packages/e/entt/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+    set_homepage("https://github.com/skypjack/entt")
+    set_description("Gaming meets modern C++ - a fast and reliable entity component system (ECS) and much more.")
+
+    set_urls("https://github.com/skypjack/entt/archive/$(version).tar.gz",
+             "https://github.com/skypjack/entt.git")
+
+    add_versions("v3.6.0", "94b7dc874acd0961cfc28cf6b0342eeb0b7c58250ddde8bdc6c101e84b74c190")
+    add_deps("cmake")
+
+    on_install("", function (package)
```
waruqi, line 12: we only need `on_install(function (package)`, if this package support all platforms

packages/e/entt/xmake.lua
yamashi: It's header only

packages/e/entt/xmake.lua
waruqi: need not `""`

packages/e/entt/xmake.lua
waruqi: ok

PR #303, opened 2021-02-26
Fix nana on linux

PucklaMotzer09:
Adds a patch which replaces ```u8string``` with ```string```. It also copies all headers and libraries and adds nana as a link.

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -25,6 +25,7 @@ jobs:
           dnf -y upgrade git
           dnf -y install perl
           dnf -y install libffi-devel
+          dnf -y install zlib-devel
       - uses: actions/checkout@v1
       - uses: xmake-io/github-action-setup-xmake@v1
         with:
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -24,6 +24,7 @@ jobs:
           dnf -y install copr-cli make gcc-c++
           dnf -y upgrade git
           dnf -y install perl
+          dnf -y install libffi-devel
       - uses: actions/checkout@v1
       - uses: xmake-io/github-action-setup-xmake@v1
         with:
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -26,6 +26,7 @@ jobs:
           dnf -y install perl
           dnf -y install libffi-devel
           dnf -y install zlib-devel
+          dnf -y install libxft-dev
       - uses: actions/checkout@v1
       - uses: xmake-io/github-action-setup-xmake@v1
         with:
--- a/packages/n/nana/xmake.lua
+++ b/packages/n/nana/xmake.lua
@@ -34,6 +34,7 @@ package("nana")
     end)
 
     on_install("linux", "windows", function (package)
+        -- The 'and' operator, which is an equivalent of '&&', is not supported by MSVC
         if package:is_plat("windows") then
             local file_name = path.join(os.curdir(), "source", "system", "split_string.cpp")
             io.gsub(file_name, " and ", " && ")
@@ -42,18 +43,6 @@ package("nana")
         local configs = {"-DNANA_CMAKE_ENABLE_JPEG=OFF", "-DNANA_CMAKE_ENABLE_PNG=OFF", "-DBUILD_SHARED_LIBS=OFF"}
         if package:config("nana_filesystem_force") then
             table.insert(configs, "-DNANA_CMAKE_NANA_FILESYSTEM_FORCE=ON")
-        end
-        if package:is_plat("windows") then
-            local cmake_vsr = "MultiThreaded"
-            local pvsr = package:config("vs_runtime")
-            if pvsr == "MTd" then
-                cmake_vsr = "MultiThreadedDebug"
-            elseif pvsr == "MD" then
-                cmake_vsr = "MultiThreadedDLL"
-            elseif pvsr == "MDd" then
-                cmake_vsr = "MultiThreadedDebugDLL"
-            end
-            table.insert(configs, "-DMSVC_RUNTIME_LIBRARY=" .. cmake_vsr)
         end
         import("package.tools.cmake").build(package, configs, {buildir = "build_xmake"})
 
--- a/packages/n/nana/xmake.lua
+++ b/packages/n/nana/xmake.lua
@@ -20,7 +20,8 @@ package("nana")
         add_syslinks("ole32", "shell32", "kernel32", "user32", "gdi32", "winspool", "comdlg32", "advapi32")
         add_defines("_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING")
     elseif is_plat("linux") then
-        add_syslinks("pthread", "X11", "Xft", "fontconfig")
+        add_deps("libxcursor")
+        add_syslinks("pthread", "fontconfig", "Xft")
         add_links("nana")
     end
 
--- a/packages/n/nana/xmake.lua
+++ b/packages/n/nana/xmake.lua
@@ -26,7 +26,6 @@ package("nana")
         add_deps("libxcursor")
         add_syslinks("pthread", "fontconfig", "Xft")
     end
-    add_links("nana")
 
     on_load("linux", "windows", function (package)
         if package:config("nana_filesystem_force") then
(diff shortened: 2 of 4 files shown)


packages/n/nana/xmake.lua
```diff
@@ -18,6 +21,7 @@ package("nana")
         add_defines("_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING")
     elseif is_plat("linux") then
         add_syslinks("pthread", "X11", "Xft", "fontconfig")
+        add_links("nana")
```
waruqi, line 24: we need not add it, xmake will find links in lib directory

SirLynix, line 24: ?

packages/n/nana/xmake.lua
```diff
@@ -32,8 +36,9 @@ package("nana")
         end
         import("package.tools.cmake").install(package, configs)
 
-        if is_plat("windows") then
-            os.cp("include", package:installdir())
+        os.cp("include", package:installdir())
+        if is_plat("linux") then
```
waruqi, line 40: package:is_plat

packages/n/nana/xmake.lua
```diff
@@ -32,8 +36,9 @@ package("nana")
...
+            os.cp("**.a", package:installdir("lib"))
```
waruqi, line 41: why? cmake.install did not copy them? 

and so for shared

packages/n/nana/xmake.lua
PucklaMotzer09: Yes it should do this. But for some reason it didn't on my system

packages/n/nana/xmake.lua
PucklaMotzer09: Yes it didn't copy any files

packages/n/nana/xmake.lua
PucklaMotzer09: That's also what I think

packages/n/nana/xmake.lua
PucklaMotzer09: In [this document](https://github.com/qPCR4vir/nana-docs/wiki/Installation) they only write about static linking. So I think a shared build is not supported

packages/n/nana/xmake.lua
waruqi: If the .a library exists in the installdir/lib directory, you should be able to find this links

We should not add it. If you can't find links on your machine, you should open a bug issue.

packages/n/nana/xmake.lua
waruqi: you can pass build path to cmake

and copy build/xx/.a files 

see https://github.com/xmake-io/xmake-repo/blob/7b44576aa1b54d67a04e4088e816d6a7143c4f0c/packages/b/box2d/xmake.lua#L17

Try to avoid recursive pattern matching

packages/n/nana/xmake.lua
PucklaMotzer09: cmake creates nothing inside the install directory. I think nana doesn't support an install prefix.

PucklaMotzer09:
Now there is another problem with linux:
```
error: @programdir/core/main.lua:294: ...mdir/core/sandbox/modules/import/core/sandbox/module.lua:431: cannot import module: package.manager.pkg_config.find_package, not found!
```

packages/n/nana/xmake.lua
```diff
@@ -8,32 +8,61 @@ package("nana")
...
+            local pvsr = package:config("vs_runtime")
+            local d = package:config("debug")
+            if pvsr == "MT" and d then
+                cmake_vsr = "MultiThreadedDebug"
+            elseif pvsr == "MD" and not d then
+                cmake_vsr = "MultiThreadedDLL"
+            elseif pvsr == "MD" and d then
+                cmake_vsr = "MultiThreadedDebugDLL"
+            end
+            table.insert(configs, "-DMSVC_RUNTIME_LIBRARY=" .. cmake_vsr)
```
waruqi, line 57: we need not add them, xmake will pass vs runtime 

https://github.com/xmake-io/xmake/blob/8a1528fdb97a228ae62113d3b6c1b4e0a3e19381/xmake/modules/package/tools/cmake.lua#L231-L246

waruqi:
> Now there is another problem with linux:
> 
> ```
> error: @programdir/core/main.lua:294: ...mdir/core/sandbox/modules/import/core/sandbox/module.lua:431: cannot import module: package.manager.pkg_config.find_package, not found!
> ```

I have fixed it.

packages/n/nana/xmake.lua
PucklaMotzer09: I know. I found the code for it, but why are we getting the errors?

packages/n/nana/xmake.lua
```diff
@@ -8,32 +8,60 @@ package("nana")
...
+    on_load("linux", "windows", function (package)
+        if package:config("nana_filesystem_force") then
+            package:add("defines", "NANA_FILESYSTEM_FORCE")
+        end
+    end)
 
     on_install("linux", "windows", function (package)
-        if is_plat("windows") then
+        if package:is_plat("windows") then
             local file_name = path.join(os.curdir(), "source", "system", "split_string.cpp")
```
waruqi, line 38: Please add some comments to explain what problem is fixed

packages/n/nana/xmake.lua
PucklaMotzer09: I think I know why the CMAKE_MSVC_RUNTIME_LIBRARY macro doesn't work. On [this page](https://cmake.org/cmake/help/latest/policy/CMP0091.html#policy:CMP0091) at the bottom it states, that if the policy is not explicitly set it uses the OLD behaviour, which is ignoring the macro.

PucklaMotzer09:
There is a problem with Fedora:
```
error: @programdir/core/main.lua:294: /__w/xmake-repo/xmake-repo/packages/p/python/xmake.lua:65: libffi-dev not found, please use your system package manager to install it.
```

packages/n/nana/xmake.lua
waruqi: I tested it, and even if this link is not set, xmake can automatically find the links.

As long as you copy libnana.a to the lib directory, xmake will find it automatically, you don’t need to add `add_links("nana")`

It was not found on your machine before, maybe you haven't added `os.cp("*.a", "lib")` at the beginning, so there is no library file in the lib directory, so xmake didn't add nana links

waruqi:
> There is a problem with Fedora:
> 
> ```
> error: @programdir/core/main.lua:294: /__w/xmake-repo/xmake-repo/packages/p/python/xmake.lua:65: libffi-dev not found, please use your system package manager to install it.
> ```



> There is a problem with Fedora:
> 
> ```
> error: @programdir/core/main.lua:294: /__w/xmake-repo/xmake-repo/packages/p/python/xmake.lua:65: libffi-dev not found, please use your system package manager to install it.
> ```

Fedora is a new ci that I added later, so this package has not been tested on it before.

There are still some dependency issues in the python package that have not been resolved for the time being, some system libraries need to be added, for example: libffi-dev

I will try to continue improving the python package in the next few days. At present, you can only ignore this error temporarily, or try to manually install libffi-dev on fedora.yml temporarily. Remove it until ci passes.

Of course, if you are interested, you can also help improve python. = =

packages/n/nana/xmake.lua
PucklaMotzer09: I now added a patch which sets the policy and now all windows builds work. Maybe there is a way to fix this globally

packages/n/nana/xmake.lua
PucklaMotzer09: You are right. I removed it

packages/n/nana/xmake.lua
waruqi: Global settings, there will be compatibility issues, not all libraries can support this new policy, there may be some conflicts.

We can only turn on this policy for specific packages.

.github/workflows/fedora.yml
```diff
@@ -24,6 +24,8 @@ jobs:
           dnf -y install copr-cli make gcc-c++
           dnf -y upgrade git
           dnf -y install perl
+          dnf -y install libffi-devel
+          dnf -y install zlib-devel
```
waruqi, line 28: maybe need install xft package on fedora. xmake-repo has not added this package yet

waruqi: or we can add libxft in xmake-repo. It seems it is not very complicated

https://github.com/Homebrew/linuxbrew-core/blob/master/Formula/libxft.rb

waruqi:
I have merged this patch and added libxft to support fedora

PR #306, opened 2021-02-28
Add tclap package

AntoineJT:
This commit adds tclap, a templatized C++ library for parsing commandline arguments.

I'm not sure if `-DBUILD_DOC=false` really disable documentation building since I don't have doxygen installed on my machine.
I don't know how to patch the CMakeLists.txt file, but it would be good to avoid useless computation by building examples.

I have tested it and added python3 as a dependency (I don't know if this information is relevant but here it is: I've got python 3.9.2 installed on my machine, the latest in the repo is 3.8.5, and so xmake downloaded python 3.8.5) because it is required to run the tests.

I don't know, maybe we don't want the tests to run, and in this case, we could later patch the CMakeLists.txt file by removing the following lines:
```
add_subdirectory(docs)
add_subdirectory(examples)
add_subdirectory(tests)
```
If we do that, we could remove the `-DBUILD_DOC=false` line in the package script.

### DIFF
--- a/packages/t/tclap/xmake.lua
+++ b/packages/t/tclap/xmake.lua
@@ -6,15 +6,14 @@ package("tclap")
     set_urls("https://netcologne.dl.sourceforge.net/project/tclap/tclap-$(version).tar.bz2")
     add_versions("1.4.0-rc1", "33e18c7828f76a9e5f2a00afe575156520e383693059ca9bc34ff562927e20c6")
 
-    -- NOTE it could be useful to patch CMakeLists.txt
-    -- in order to avoid building examples
-
-    -- python3 is required by the tests module
-    add_deps("cmake", "python 3.*")
+    add_deps("cmake")
 
     on_install(function (package)
-        -- We don't want to build the doc
-        local configs = {"-DBUILD_DOC=false"}
+        io.replace("CMakeLists.txt", "add_subdirectory(docs)", "", {plain = true})
+        io.replace("CMakeLists.txt", "add_subdirectory(examples)", "", {plain = true})
+        io.replace("CMakeLists.txt", "add_subdirectory(tests)", "", {plain = true})
+
+        local configs = {}
         import("package.tools.cmake").install(package, configs)
     end)
 
--- a/packages/t/tclap/xmake.lua
+++ b/packages/t/tclap/xmake.lua
@@ -1,0 +1,28 @@
+package("tclap")
+    set_homepage("https://sourceforge.net/projects/tclap/")
+    set_description("This is a simple templatized C++ library for parsing command line arguments.")
+    set_license("MIT")
+
+    set_urls("https://netcologne.dl.sourceforge.net/project/tclap/tclap-$(version).tar.bz2")
+    add_versions("1.4.0-rc1", "33e18c7828f76a9e5f2a00afe575156520e383693059ca9bc34ff562927e20c6")
+
+    -- NOTE it could be useful to patch CMakeLists.txt
+    -- in order to avoid building examples
+
+    -- python3 is required by the tests module
+    add_deps("cmake", "python 3.*")
+
+    on_install(function (package)
+        -- We don't want to build the doc
+        local configs = {"-DBUILD_DOC=false"}
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            #include <tclap/CmdLine.h>
+            void test() {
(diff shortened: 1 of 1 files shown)


packages/t/tclap/xmake.lua
```diff
@@ -0,0 +1,28 @@
...
+
+    -- NOTE it could be useful to patch CMakeLists.txt
+    -- in order to avoid building examples
+
+    -- python3 is required by the tests module
+    add_deps("cmake", "python 3.*")
+
+    on_install(function (package)
+        -- We don't want to build the doc
+        local configs = {"-DBUILD_DOC=false"}
```
waruqi, line 17: `-DBUILD_DOC=OFF`

packages/t/tclap/xmake.lua
```diff
@@ -0,0 +1,28 @@
...
+    add_versions("1.4.0-rc1", "33e18c7828f76a9e5f2a00afe575156520e383693059ca9bc34ff562927e20c6")
+
+    -- NOTE it could be useful to patch CMakeLists.txt
+    -- in order to avoid building examples
+
+    -- python3 is required by the tests module
+    add_deps("cmake", "python 3.*")
+
+    on_install(function (package)
+        -- We don't want to build the doc
```
waruqi, line 16: Please remove unnecessary comments.

packages/t/tclap/xmake.lua
```diff
@@ -0,0 +1,28 @@
...
+    set_license("MIT")
+
+    set_urls("https://netcologne.dl.sourceforge.net/project/tclap/tclap-$(version).tar.bz2")
+    add_versions("1.4.0-rc1", "33e18c7828f76a9e5f2a00afe575156520e383693059ca9bc34ff562927e20c6")
+
+    -- NOTE it could be useful to patch CMakeLists.txt
+    -- in order to avoid building examples
+
+    -- python3 is required by the tests module
+    add_deps("cmake", "python 3.*")
```
waruqi, line 13: please disable tests module and remove python deps

packages/t/tclap/xmake.lua
```diff
@@ -0,0 +1,28 @@
+package("tclap")
+    set_homepage("https://sourceforge.net/projects/tclap/")
+    set_description("This is a simple templatized C++ library for parsing command line arguments.")
+    set_license("MIT")
+
+    set_urls("https://netcologne.dl.sourceforge.net/project/tclap/tclap-$(version).tar.bz2")
+    add_versions("1.4.0-rc1", "33e18c7828f76a9e5f2a00afe575156520e383693059ca9bc34ff562927e20c6")
+
+    -- NOTE it could be useful to patch CMakeLists.txt
+    -- in order to avoid building examples
```
waruqi, line 10: we can also use `io.gsub` or `io.replace` to patch CMakeLists.txt. for example

```lua
packages/l/libtiff/xmake.lua
18:        io.replace("CMakeLists.txt", "add_subdirectory(contrib)", "", {plain = true})
19:        io.replace("CMakeLists.txt", "add_subdirectory(man)", "", {plain = true})
20:        io.replace("CMakeLists.txt", "add_subdirectory(html)", "", {plain = true})
21:        io.replace("CMakeLists.txt", "add_subdirectory(test)", "", {plain = true})
22:        io.replace("CMakeLists.txt", "add_subdirectory(tools)", "", {plain = true})
```

packages/t/tclap/xmake.lua
waruqi: and we need add shared libraries

```lua
```

AntoineJT: It's an header only library

packages/t/tclap/xmake.lua
AntoineJT: I don't know how to patch CMakeLists.txt file
I saw it has a `add_patches` function in xmake but I don't know how to give the path of the file to this

packages/t/tclap/xmake.lua
AntoineJT: Thank you, I will do that

waruqi:
It’s weird that the source directory unzipped on msys2 is empty, I need to take a moment to look at it

AntoineJT:
Isn't this related to the fact it has a top level folder in the zip?
But for some reason it doesn't work on my machine when prefixing CMakeLists.txt with `"tclap-".. package:version()` :shrug:

waruqi:
This is because tar on windows does not support decompressing .tar.bz2, I have improve xmake to use bzip2 to extract it.

PR #308, opened 2021-02-28
CEF - 32 bits windows support and 88.2.9

yamashi:


waruqi:
You can temporarily modify ci to test static with MT only, and when ci passes, I will revert ci patch and merge it.

```
      matrix:
        os: [windows-latest]
        kind: [static]
        arch: [x64, x86]
        vs_runtime: [MT]
```

PR #335, opened 2021-03-24
Add godotcpp

PucklaMotzer09:
Adds [godotcpp](https://github.com/godotengine/godot-cpp). This also removes [godotheaders](https://github.com/godotengine/godot-headers), because this repo is intended to be used with godotcpp.


PucklaMotzer09:
**android** is failing, because it wants to link to **z**, **ssl** and **crypto**. These are dependencies of python and they get applied to **godotcpp** through **scons**. But **godotcpp** doesn't need these libraries. So they need to somehow get removed. Could this be implemented so that all dependencies of a binary package don't get applied to its children?

waruqi:
> **android** is failing, because it wants to link to **z**, **ssl** and **crypto**. These are dependencies of python and they get applied to **godotcpp** through **scons**. But **godotcpp** doesn't need these libraries. So they need to somehow get removed. Could this be implemented so that all dependencies of a binary package don't get applied to its children?

You can disable android package first and open a issue on github, I will fix it later.

packages/g/godotcpp/xmake.lua
```diff
@@ -0,0 +1,96 @@
...
+        local configs = {
+            "platform=" .. scons_plat,
+            "bits=" .. scons_bits,
+            "generate_bindings=yes",
+            "target=" .. (package:debug() and "debug" or "release"),
+            "use_mingw=" .. (package:is_plat("mingw", "cygwin", "msys") and "yes" or "no"),
+        }
+
+        if package:is_plat("android") then
+            table.insert(configs, "android_arch=" .. scons_android_arch)
```
waruqi, line 47: I have added this configs to tools/scons

we need only write:

```lua

    on_install("linux", "windows", "macosx", "mingw", "cygwin", "iphoneos", "msys", "android", function (package)
        local configs = {"generate_bindings=yes"}
            io.gsub("SConstruct", "/MD", "/" .. package:config("vs_runtime"))
        end
        -- this fixes an error on ios and osx (https://godotengine.org/qa/65616/problems-compiling-gdnative-c-example-on-osx)
        if package:is_plat("macosx", "iphoneos") then
            io.gsub("SConstruct", "(-std=c%+%+14)", "-std=c++17")
        end

        import("package.tools.scons").build(package, configs)
        os.cp("bin/*." .. (package:is_plat("windows") and "lib" or "a"), package:installdir("lib"))
        os.cp("include/core/*.hpp", package:installdir("include/core"))
        os.cp("include/gen/*.hpp",  package:installdir("include/gen"))
        os.cp("godot-headers/android",            package:installdir("include"))
        os.cp("godot-headers/arvr",               package:installdir("include"))
        os.cp("godot-headers/gdnative",           package:installdir("include"))
        os.cp("godot-headers/nativescript",       package:installdir("include"))
        os.cp("godot-headers/net",                package:installdir("include"))
        os.cp("godot-headers/pluginscript",       package:installdir("include"))
        os.cp("godot-headers/videodecoder",       package:installdir("include"))
        os.cp("godot-headers/*.h",                package:installdir("include"))
    end)
```

packages/g/godotcpp/xmake.lua
```diff
@@ -0,0 +1,96 @@
...
+        elseif package:is_plat("iphoneos") then
+            table.insert(configs, "ios_arch=" .. scons_ios_arch)
+        elseif package:is_plat("windows") then
+            io.gsub("SConstruct", "/MD", "/" .. package:config("vs_runtime"))
+        end
+        -- this fixes an error on ios and osx (https://godotengine.org/qa/65616/problems-compiling-gdnative-c-example-on-osx)
+        if package:is_plat("macosx", "iphoneos") then
+            io.gsub("SConstruct", "(-std=c%+%+14)", "-std=c++17")
```
waruqi, line 55:
```lua
            io.replace("SConstruct", "-std=c++14", "-std=c++17", {plain = true})
```

waruqi:
Why not use cmake? https://github.com/godotengine/godot-cpp/blob/master/CMakeLists.txt

There are more dependencies when using scons, such as python/openssl ...

PR #339, opened 2021-03-29
Update imgui to v1.82

WSSDude420:

### DIFF
--- a/packages/i/imgui/xmake.lua
+++ b/packages/i/imgui/xmake.lua
@@ -12,6 +12,10 @@ package("imgui")
     add_versions("v1.79", "f1908501f6dc6db8a4d572c29259847f6f882684b10488d3a8d2da31744cd0a4")
     add_versions("v1.75", "1023227fae4cf9c8032f56afcaea8902e9bfaad6d9094d6e48fb8f3903c7b866")
 
+    if is_plat("windows", "mingw") then
+        add_syslinks("Imm32")
+    end
+
     on_install("macosx", "linux", "windows", "mingw", "android", "iphoneos", function (package)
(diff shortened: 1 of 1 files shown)


waruqi:
> checkinfo: ...gramdir\core\sandbox\modules\import\core\tool\linker.lua:75: @programdir\core\sandbox\modules\os.lua:259: D:/a/_temp/msys/msys64/mingw64/lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\runneradmin\AppData\Local\.xmake\packages\i\imgui\v1.82\bc477f71d6e746518a0511230c2d167d\lib/imgui.lib(imgui.cpp.obj):imgui.cpp:(.text+0x22b2f): undefined reference to `ImmGetContext'

You need add Imm32 syslinks for new version

```lua
if is_plat("windows", "mingw") then
    add_syslinks("Imm32")
end
```

packages/i/imgui/xmake.lua
```diff
@@ -17,6 +18,10 @@ package("imgui")
                 set_kind("static")
                 add_files("*.cpp")
                 add_headerfiles("imgui.h", "imconfig.h")
+
+                if is_plat("windows", "mingw") then
+                    add_syslinks("Imm32")
+                end
```
waruqi, line 24: please move to outside.

```lua
add_versions(...)

if is_plat("windows", "mingw") then
    add_syslinks("Imm32")
end
```

WSSDude420: I was about to ask if that is the right place. Moved it :)

PR #341, opened 2021-03-29
Update spdlog to 1.8.5

WSSDude420:


SirLynix:
The version tag should begin with the v (and it should be removed from url), also the correct checksum is `6e66c8ed4c014b0fb00c74d34eea95b5d34f6e4b51b746b1ea863dc3c2e854fd`

WSSDude420:
I used PowerShell's Get-FileHash for this, it outputed that one. It seems to be correct, dont know how you got to that one?

I quickly checked with other version listed here, hashes it produces seems correct.

Can change tags in versions I suppose, didnt want to do to many changes :D

SirLynix:
> I used PowerShell's Get-FileHash for this, it outputed that one. It seems to be correct, dont know how you got to that one?

Got it using the CRC-SHA option from 7zip.
The SHA256 you put in the version list is the .tar.gz one, but this package downloads the .zip file (which has the checksum I put above).

WSSDude420:
Oh right, missed it downloaded zip here.

Thanks!

WSSDude420:
Welp, guess I'll commit right hash with next PR then.

SirLynix:
@waruqi The last version checksum is incorrect, the CI works because it does fallback on the git.

waruqi:
> @waruqi The last version checksum is incorrect, the CI works because it does fallback on the git.

ok, wait next pr. @WSSDude420

PR #381, opened 2021-04-24
Add NewtonDynamics

SirLynix:


SirLynix:
I'm beginning to think it would be easier to make a xmake.lua file to compile it.

packages/n/newtondynamics/xmake.lua
```diff
@@ -0,0 +1,72 @@
...
+    end)
+
+    on_install("windows", "linux", "macosx", "mingw", "android", function (package)
+        -- prevent cmakelists.txt from overwriting CMAKE_INSTALL_PREFIX
+        io.replace("CMakeLists.txt", [[set(CMAKE_INSTALL_PREFIX "win64sdk" CACHE PATH "..." FORCE)]], "", {plain = true})
+        io.replace("CMakeLists.txt", [[set(CMAKE_INSTALL_PREFIX "win32sdk" CACHE PATH "..." FORCE)]], "", {plain = true})
+
+        local configs = {"-DNEWTON_BUILD_SANDBOX_DEMOS=OFF", "-DNEWTON_BUILD_CORE_ONLY=ON"}
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.insert(configs, "-DDNEWTON_BUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
```
waruqi, line 51: `-DNEWTON_BUILD_SHARED_LIBS `, not `-DD`

waruqi:
> libdgCore.so: undefined reference to `pthread_create'
/usr/bin/ld: /home/ruki/.xmake/packages/n/newtondynamics/v3.14c/6a5e3ff14e21470782bab40bfd1b098d/lib/libdgPhysics.so: undefined reference to `dlopen'

if is shared then
    Maybe we should patch CMakelist.txt to add pthread link to -ldgCore target and add -ldl to dgPhysics target.
end

waruqi:
> I'm beginning to think it would be easier to make a xmake.lua file to compile it.

maybe, you can try it.

packages/n/newtondynamics/xmake.lua
SirLynix: 😅

packages/n/newtondynamics/port/xmake.lua
```diff
@@ -0,0 +1,55 @@
...
+        end
+        if is_plat("android") then
+            add_defines("_ARM_VER")
+        end
+    elseif is_plat("macosx", "iphoneos") then
+        add_defines("_MACOSX_VER")
+    end
+
+    if is_plat("windows") then
+        if get_config("kind") == "static" then
```
waruqi, line 35: is_kind("static")

SirLynix: Fixed.
Also, is it normal that this define, while being public, is not included in the package info? I had to use on_load in the package to have it.

waruqi: If you need to export definitions to users, you should set it in the external package scope through add_defines or set it in on_load.

The definition in port/xmake.lua only affects the compilation of the package itself, and does not export the definition to the user project.

waruqi:
android not add pthread links

PR #394, opened 2021-05-04
Improve libsdl package

SirLynix:
I added a on_fetch function which tries to use the sdl2-config tool and fallbacks on pkg-config

I also added an option to allow disable linking to SDL2main (which then automatically defines SDL_MAIN_HANDLED)

### DIFF
--- a/packages/l/libsdl/xmake.lua
+++ b/packages/l/libsdl/xmake.lua
@@ -34,6 +34,10 @@ package("libsdl")
             package:add("links", "SDL2")
             package:add("defines", "SDL_MAIN_HANDLED")
         end
+
+        if package:is_plat("macosx") and package:version():ge("2.0.14") then
+            package:add("frameworks", "CoreHaptics", "GameController")
+        end
     end)
 
     on_fetch("linux", "macosx", function (package, opt)
@@ -119,12 +123,6 @@ package("libsdl")
             table.insert(configs, "--enable-shared=no")
         end
         import("package.tools.autoconf").install(package, configs)
-    end)
-
-    on_load(function (package)
-        if package:is_plat("macosx") and package:version():ge("2.0.14") then
-            package:add("frameworks", "CoreHaptics", "GameController")
-        end
     end)
 
     on_test(function (package)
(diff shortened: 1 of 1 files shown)


SirLynix:
I don't get why it's linking "SDL2_test" now.

packages/l/libsdl/xmake.lua
```diff
@@ -23,9 +23,76 @@ package("libsdl")
...
+            local sdl2conf = try {function() return os.iorunv("sdl2-config", {"--version", "--cflags", "--libs"}) end}
+            if sdl2conf then
+                sdl2conf = os.argv(sdl2conf)
+                local sdl2ver = table.remove(sdl2conf, 1)
+                if opt.require_version then
+                    import("core.base.semver")
+                    if not semver.satisfies(sdl2ver, opt.require_version) then
+                        return -- system sdl2 doesn't satisfy version requirements
+                    end
+                end
```
waruqi, line 55: please remove this `if end` block, because we need only set result.version, xmake/dev will match version automatically.

packages/l/libsdl/xmake.lua
```diff
@@ -23,9 +23,76 @@ package("libsdl")
...
+                        result.defines = result.defines or {}
+                        table.insert(result.defines, define)
+                    end
+                end
+
+                return result
+            end
+
+            -- finding using sdl2-config didn't work, fallback on pkgconfig
+            return find_package("pkgconfig::sdl2", {require_version = opt.require_version})
```
waruqi, line 92: please use `package:find_package("pkgconfig::sdl2", opt)` on xmake/dev

```lua
if package.find_package then
        return package::find_package("pkgconfig::sdl2", opt)
end
```

packages/l/libsdl/xmake.lua
```diff
@@ -89,7 +83,11 @@ package("libsdl")
             end
 
             -- finding using sdl2-config didn't work, fallback on pkgconfig
-            return find_package("pkgconfig::sdl2", {require_version = opt.require_version})
+            if package.find_package then
+                return package.find_package("pkgconfig::sdl2", opt)
```
waruqi, line 87: `return package:find_package("..") `

it should be `:` not `.` here

SirLynix: Oh yeah, my mistake.

waruqi:
I will fix ci fails on fedora on next commit.

PR #404, opened 2021-05-06
update libmpdclient package

Phate6660:
This fixes an issue I encountered where the `version.h` header wasn't 
being installed due to it being generated at build time, and then
removes the `version.h.in` file that isn't needed.

### DIFF
--- a/packages/l/libmpdclient/xmake.lua
+++ b/packages/l/libmpdclient/xmake.lua
@@ -1,12 +1,15 @@
 package("libmpdclient")
     set_homepage("https://musicpd.org/libs/libmpdclient/")
     set_description("A stable, documented, asynchronous API library for interfacing MPD in the C, C++ & Objective C languages.")
+    set_kind("shared")
     add_urls("https://musicpd.org/download/libmpdclient/2/libmpdclient-$(version).tar.xz")
     add_versions("2.19", "158aad4c2278ab08e76a3f2b0166c99b39fae00ee17231bd225c5a36e977a189")
     add_deps("meson", "ninja")
 
     on_install("linux", function (package)
-        import("package.tools.meson").install(package)
+        local configs = {}
+        table.insert(configs, "-Ddefault_library=shared")
+        import("package.tools.meson").install(package, configs)
         os.cp("include", package:installdir())
         os.cp("build_*/version.h", package:installdir() .. "/include/mpd")
         os.rm(package:installdir() .. "/include/mpd/version.h.in")
--- a/packages/l/libmpdclient/xmake.lua
+++ b/packages/l/libmpdclient/xmake.lua
@@ -8,6 +8,8 @@ package("libmpdclient")
     on_install("linux", function (package)
         import("package.tools.meson").install(package)
         os.cp("include", package:installdir())
+        os.cp("build_*/version.h", package:installdir() .. "/include/mpd")
+        os.rm(package:installdir() .. "/include/mpd/version.h.in")
     end)
 
     on_test(function (package)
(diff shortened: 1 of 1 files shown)


packages/l/libmpdclient/xmake.lua
```diff
@@ -8,6 +8,8 @@ package("libmpdclient")
     on_install("linux", function (package)
         import("package.tools.meson").install(package)
         os.cp("include", package:installdir())
+        os.cp("build_*/version.h", package:installdir() .. "/include/mpd")
```
waruqi, line 11: import("package.tools.meson").install(package, {}, {buildir = "build"})
then 

os.cp("build/version.h" ...

Phate6660: I can't use `build` as that folder is already present with files in it.
So I did that using `build_real` instead but it fails to build.

Because even though `ninja` is trying to use the `build_real` dir, meson still uses `build_$HASH`.

waruqi: oh, this is xmake's issue. I have fixed it on dev branch.

packages/l/libmpdclient/xmake.lua
```diff
@@ -8,6 +8,8 @@ package("libmpdclient")
     on_install("linux", function (package)
         import("package.tools.meson").install(package)
```
waruqi, line 9: We need to use `package:config("shared")` to configure to distinguish static/shared library.

Phate6660: I'm not exactly what sure what I'm supposed to do here to be honest.
I've tried various combinations of adding `package:config("shared")` and I don't think I'm getting it.

waruqi: like these:

https://github.com/xmake-io/xmake-repo/blob/a8fa2de0c922e364a74921b8e118fce229a4e04e/packages/l/libisl/xmake.lua#L14-L15

https://github.com/xmake-io/xmake-repo/blob/a8fa2de0c922e364a74921b8e118fce229a4e04e/packages/l/libogg/xmake.lua#L19

But I don't know what configuration parameters should be passed to meson to set static/shared, and I don't know meson very well.

```
        local configs = {} -- add static/shared configs?
        import("package.tools.meson").install(package, configs)
```

waruqi: https://github.com/mesonbuild/meson/issues/2910 maybe -Ddefault_library=static and -Ddefault_library=shared

Phate6660: I think I got it now.

packages/l/libmpdclient/xmake.lua
Phate6660: All right, will install xmake from the dev branch then.

packages/l/libmpdclient/xmake.lua
```diff
@@ -1,12 +1,15 @@
 package("libmpdclient")
     set_homepage("https://musicpd.org/libs/libmpdclient/")
     set_description("A stable, documented, asynchronous API library for interfacing MPD in the C, C++ & Objective C languages.")
+    set_kind("shared")
```
waruqi, line 4: no, 

only 

```console
        table.insert(configs, "-Ddefault_library=" .. (package:config("shared") and "shared" or "static"))
```

packages/l/libmpdclient/xmake.lua
waruqi: we can improve it again after the new version is released.

Phate6660: All right :+1:

waruqi:
thanks.

waruqi:
I fixed missing links for this package, see https://github.com/xmake-io/xmake-repo/pull/405

PR #431, opened 2021-05-26
add pkgconf package

Phate6660:
Currently still a draft, turns out it's installable on Windows too so I'll need to add Windows support to the pacakge.

### DIFF
--- a/packages/p/pkgconf/xmake.lua
+++ b/packages/p/pkgconf/xmake.lua
@@ -3,7 +3,7 @@ package("pkgconf")
     set_homepage("http://pkgconf.org")
     set_description("A program which helps to configure compiler and linker flags for development frameworks.")
     add_urls("https://distfiles.dereferenced.org/pkgconf/pkgconf-$(version).tar.xz")
-    add_versions("1.1.0", "5f1ef65d73a880fa5e7012102a17f7b32010e5e46139aed85851a541ba828a63")
+    add_versions("1.7.4", "d73f32c248a4591139a6b17777c80d4deab6b414ec2b3d21d0a24be348c476ab")
 
     on_load("windows", function(package)
         package:add("deps", "cmake")
--- a/packages/p/pkgconf/xmake.lua
+++ b/packages/p/pkgconf/xmake.lua
@@ -22,7 +22,7 @@ package("pkgconf")
     end)
 
     on_install("windows", function(package)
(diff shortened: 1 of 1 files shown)


packages/p/pkgconf/xmake.lua
```diff
@@ -0,0 +1,16 @@
...
+    on_install("linux", "bsd", function(package)
+        local configs = {}
+        table.insert(configs, "--enable-shared=" .. (package:config("shared") and "yes" or "no")) 
+        table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes")) 
+        if package:config("pic") ~= false then
+            table.insert(configs, "--with-pic")
+        end
+
+        import("package.tools.autoconf").install(package, configs)
+    end)
```
waruqi, line 16: please add `on_test`

packages/p/pkgconf/xmake.lua
```diff
@@ -0,0 +1,16 @@
+package("pkgconf")
```
waruqi, line 1: add `set_kind("binary")`

packages/p/pkgconf/xmake.lua
```diff
@@ -0,0 +1,29 @@
...
+        if package:config("pic") ~= false then
+            table.insert(configs, "--with-pic")
+        end
+
+        import("package.tools.autoconf").install(package, configs)
+    end)
+
+    on_install("windows", function(package)
+        import("package.tools.cmake").install(package)
+    end)
```
waruqi, line 25: does pkgconf support windows?

Phate6660: https://github.com/pkgconf/pkgconf/blob/master/README.md#compiling-pkgconf-and-libpkgconf-with-cmake-usually-for-windows

It says that it does, but I think it's README is outdated. It doesn't have any CMake stuff.
However I do see meson stuff, so maybe they switched build systems and forgot to document it?

Phate6660: Also turns out this is an old pkgconf version because their website is outdated and the download link points to a really old version.

Phate6660: And yep, building on Windows was moved to Meson. So I'll update that.

Phate6660:
I see passing Windows CIs!

PR #432, opened 2021-05-26
use `pkgconf` instead of `pkg-config` when on bsd in imagemagick package

Phate6660:


packages/i/imagemagick/xmake.lua
```diff
@@ -46,9 +46,19 @@ package("imagemagick")
...
             package:add("syslinks", "pthread")
         end
+
+        if package:is_plat("linux") then
+            package:add("deps", "pkg-config")
+        end
+
+        if package:is_plat("bsd") then
+            package:add("deps", "pkgconf")
+            package:set_env("PKG_CONFIG", path.join(packackage:dep("pkgconf"):installdir("bin"), "pkgconf"))
```
waruqi, line 60: package:addenv

packages/i/imagemagick/xmake.lua
```diff
@@ -46,9 +46,19 @@ package("imagemagick")
...
             package:add("syslinks", "pthread")
         end
+
+        if package:is_plat("linux") then
+            package:add("deps", "pkg-config")
+        end
+
+        if package:is_plat("bsd") then
+            package:add("deps", "pkgconf")
+            package:add_env("PKG_CONFIG", path.join(packackage:dep("pkgconf"):installdir("bin"), "pkgconf"))
```
waruqi, line 60: not add_env,  is addenv

waruqi:
```lua
local pkgconf = package:find_tool("pkgconf")
if pkgconf then
    package:addenv("PKG_CONFIG", pkgconf.program)
end
```

Phate6660:
That Fedora CI was just due to a timeout when downloading the tarball.
Given that all the others passed, it's probably safe to say that would have too.
But we could run the CI again.

Besides that, it seems like the BSD CIs passed!

PR #441, opened 2021-05-30
libvorbis: Add vorbisenc and vorbisfile support

SirLynix:
I think this could be improved by making separate xmake packages for vorbisenc and vorbisfile later (as they're separate packages for pkgconfig for example), but I'm not sure how to do that.

### DIFF
--- a/packages/l/libvorbis/xmake.lua
+++ b/packages/l/libvorbis/xmake.lua
@@ -33,20 +33,20 @@ package("libvorbis")
                     return
                 end
 
-                result.sysincludedirs = table.join(vorbisenc.sysincludedirs or vorbisenc.includedirs, result.sysincludedirs or {})
-                result.linkdirs = table.join(vorbisenc.linkdirs, result.linkdirs or {})
-                result.links = table.join(vorbisenc.links, result.links or {})
+                result.sysincludedirs = table.join(vorbisenc.sysincludedirs or vorbisenc.includedirs, result.sysincludedirs)
+                result.linkdirs = table.join(vorbisenc.linkdirs, result.linkdirs)
+                result.links = table.join(vorbisenc.links, result.links)
             end
-            
+
             if package:config("vorbisfile") then
                 local vorbisfile = find_package("vorbisfile", opt)
                 if not vorbisfile then
                     return
                 end
 
-                result.sysincludedirs = table.join(vorbisfile.sysincludedirs or vorbisfile.includedirs, result.sysincludedirs or {})
-                result.linkdirs = table.join(vorbisfile.linkdirs, result.linkdirs or {})
-                result.links = table.join(vorbisfile.links, result.links or {})
+                result.sysincludedirs = table.join(vorbisfile.sysincludedirs or vorbisfile.includedirs, result.sysincludedirs)
+                result.linkdirs = table.join(vorbisfile.linkdirs, result.linkdirs)
+                result.links = table.join(vorbisfile.links, result.links)
             end
 
             return result
--- a/packages/l/libvorbis/xmake.lua
+++ b/packages/l/libvorbis/xmake.lua
@@ -10,8 +10,8 @@ package("libvorbis")
 
     add_versions("1.3.7", "0e982409a9c3fc82ee06e08205b1355e5c6aa4c36bca58146ef399621b0ce5ab")
 
-    add_configs("with_vorbisenc",  {description = "Includes vorbisenc", default = true, type = "boolean"})
-    add_configs("with_vorbisfile", {description = "Includes vorbisfile", default = true, type = "boolean"})
+    add_configs("vorbisenc",  {description = "Includes vorbisenc", default = true, type = "boolean"})
+    add_configs("vorbisfile", {description = "Includes vorbisfile", default = true, type = "boolean"})
 
     add_deps("cmake", "libogg")
 
@@ -23,7 +23,7 @@ package("libvorbis")
             end
             local result = table.copy(vorbis)
 
-            if package:config("with_vorbisenc") then
+            if package:config("vorbisenc") then
                 local vorbisenc = find_package("vorbisenc")
                 if not vorbisenc then
                     return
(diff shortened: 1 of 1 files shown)


packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,66 @@ package("libvorbis")
 
     add_versions("1.3.7", "0e982409a9c3fc82ee06e08205b1355e5c6aa4c36bca58146ef399621b0ce5ab")
 
+    add_configs("with_vorbisenc",  {description = "Includes vorbisenc", default = true, type = "boolean"})
+    add_configs("with_vorbisfile", {description = "Includes vorbisfile", default = true, type = "boolean"})
```
waruqi, line 14: please remove `with_` prefix

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,66 @@ package("libvorbis")
...
     add_versions("1.3.7", "0e982409a9c3fc82ee06e08205b1355e5c6aa4c36bca58146ef399621b0ce5ab")
 
+    add_configs("vorbisenc",  {description = "Includes vorbisenc", default = true, type = "boolean"})
+    add_configs("vorbisfile", {description = "Includes vorbisfile", default = true, type = "boolean"})
+
     add_deps("cmake", "libogg")
 
+    on_fetch(function (package, opt)
+        if opt.system then
+            local vorbis = find_package("vorbis")
```
waruqi, line 20: package:find_package("", opt)

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,66 @@ package("libvorbis")
...
+            if not vorbis then
+                return
+            end
+            local result = table.copy(vorbis)
+
+            if package:config("vorbisenc") then
+                local vorbisenc = find_package("vorbisenc")
```
waruqi, line 27: here

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,66 @@ package("libvorbis")
...
+                if not vorbisenc then
+                    return
+                end
+
+                result.includedirs = table.join(vorbisenc.includedirs, result.includedirs or {})
```
waruqi, line 32: and sysincludedirs

SirLynix: Should I replace includedirs by sysincludedirs or add a line to join sysincludedirs?

waruqi: join(includedirs or sysincludedirs)

packages/l/libvorbis/xmake.lua
```diff
@@ -17,30 +17,34 @@ package("libvorbis")
 
     on_fetch(function (package, opt)
         if opt.system then
-            local vorbis = find_package("vorbis")
+            local vorbis = find_package("vorbis", opt)
```
waruqi, line 20: package:find_package after v2.5.3

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,71 @@ package("libvorbis")
...
+            local result = table.copy(vorbis)
+            if result.includedirs then
+                result.sysincludedirs = result.includedirs
+                result.includedirs = nil
+            end
+
+            if package:config("vorbisenc") then
+                local vorbisenc = package:find_package("vorbisenc", opt)
+                if not vorbisenc then
+                    return
```
waruqi, line 33: I do not think that it should use return.

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,71 @@ package("libvorbis")
...
+                end
+
+                result.sysincludedirs = table.join(vorbisenc.sysincludedirs or vorbisenc.includedirs, result.sysincludedirs)
```
waruqi, line 36: result.includedirs = table.join(result.sysincludedirs, vorbisenc.sysincludedirs or vorbisenc.includedirs)

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,71 @@ package("libvorbis")
...
+                result.linkdirs = table.join(vorbisenc.linkdirs, result.linkdirs)
+                result.links = table.join(vorbisenc.links, result.links)
+            end
```
waruqi, line 39: You can refer

https://github.com/xmake-io/xmake-repo/blob/30e6efbbfb178d285d7031d25222bd7ee0898dbc/packages/b/brotli/xmake.lua#L27-L44

waruqi: I will consider add a helper function to simply merge multi-package results.

packages/l/libvorbis/xmake.lua
SirLynix: I think it should, if the user asked for vorbisenc and only vorbis is found it will be a problem, so in this case it will build the xmake package which has this library

packages/l/libvorbis/xmake.lua
SirLynix: I put it in this order on purpose, since vorbisenc and vorbisfile libraries must be there before vorbis as they depend on it

packages/l/libvorbis/xmake.lua
waruqi: got it

packages/l/libvorbis/xmake.lua
```diff
@@ -10,13 +10,67 @@ package("libvorbis")
...
+            local result = table.copy(vorbis)
+
+            if package:config("vorbisenc") then
+                local vorbisenc = package:find_package("vorbisenc", opt)
+                if not vorbisenc then
+                    return
+                end
+
+                result.includedirs = table.join(vorbisenc.sysincludedirs or vorbisenc.includedirs, result.includedirs)
```
waruqi, line 32: we should join result.includedir first. because vorbisenv.includedirs may be nil, it will break table.join and result.incluedirs will be missing 

```lua
result.includedirs = table.join(result.includedirs, vorbisenc.sysincludedirs or vorbisenc.includedirs)
```

```console
$ xmake l find_package libvorbisfile
{
  linkdirs = {
    "/usr/lib/x86_64-linux-gnu"
  },
  links = {
    "vorbisfile"
  }
}
```

SirLynix: I reworked on_fetch based on what brotli does, it will probably be easier if you introduce a function to merge pkg infos later this way.

PR #456, opened 2021-06-15
Add function_ref and add headeronly flag on some libs

SirLynix:
Made by the same guy that made the `expected` package, maybe we should name them tl_expected and tl_function_ref as they're common names.

### DIFF
--- a/packages/c/catch2/xmake.lua
+++ b/packages/c/catch2/xmake.lua
@@ -1,5 +1,6 @@
 package("catch2")
 
+    set_kind("library", {headeronly = true})
     set_homepage("https://github.com/catchorg/Catch2")
     set_description("Catch2 is a multi-paradigm test framework for C++. which also supports Objective-C (and maybe C). ")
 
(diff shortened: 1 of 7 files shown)


waruqi:
> Made by the same guy that made the `expected` package, maybe we should name them tl_expected and tl_function_ref as they're common names.

But maybe some people have already used the expected package, which may cause the existing project to be broken.

SirLynix:
> But maybe some people have already used the expected package, which may cause the existing project to be broken.
I did think of that, maybe we could add tl_expected as a duplicate of expected and add a warning in its on_load saying "please update package expected to tl_expected" for a while before removing it.

waruqi:
> > But maybe some people have already used the expected package, which may cause the existing project to be broken.
> 
> I did think of that, maybe we could add tl_expected as a duplicate of expected and add a warning in expected on_load saying "please update package expected to tl_expected" for a while before removing it. Do you want me to do it?

that's ok, thanks.

PR #457, opened 2021-06-16
update python

xq114:


waruqi:
切到 py 3.9.5后 linux上 meson似乎跪了

https://github.com/xmake-io/xmake-repo/runs/2862737512?check_suite_focus=true

PR #478, opened 2021-07-10
Add `gtest_main` and configs to `gtest`

### LINKED ISSUE #475: `gtest` package is missing the `gtest_main` target
## The problem

The `gtest` package is missing the `gtest_main` target, making it impossible to link a binary without a hand-written `main` function:

```
error: LINK : fatal error LNK1561: entry point must be defined
```

## Proposed solution

The original `CMakeLists.txt` [links](https://github.com/google/googletest/blob/master/googletest/CMakeLists.txt#L146) the `gtest_main` target to `gtest`, so I believe the best way to opt-in to `gtest_main` from an `xmake.lua` script would look like this:

```lua
add_packages("gtest", {links="gtest_main"})
```

I am currently trying to get this to work locally, unfortunately without success so far (I am new to xmake). If I manage to find a clean way to get it to work I will probably open a pull request. But for now, I don't have a solution, which is why I'm opening this issue.

nasso:
Configuration options have been added to decide whether or not `gtest_main` should be linked against.
For consistency, a similar option was added for the `gmock` target.

Example usage with C++ 20:

```lua
-- use master for C++ 20 support (see https://github.com/google/googletest/issues/2914)
add_requires("gtest master", {configs = {main = true, gmock = true}})

target("my_test")
    set_languages("cxx20")
    set_kind("binary")

    add_files("my_test.cpp")
    add_packages("gtest")

    if is_plat("windows") then
        -- fixes "LINK : fatal error LNK1561: entry point must be defined"
        add_ldflags("/subsystem:console")
    end
```

Fixes #475.

### DIFF
--- a/packages/g/gtest/xmake.lua
+++ b/packages/g/gtest/xmake.lua
@@ -10,6 +10,9 @@ package("gtest")
     add_versions("github:1.10.0", "release-1.10.0")
     add_versions("archive:1.10.0", "94c634d499558a76fa649edb13721dce6e98fb1e7018dfaeba3cd7a083945e91")
 
+    add_configs("main",  {description = "Link to the gtest_main entry point.", default = false, type = "boolean"})
+    add_configs("gmock", {description = "Link to the googlemock library.", default = true, type = "boolean"})
+
     if is_plat("linux") then
         add_syslinks("pthread")
     end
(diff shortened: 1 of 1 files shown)


packages/g/gtest/xmake.lua
```diff
@@ -23,16 +35,39 @@ package("gtest")
...
+        if package:config("main") then
+            package:add("links", "gtest_main")
+        end
+
+        if package:config("gmock") then
+            package:add("links", "gmock")
+        end
+
+        package:add("links", "gtest")
+    end)
```
waruqi, line 69: we need not this on_load, because you have set them in on_install/set_default. 

please remove it.

packages/g/gtest/xmake.lua
```diff
@@ -10,6 +10,18 @@ package("gtest")
     add_versions("github:1.10.0", "release-1.10.0")
     add_versions("archive:1.10.0", "94c634d499558a76fa649edb13721dce6e98fb1e7018dfaeba3cd7a083945e91")
 
+    add_configs("main", {
```
waruqi, line 13: please use single line

nasso:
Should I squash the commits?

waruqi:
thanks

PR #494, opened 2021-07-11
Add `fx-gltf` package

nasso:
[fx-gltf] is a simpler and more efficient glTF loader than [tinygltf].

It's a header only library, so the installation simply copies the single header to the installation directory.

Because [fx-gltf] claims to be compatible with both C++14 and C++17, both are tested in `on_test` with `package:check_cxxsnippets`.

[fx-gltf]: https://github.com/jessey-git/fx-gltf
[tinygltf]: https://github.com/syoyo/tinygltf

### DIFF
--- a/packages/f/fx-gltf/xmake.lua
+++ b/packages/f/fx-gltf/xmake.lua
@@ -1,0 +1,24 @@
+package("fx-gltf")
+    set_homepage("https://github.com/jessey-git/fx-gltf")
+    set_description("A C++14/C++17 header-only library for simple, efficient, and robust serialization/deserialization of glTF 2.0")
+    set_license("MIT")
+
+    add_urls("https://github.com/jessey-git/fx-gltf/archive/refs/tags/v$(version).tar.gz", {alias = "archive"})
+    add_urls("https://github.com/jessey-git/fx-gltf.git", {alias = "github"})
+
+    add_versions("github:1.2.0", "v1.2.0")
+    add_versions("archive:1.2.0", "d8eaceba72ea6574b982c7b0d2328fd3f8ad519db4a37cf63cd3f8020d7722bf")
+
+    add_deps("nlohmann_json")
+
+    on_install(function(pkg)
+        os.cp("include/fx/gltf.h", pkg:installdir("include", "fx"))
+    end)
+
+    on_test(function(pkg)
+        assert(pkg:check_cxxsnippets({test = [[
+            void test() {
+                fx::gltf::Document doc;
+            }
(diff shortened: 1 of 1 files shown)


packages/f/fx-gltf/xmake.lua
```diff
@@ -0,0 +1,24 @@
+package("fx-gltf")
+    set_homepage("https://github.com/jessey-git/fx-gltf")
```
waruqi, line 2: we need set headeronly kind, like this

https://github.com/xmake-io/xmake-repo/blob/c1f4645941199156d5a1836267d4597bd2364acc/packages/d/dr_mp3/xmake.lua#L3

packages/f/fx-gltf/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+    set_description("A C++14/C++17 header-only library for simple, efficient, and robust serialization/deserialization of glTF 2.0")
+    set_license("MIT")
+
+    add_urls("https://github.com/jessey-git/fx-gltf/archive/refs/tags/v$(version).tar.gz", {alias = "archive"})
+    add_urls("https://github.com/jessey-git/fx-gltf.git", {alias = "github"})
+
+    add_versions("github:1.2.0", "v1.2.0")
+    add_versions("archive:1.2.0", "d8eaceba72ea6574b982c7b0d2328fd3f8ad519db4a37cf63cd3f8020d7722bf")
```
waruqi, line 10: we need only the following line.

add_versions("v1.2.0", "d8eaceba72ea6574b982c7b0d2328fd3f8ad519db4a37cf63cd3f8020d7722bf")

xmake will checkout v1.2.0 tag if git url is added and tag exists

waruqi: like this https://github.com/xmake-io/xmake-repo/blob/3dafd241e265fcc3b7282d9b7ca1b2aeff83d4f6/packages/t/tbox/xmake.lua#L11

packages/f/fx-gltf/xmake.lua
nasso: ooh i didn't know about this, i was looking for it, thanks!

packages/f/fx-gltf/xmake.lua
nasso: should i remove the aliases then?

waruqi: yeah, and remove `v` prefix of $(version) in url

waruqi:
thanks

PR #495, opened 2021-07-11
Add hffix v1.1.0

jj683:


jj683:
Sorry, I do not understand why windows tests runs even with:
```on_install("linux", "macosx", function (package)```

File: packages/h/hffix/xmake.lua [hunk 1]
```diff
@@ -0,0 +1,17 @@
+package("hffix")
+
+    set_homepage("https://github.com/jamesdbrock/hffix")
```
waruqi, line 3: we need set headeronly kind

see https://github.com/xmake-io/xmake-repo/blob/c1f4645941199156d5a1836267d4597bd2364acc/packages/d/dr_mp3/xmake.lua#L3

jj683: Thanks, fixed.

PR #508, opened 2021-07-22
luajit: fix x86_64 GC

SirLynix:
Reported by yamashi, this seems to fix a crash related to garbage collection (especially when ASLR is disabled)

### DIFF
--- a/packages/l/luajit/port/xmake.lua
+++ b/packages/l/luajit/port/xmake.lua
@@ -180,9 +180,6 @@ target("luajit")
     if is_kind("shared") and is_plat("windows") then
         add_defines("LUA_BUILD_AS_DLL")
     end
-    if not is_plat("windows") then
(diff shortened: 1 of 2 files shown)


psiberx:
Hi. I'm the author of the fix referred by yamashi.

1. I think GC64 should be optional for x64. This matches the behavior of the original LuaJIT build scripts (Makefile, MSVC).
2. In the current PR you add two DynASM sources at the same time. It should just be `vm_x64.dasc` when GC64 is on and `vm_x86.dasc` when GC64 is off.

I have a fix with optional GC64 ready and I was going to commit it today after some testing. 
If you wanna complete it yourself I can share what I have.

SirLynix:
Sure! Go ahead, because I have no idea what I'm doing here 😅

waruqi:
see https://github.com/xmake-io/xmake-repo/pull/510

PR #523, opened 2021-07-28
add date 3.0.1

jj683:


packages/d/date/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+    add_urls("https://github.com/HowardHinnant/date/archive/refs/tags/$(version).zip",
+             "https://github.com/HowardHinnant/date.git")
+
+    add_versions("v3.0.1", "f4300b96f7a304d4ef9bf6e0fa3ded72159f7f2d0f605bdde3e030a0dba7cf9f")
+
+    add_deps("cmake")
+
+    on_install(function (package)
+        local configs = {"-DBUILD_TZ_LIB=ON",
+                         "-DUSE_SYSTEM_TZ_DB=ON"}
```
waruqi, line 16: Why are the built-in build tz lib and system libraries enabled at the same time? If there is a system library, will it build tz lib by itself?

jj683: https://github.com/HowardHinnant/date/blob/master/CMakeLists.txt#L30

"BUILD_TZ_LIB": enables building "tz.cpp" (timezone) library.

"USE_SYSTEM_TZ_DB": for timezones we want to use local operating system information. Otherwise date library downloading their database with curl from the internet.

waruqi: ok, got it. thanks!

PR #571, opened 2021-08-16
Add more extsources to libx11

PucklaMotzer09:


SirLynix:
You're not obligated to make one PR per package, you can make a big PR that changes multiples packages at once

PR #572, opened 2021-08-16
Add extsources to make

PucklaMotzer09:


SirLynix:
I think this is unnecessary as xmake already tries to fetch packages using the package name. @waruqi could you confirm?

PucklaMotzer09:
> xmake already tries to fetch packages using the package name

Yes, if this is true then this is unncessary

waruqi:
we can input `m` to modify package source and select certain 3rd package manager. see

https://github.com/xmake-io/xmake/issues/1140#issuecomment-849270854
https://github.com/xmake-io/xmake/issues/1409

PR #574, opened 2021-08-17
Add MinGW support to libcurl, sol2 and lua

SirLynix:


SirLynix:
About iphoneos, I found this:
http://lua-users.org/lists/lua-l/2017-09/msg00242.html

I can implement it as a patch, what do you think?

waruqi:
> About iphoneos, I found this:
> http://lua-users.org/lists/lua-l/2017-09/msg00242.html
> 
> I can implement it as a patch, what do you think?

you can try it.

packages/l/lua/xmake.lua
```diff
@@ -26,10 +26,31 @@ package("lua")
...
+[[#if defined(LUA_CORE)
+/* disable system calls as it's not available on this sytem */
+#define system(s) ((s)==NULL ? 0 : -1)
+#endif
+
+/* =================================================================== */
+
+/*
+** Local configuration. You can use this space to add your redefinitions
+]], {plain = true})
```
waruqi, line 45: too long patch, you can use add_patches or io.open("", "a")/file:write to append content to tail

packages/l/libcurl/xmake.lua
```diff
@@ -106,3 +106,4 @@ package("libcurl")
     on_test(function (package)
         assert(package:has_cfuncs("curl_version", {includes = "curl/curl.h"}))
     end)
+
```
waruqi, line 109: remove empty line

packages/l/lua/xmake.lua
SirLynix: I'm not sure I can write a patch that will work with all versions of Lua, but I'll try.

Also this can't be tail since it has to be before the last #endif

packages/l/libcurl/xmake.lua
SirLynix: It was to force trigger package check

waruqi: you can reverse a commit and push force each time

SirLynix:
Ok iphoneos is fixed, now all that remains is cross:
```
error: @programdir/core/sandbox/modules/os.lua:263: error: @programdir/core/main.lua:279: @programdir/actions/config/main.lua:157: target(lualib): toolchain not found!
```

I'm not sure of what should be done here

waruqi:
> Ok iphoneos is fixed, now all that remains is cross:
> 
> ```
> error: @programdir/core/sandbox/modules/os.lua:263: error: @programdir/core/main.lua:279: @programdir/actions/config/main.lua:157: target(lualib): toolchain not found!
> ```
> 
> I'm not sure of what should be done here

I have fixed it in xmake/dev

PR #580, opened 2021-08-20
Add docking branch version to imgui package

AntoineJT:
This PR adds the latest stable docking branch version of imgui
to its xmake.lua package. It is up-to-date with the main one.
See: ocornut/imgui#4424

The extra code in the on_load function matches all the versions
containing '-docker' in their name and removes the precompiled
archive url, thus avoiding an obvious curl failure.
This allows to reduce CI build time by avoiding requesting
an HTTP 404 resource.

### DIFF
--- a/packages/i/imgui/xmake.lua
+++ b/packages/i/imgui/xmake.lua
@@ -6,6 +6,7 @@ package("imgui")
     add_urls("https://github.com/ocornut/imgui/archive/$(version).tar.gz",
              "https://github.com/ocornut/imgui.git")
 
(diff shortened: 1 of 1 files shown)


packages/i/imgui/xmake.lua
```diff
@@ -28,6 +29,9 @@ package("imgui")
             package:add("deps", "glfw")
             package:add("defines", "IMGUI_IMPL_OPENGL_LOADER_GLAD")
         end
+        if string.find(package:version_str(), "-docking") ~= nil then
+            package:set("urls", {"https://github.com/ocornut/imgui.git"})
```
waruqi, line 33: we need not this configuration, only need add_version

PR #587, opened 2021-08-23
add ogre-next

xq114:
macosx 上报错找不到符号sysctlbyname，我这没法调试

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -22,7 +22,7 @@ jobs:
           dnf -y install gfortran
           dnf -y install mesa-libGL-devel mesa-libGLU-devel
           dnf -y install @development-tools @rpm-development-tools
-          dnf -y install copr-cli make gcc-c++
+          dnf -y install copr-cli make gcc-c++ libatomic
           dnf -y upgrade git
           dnf -y install perl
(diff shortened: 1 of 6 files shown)


SirLynix:
Since this package is not compatible with ogre1, shouldn't we call it ogre2 or ogre-next?

waruqi:
> macosx 上报错找不到符号sysctlbyname，我这没法调试

    #include <sys/sysctl.h> 没加，master 版本似乎加上了，可以打个 patch，参考

https://github.com/OGRECave/ogre-next/blob/4b997cc4963f02ed12d694dfa87306bf8bfdc7e8/OgreMain/src/OgrePlatformInformation.cpp#L49-L52

waruqi:
> Since this package is not compatible with ogre1, shouldn't we call it ogre2 or ogre-next?

@xq114 It might be more appropriate to change to ogre-next, keep it consistent with the name on the github repo

xq114:
> Since this package is not compatible with ogre1, shouldn't we call it ogre2 or ogre-next?

Yes, ogre-next is a more appropriate name, since there is version 13.0.0 for ogre(1?). Mixing ogre and ogre-next can be confusing. I've renamed the package

PR #590, opened 2021-08-23
Add extsources for various packages

ImperatorS79:


waruqi:
Because some packages use extsources, getting packages from the system may cause the installation script of some packages to fail, and you need to fix them at the same time. 

For example, libsdl, you can do some repairs according to the error message on ci

ImperatorS79:
The function `on_install` is executed even for system package? Why?

waruqi:
> The function `on_install` is executed even for system package? Why?

No, the package it depends on has become a system library, and it still needs to run on_install to install it.

For example, in a failure case on ci, in the process of libsdl/on_install, after some of its dependent packages became system libraries, it failed to obtain some fetch information in it.

https://github.com/xmake-io/xmake-repo/blob/544507df931539f9b4cc401446cc904614b4fdd9/packages/l/libsdl/xmake.lua#L129

> error: ...r/work/xmake-repo/xmake-repo/packages/l/libsdl/xmake.lua:129: attempt to index a nil value

you can refer the solution of other people. like this

https://github.com/SirLynix/xmake-repo/blob/71b02a9c7bf1ad1fb865fbb9967092c8f6a5b0aa/packages/a/assimp/xmake.lua#L45

PR #599, opened 2021-08-27
Add extsources to libsdl

ImperatorS79:


waruqi:
it exists some conflicting files.

PR #603, opened 2021-08-27
Add extsources to stb

ImperatorS79:


waruqi:
please re-push it, it did not trigger ci on master branch.

PR #609, opened 2021-08-29
add openssl 1.1.1l & support static lib on windows

idealvin:
- add openssl 1.1.1l 
- support static lib on windows


packages/o/openssl/xmake.lua
```diff
@@ -14,6 +15,7 @@ package("openssl")
 
     if is_plat("windows") then
         add_links("libssl", "libcrypto")
+		add_syslinks("Ws2_32", "User32", "Crypt32", "Advapi32")
```
waruqi, line 18: 缩进不对 tab改成space4

PR #620, opened 2021-09-05
Add extsources to a-b-c packages

ImperatorS79:


packages/b/bzip2/xmake.lua
```diff
@@ -5,6 +5,14 @@ package("bzip2")
 
     add_urls("https://sourceware.org/pub/bzip2/bzip2-$(version).tar.gz")
     add_versions("1.0.8", "ab5a03176ee106d3f0fa90e381da478ddae405918153cca248e682cd0c4a2269")
+    
+    if is_plat("mingw") and is_subhost("msys") then
+        add_extsources("pacman::bzip2")
+    elseif is_plat("linux") then
+        add_extsources("pacman::bzip2", "apt::libbz2-dev")
```
SirLynix, line 12: apt::libbz2-dev isn't right as bzip2 package include binaries

ImperatorS79: Yes, so I need to put both I assume ?

SirLynix: Well it looks like apt/find_package doesn't support binaries anyway, better ask @waruqi

waruqi: > Well it looks like apt/find_package doesn't support binaries anyway, better ask @waruqi

right, only for library.

waruqi:
You can split this patch, submit multiple pr to fix them.

PR #627, opened 2021-09-08
Remake bzip2 using xmake.lua

SirLynix:
bzip2 didn't support dynamic build and didn't compile on mingw, so I remade it using a xmake.lua that should compile on every platform

### DIFF
--- a/packages/b/bzip2/port/xmake.lua
+++ b/packages/b/bzip2/port/xmake.lua
@@ -3,8 +3,10 @@ add_rules("mode.debug", "mode.release")
 target("bz2")
     set_kind("$(kind)")
     set_languages("c89")
-    if is_plat("windows", "mingw") and is_kind("static") then
+    if is_kind("static") then
         add_defines("BZ_STATIC", {public = true})
+    else
+        add_defines("BZ_EXPORT")
     end
 
     add_headerfiles("bzlib.h")
--- a/packages/b/bzip2/port/xmake.lua
+++ b/packages/b/bzip2/port/xmake.lua
@@ -1,5 +1,9 @@
 add_rules("mode.debug", "mode.release")
 
+option("build_bin")
+    set_default(true)
+    set_showmenu(true)
+
 target("bz2")
     set_kind("$(kind)")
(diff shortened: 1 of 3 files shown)


SirLynix:
@waruqi I removed this line
```lua
        package:addenv("PATH", "bin")
```

because I assumed xmake does it automatically, but I'm not sure about it, should I add it back in on_load?

waruqi:
> @waruqi I removed this line
> 
> ```lua
>         package:addenv("PATH", "bin")
> ```
> 
> because I assumed xmake does it automatically, but I'm not sure about it, should I add it back in on_load/on_install?

You need add it. only binary package (`set_kind("binary") in package()`) will add it automatically.

packages/b/bzip2/xmake.lua
```diff
@@ -14,6 +14,8 @@ package("bzip2")
     end)
 
     on_install(function (package)
+        package:addenv("PATH", "bin")
```
waruqi, line 17: you can move to on_load

packages/b/bzip2/port/xmake.lua
```diff
@@ -18,38 +22,42 @@ target("bz2")
...
-            os.ln(path.join(binarydir, "bzmore"), path.join(binarydir, "bzless"))
-        end
-    end)
-
-target("bzip2recover")
-    set_kind("binary")
-    add_deps("bz2")
-    add_files("bzip2recover.c")
+if has_config("build_bin") then
+
```
waruqi, line 26: we can also use `set_default(false)` to disable it by default. then we use `xmake build --all` to build all

waruqi: oh, but package.tools.xmake has not support it yet. = =

xq114:
The patch is not needed, and it has caused problems with other packages like libarchive. bzip2 already has a def file libbz2.def, `__declspec(dllexport)` is not required actually. I'm reverting the patch now (https://github.com/xmake-io/xmake-repo/pull/702), if it does affect your code, please make me aware.

PR #628, opened 2021-09-08
Freetype: Fix package finding for MinGW

SirLynix:
I created a monster.

FindSomething.cmake worked on Windows but not on MinGW, so I added all include_dirs and libraries cmake options which fixed it.

### DIFF
--- a/packages/f/freetype/xmake.lua
+++ b/packages/f/freetype/xmake.lua
@@ -54,27 +54,26 @@ package("freetype")
         local configs = {}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
-        local function add_dep(dep, cmakeConf, cmakeDisableConf)
-            if package:config(dep) then
-                table.insert(configs, "-DFT_WITH_" .. cmakeConf .. "=ON")
-            else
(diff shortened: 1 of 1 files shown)


packages/f/freetype/xmake.lua
```diff
@@ -54,27 +54,26 @@ package("freetype")
...
-            local brotli = package:dep("brotli")
-            if brotli and not brotli:is_system() then
-                local fetchinfo = brotli:fetch()
-                table.insert(configs, "-DBROTLIDEC_INCLUDE_DIRS=" .. table.concat(fetchinfo.includedirs or fetchinfo.sysincludedirs, ";"))
-                table.insert(configs, "-DBROTLIDEC_LIBRARIES=" .. table.concat(fetchinfo.libfiles, ";"))
+                local lib = package:dep(opt.pkg or opt.conf)
+                if lib and not lib:is_system() then
+                    local includeconf = opt.cmakeinclude or (opt.cmakewith .. "_INCLUDE_DIRS")
+                    local libconf = opt.cmakelib or (opt.cmakewith .. "_LIBRARIES")
+                    local fetchinfo = lib:fetch()
```
waruqi, line 65: I think you should add `if fetchinfo then` or assert

SirLynix: In what cases can :fetch() return nil in this case?

waruqi: > In what cases can :fetch() return nil in this case?

Normally, it will not return nil, but it will always be fine after checking.

PR #636, opened 2021-09-12
Add extsources to atk

ImperatorS79:
I removed cairo to see if it works.

### DIFF
--- a/packages/a/atk/xmake.lua
+++ b/packages/a/atk/xmake.lua
@@ -9,8 +9,12 @@ package("atk")
     end})
     add_versions("2.36.0", "fb76247e369402be23f1f5c65d38a9639c1164d934e40f6a9cf3c9e96b652788")
 
-    if is_plat("linux") then
-        add_extsources("apt::libatk1.0-dev")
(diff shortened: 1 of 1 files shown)


waruqi:
Can you help me to add pkgconfig::xx extsources? it will be more accurate than apt::

waruqi:
pcre url broken, maybe you need wait for https://github.com/xmake-io/xmake-repo/pull/677

waruqi:
pcre has been fixed

waruqi:
Since you modify too many packages at once, this will trigger a large number of builds. On msys ci, it was too slow, so it was forcibly cancelled.

In addition, the collada-dom package will be broken on windows 2016.

I suggest to split it into multiple smaller prs to facilitate the passage and merging of ci. We can also temporarily ignore the problematic collada-dom so that we can fix it in a targeted manner later.

ImperatorS79:
The failing step fails because https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.2.tar.gz returns error 404 but it works for me.

waruqi:
I rerun it.

waruqi:
thanks

PR #641, opened 2021-09-15
Fix freetype for MinGW, again

SirLynix:
Just found out there was a mistake with link order (zlib being used by png) and bzip2 cmake include config

### DIFF
--- a/packages/f/freetype/xmake.lua
+++ b/packages/f/freetype/xmake.lua
@@ -45,9 +45,9 @@ package("freetype")
         end
 
         add_dep("bzip2")
+        add_dep("zlib")
         add_dep("png", "libpng")
         add_dep("woff2", "brotli")
-        add_dep("zlib")
     end)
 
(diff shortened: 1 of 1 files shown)


packages/f/freetype/xmake.lua
```diff
@@ -72,7 +72,7 @@ package("freetype")
                 table.insert(configs, "-DCMAKE_DISABLE_FIND_PACKAGE_" .. (opt.cmakedisable or opt.cmakewith) .. "=ON")
             end
         end
-        add_dep({conf="bzip2", cmakewith="BZIP2", cmakedisable="BZip2"})
+        add_dep({conf="bzip2", cmakewith="BZIP2", cmakedisable="BZip2", cmakeinclude="BZIP2_INCLUDE_DIR"})
```
waruqi, line 75: please add space `key = value`

SirLynix: Done

PR #642, opened 2021-09-16
add backward-cpp

jj683:


packages/b/backward-cpp/xmake.lua
```diff
@@ -0,0 +1,26 @@
...
+        local configs = {"-DBACKWARD_TESTS=OFF"}
+        table.insert(configs, "-DBACKWARD_SHARED=" .. (package:config("shared") and "ON" or "OFF"))
+        import("package.tools.cmake").build(package, configs, {buildir = "build_xmake"})
+
+        os.cp("backward.hpp", package:installdir("include/backward"))
+        if package:is_plat("windows") then
+            os.trycp(path.join("build_xmake", "*", "*.lib"), package:installdir("lib"))
+        else
+            os.trycp(path.join("build_xmake", "*.a"), package:installdir("lib"))
+        end
```
jj683, line 21: By default, cmake copies "backward.hpp" without additional "backward" folder.

waruqi: why not use cmake.install? no install scripts?

asafrnv: We can, but "backward.hpp" will be located in the "root" folder and we have to include it like this:
`#include "backward.hpp"`

If it's not a problem, I'll use cmake.install.

jj683: We can, but "backward.hpp" will be located in the "root" folder and we have to include it like this:
`#include "backward.hpp"`

If it's not a problem, I'll use cmake.install.

waruqi: you can add `add_includedirs(".")` to rewrite default include dir

waruqi: you can continue to copy/move header files after doing cmake.install

jj683: Sounds good, will fix it soon.

PR #679, opened 2021-11-07
Add extsources to benchmark

ImperatorS79:
Maybe I can add commit one by one until it crashes to see how far I can go in one PR ?


waruqi:
right, you can one by one

waruqi:
maybe build binutils is too slower on msys2

PR #721, opened 2021-11-29
update vulkan packages to 1.2.198

Bricklou:
Updated:
* `vulkan-headers` to sdk 1.2.198
* `vulkan-hpp` to sdk 1.2.198
* `vulkan-loader` to sdk 1.2.198
* `vulkan-tools` to sdk 1.2.198
* `vulkan-validationlayers` to sdk 1.2.198

### DIFF
--- a/packages/s/spirv-headers/xmake.lua
+++ b/packages/s/spirv-headers/xmake.lua
@@ -11,6 +11,7 @@ package("spirv-headers")
     add_urls("https://github.com/KhronosGroup/SPIRV-Headers/archive/$(version).tar.gz", {version = function (version) return map[tostring(version)] end})
     add_versions("1.5.4+2", "df2ad2520be4d95a479fa248921065885bbf435a658349a7fc164ad7b26b68c6")
     add_versions("1.5.5", "9eb56548460fd8850250ebf78071528fb66c2a5db2ef535edc1d493b2581ec66")
+    add_versions("sdk-1.2.198.0", "3301a23aca0434336a643e433dcacacdd60000ab3dd35dc0078a297c06124a12")
 
     add_deps("cmake")
 
--- a/packages/s/spirv-headers/xmake.lua
+++ b/packages/s/spirv-headers/xmake.lua
@@ -7,11 +7,12 @@ package("spirv-headers")
 
     local map = {["1.5.4+2"] = "1.5.4.raytracing.fixed",
                  -- Workaround since no new releases are available for spirv-headers, will be removed in future
-                 ["1.5.5"] = "e71feddb3f17c5586ff7f4cfb5ed1258b800574b"}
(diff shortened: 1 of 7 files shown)


waruqi:
windows ci fails. Maybe we need also update SPIRV-Headers to sdk-1.2.198.0? https://github.com/KhronosGroup/SPIRV-Headers/releases/tag/sdk-1.2.198.0

https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3553

Or you can split this pr and update Vulkan-ValidationLayers separately so that the patch can be merged quickly

packages/s/spirv-headers/xmake.lua
```diff
@@ -11,6 +11,7 @@ package("spirv-headers")
     add_urls("https://github.com/KhronosGroup/SPIRV-Headers/archive/$(version).tar.gz", {version = function (version) return map[tostring(version)] end})
     add_versions("1.5.4+2", "df2ad2520be4d95a479fa248921065885bbf435a658349a7fc164ad7b26b68c6")
     add_versions("1.5.5", "9eb56548460fd8850250ebf78071528fb66c2a5db2ef535edc1d493b2581ec66")
+    add_versions("sdk-1.2.198.0", "3301a23aca0434336a643e433dcacacdd60000ab3dd35dc0078a297c06124a12")
```
waruqi, line 14: we only support semver https://semver.org/. 

try `1.2.198+0-sdk` and add map item

waruqi: and we maybe need add this sdk version to vulkan validator package

Bricklou:
How can I fix this error ? :thinking: 
```log
error: @programdir\core\main.lua:280: @programdir\modules\private\action\require\impl\package.lua:868: package(spirv-headers/1.2.198+0-sdk:{"vs_runtime=MT"}): conflict dependences with package(spirv-headers/latest:{"vs_runtime=MT"})!
```
I understand that versions aren't the sames between those provided and the one I'm currently adding but nothing more.

waruqi:
> How can I fix this error ? 🤔
> 
> ```
> error: @programdir\core\main.lua:280: @programdir\modules\private\action\require\impl\package.lua:868: package(spirv-headers/1.2.198+0-sdk:{"vs_runtime=MT"}): conflict dependences with package(spirv-headers/latest:{"vs_runtime=MT"})!
> ```
> 
> I understand that versions aren't the sames between those provided and the one I'm currently adding but nothing more.

There is no good way yet. The version rules of spirv-headers have changed. Its latest version number is lower than 1.5.4, and vulkan-validationlayers 1.2.198 must depend on spirv-headers 1.2.198

And spirv-tools also depends on spirv-headers, vulkan-validationlayers also depends on them, so spirv-tools must also depend on spirv-headers 1.2.198

PR #753, opened 2021-12-11
Add mingw support to vulkan-hpp

PucklaMotzer09:
To build vulkan-hpp the VulkanHppGenerator needs to build for the host system. With this change xmake will temporarily switch the platform to the host to use cmake and then switch back.

### DIFF
--- a/packages/v/vulkan-hpp/xmake.lua
+++ b/packages/v/vulkan-hpp/xmake.lua
@@ -12,16 +12,18 @@ package("vulkan-hpp")
     add_deps("cmake")
 
     on_install("windows", "linux", "macosx", "mingw", function (package)
-        local is_mingw = package:is_plat("mingw")
-        if is_mingw then
-            package:requireinfo().plat = os.host()
+        local plat, arch
+        if package:is_plat("mingw") and package.plat_set then
+            plat = package:plat_set(os.host())
+            arch = package:arch_set(os.arch())
         end
         import("package.tools.cmake").build(package, {buildir = "build"})
-        if is_mingw then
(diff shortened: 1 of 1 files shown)


packages/v/vulkan-hpp/xmake.lua
```diff
@@ -11,10 +11,17 @@ package("vulkan-hpp")
...
+        if is_mingw then
+            package:requireinfo().plat = os.host()
+        end
         import("package.tools.cmake").build(package, {buildir = "build"})
+        if is_mingw then
+            package:requireinfo().plat = "mingw"
+        end
         os.mkdir("build")
-        if package:is_plat("windows") then
+        if os.is_host("windows") then
```
waruqi, line 24: is_host, not os.is_host

packages/v/vulkan-hpp/xmake.lua
```diff
@@ -11,10 +11,17 @@ package("vulkan-hpp")
 
     add_deps("cmake")
 
-    on_install("windows", "linux", "macosx", function (package)
+    on_install("windows", "linux", "macosx", "mingw", function (package)
+        local is_mingw = package:is_plat("mingw")
+        if is_mingw then
+            package:requireinfo().plat = os.host()
```
waruqi, line 17: it will generate VulkanHppGenerator for macOS if it is mingw on macOS

waruqi: and package:requireinfo().plat is private method

PucklaMotzer09:
Yes this achieves exactly what I want. VulkanHppGenerator should be build for the host, since it just genereates the headers

PR #758, opened 2021-12-14
Add ifort as toolchain

PucklaMotzer09:
This adds the ifort package as a toolchain. The compiler is installed using the installer from Intel in silent mode. Currently it is installed into the home directory. I would prefer to install it into the package installation directory, but I don't know how to tell the ifort toolchain of xmake where to find it then.

Also, the script is currently downloaded manually using curl, since the `add_urls` and `add_resources` functions assume that the content is a compressed folder and try to extract it. So it would be nice if you could download files different than compressed folders.


packages/i/ifort/xmake.lua
```diff
@@ -0,0 +1,68 @@
...
+        end
+    end)
+    
+    on_install("@linux", function(package)
+    local version = package:version_str()
+        os.execv("curl", {"https://registrationcenter-download.intel.com/akdlm/irc_nas/18210/l_fortran-compiler_p_" .. version .. ".3224.sh", "-o", path.join(package:cachedir(), "install_ifort.sh")})
+        os.cd(package:cachedir())
+        
+        local script_path = "./install_ifort.sh"
+        os.run("chmod +x " .. script_path)
```
waruqi, line 27: need only run `sh xxx.sh`

PucklaMotzer09:
windows is starting `bootstrapper.exe` in another process and therefore xmake thinks it is done, so currently we need to disable it. We could wait until the `bootstrapper.exe` is done and then continue, maybe.

waruqi:
> windows is starting `bootstrapper.exe` in another process and therefore xmake thinks it is done, so currently we need to disable it. We could wait until the `bootstrapper.exe` is done and then continue, maybe.

Usually, we can extract all the files after installing locally, compress them into zip files, and upload them to xmake-mirror to solve this problem.

But ifort is a commercialized project of intel, I am not sure if there is a problem with providing this kind of binary modification and distribution.

packages/i/ifort/xmake.lua
```diff
@@ -0,0 +1,75 @@
...
+        end
+    end)
+    
+    on_install("@linux", function(package)
+    local version = package:version_str()
+        os.execv("curl", {"https://registrationcenter-download.intel.com/akdlm/irc_nas/18210/l_fortran-compiler_p_" .. version .. ".3224.sh", "-o", path.join(package:cachedir(), "install_ifort.sh")})
+        os.cd(package:cachedir())
+        
+        local script_path = "install_ifort.sh"
+        local install_dir = vformat("$(env HOME)/intel/oneapi/")
```
waruqi, line 27: xmake package cannot be installed to system directory. we must use package:installdir()

waruqi:
> This adds the ifort package as a toolchain. The compiler is installed using the installer from Intel in silent mode. Currently it is installed into the home directory. I would prefer to install it into the package installation directory, but I don't know how to tell the ifort toolchain of xmake where to find it then.
> 
> Also, the script is currently downloaded manually using curl, since the `add_urls` and `add_resources` functions assume that the content is a compressed folder and try to extract it. So it would be nice if you could download files different than compressed folders.

I have improve xmake to support to download non-archived package urls.

PR #845, opened 2022-01-12
libunwind: fix description, use official home page.

cyfdecyf:
The original description is copied from `libunistring`. This PR fixes this.


waruqi:
thanks

PR #849, opened 2022-01-12
Add some pacman linux libs extsources

Arthapz:


waruqi:
oh, it will trigger to build too much packages at same time.

I think you should split into multiple prs, each containing 3-5 packages. It is easier to be passed on ci, and it is also convenient to modify ci errors to achieve fast merging

PR #914, opened 2022-01-17
[libxkbcommon] Support double backend installation

Arthapz:


packages/l/libxkbcommon/xmake.lua
```diff
@@ -29,10 +32,13 @@ package("libxkbcommon")
         local configs = {"-Denable-docs=false", "-Dc_link_args=-lm"}
         table.insert(configs, "--libdir=lib")
         if package:config("x11") then
-            table.join2(configs, {"-Denable-wayland=false", "-Dxkb-config-root=/usr/share/X11/xkb", "-Dx-locale-root=/usr/share/X11/locale"})
-        else
-            table.join2(configs, {"-Denable-x11=false", "-Dxkb-config-root=/usr/share/X11/xkb", "-Dx-locale-root=/usr/share/X11/locale"})
+            table.join2(configs, {"-Denable-x11=true", "-Dxkb-config-root=/usr/share/X11/xkb", "-Dx-locale-root=/usr/share/X11/locale"})
```
waruqi, line 35: we need disable x11 if x11 config item is false

```lua
if package:config("x11") then
else
    -- disable x11
end

> meson.build:273:8: ERROR: Problem encountered: X11 support requires xcb-xkb >= 1.10 which was not found.
You can disable X11 support with -Denable-x11=false.

PR #956, opened 2022-01-27
Update shaderwriter package

WubiCookie:


DragonJoker:
The issue encountered here comes from the use of GCC 9.
Is there a way to force GCC 10 ?

PR #988, opened 2022-02-08
gflags: disable register cmake user package registry.

cyfdecyf:
`gflags` by default adds registry file under `$HOME/.cmake` during install, which should be avoided when using Xrepo.

I've several questions related to `gflags` and `glog` pacakge, which I'd like to contribute further PRs but need guidance.

1. Is there any specific reason to use `mt=false` as default? I personally prefer to enable multi-thread by default for `gflags`, and this is also the default in vcpkg's `gflags` portfile.
2. Can I make `gflags` and `glog` packages to support "cross" platform? I wish for this because I want to specify different compilers when using xrepo-cmake. I can do this in my private repo, but would like to contribute if there's no problems for this.

### DIFF
--- a/packages/g/gflags/xmake.lua
+++ b/packages/g/gflags/xmake.lua
@@ -17,7 +17,11 @@ package("gflags")
         add_syslinks("pthread")
     end
     on_install("windows", "linux", "macosx", function (package)
-        local configs = {"-DBUILD_TESTING=OFF"}
+        local configs = {
+            "-DBUILD_TESTING=OFF",
+            "-DGFLAGS_REGISTER_BUILD_DIR=OFF",
+            "-DGFLAGS_REGISTER_INSTALL_PREFIX=OFF",
+        }
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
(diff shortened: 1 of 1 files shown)


xq114:
> 1. Is there any specific reason to use `mt=false` as default? I personally prefer to enable multi-thread by default for `gflags`, and this is also the default in vcpkg's `gflags` portfile.

https://github.com/gflags/gflags/blob/master/CMakeLists.txt#L173-L174 It just comes from the original settings here. When using gflags as a subproject, it only builds the single-threaded version by default. I don't know if there is any reason not building the multi-threaded version, since vcpkg has done switch the default to the multi-threaded version, I think it won't bother if we also do this switch.

waruqi:
> Is there any specific reason to use mt=false as default? I personally prefer to enable multi-thread by default for gflags, and this is also the default in vcpkg's gflags portfile.

As for modifying the default configuration, I'm not sure if it will affect other users currently using gflags. Because I don't know how many users are currently using gflags and don't know any other side effects of default multithreading.

> Can I make gflags and glog packages to support "cross" platform? I wish for this because I want to specify different compilers when using xrepo-cmake. I can do this in my private repo, but would like to contribute if there's no problems for this.

you can try.

PR #999, opened 2022-02-14
add hiredis

cyfdecyf:
The hiredis `CMakeLists.txt` does not provide option to build static or shared library. They are always built together.

The `Makefile` provides `dynamic`, `static` target, but seems to work only on Unix-like systems.

So I choose to use cmake in this PR.

### DIFF
--- a/packages/h/hiredis/patches/v1.0.2/cmake.patch
+++ b/packages/h/hiredis/patches/v1.0.2/cmake.patch
@@ -1,0 +1,192 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index f86c9b7..b545e21 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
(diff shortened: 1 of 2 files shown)


waruqi:
> The hiredis `CMakeLists.txt` does not provide option to build static or shared library. They are always built together.
> 
> The `Makefile` provides `dynamic`, `static` target, but seems to work only on Unix-like systems.
> 
> So I choose to use cmake in this PR.

you need remove static or shared and ensure only one kind library in installdir

linke this: 
https://github.com/xmake-io/xmake-repo/blob/3c499b1d0b912a91d3ae318c98f5cfe0649cc718/packages/w/workflow/xmake.lua#L24-L29

cyfdecyf:
hiredis master branch's cmake code supports static library (including other fixes), v1.0.2 does not.

So I used hiredis's latest cmake code to create a patch.

PR #1008, opened 2022-02-20
Fix doxygen

SirLynix:
Doxygen was broken because `https://doxygen.nl/files/doxygen-1.9.1.src.tar.gz` got removed, it seems only last version file is available.

I added the git as a fallback and removed windows binaries (as they can be handled by xmake build artifacts, also windows can now build it since we added support for winflexbison).


packages/d/doxygen/xmake.lua
```diff
@@ -4,28 +4,17 @@ package("doxygen")
...
-            add_urls("https://doxygen.nl/files/doxygen-$(version).windows.bin.zip")
-            add_versions("1.9.1", "c9782f545be757dac6e424f5347b7bbf1169da927058dc9954d801a5e8399de5")
-        elseif is_arch("x64") then
-            add_urls("https://doxygen.nl/files/doxygen-$(version).windows.x64.bin.zip")
-            add_versions("1.9.1", "deb8e6e5f21c965ec07fd32589d0332eff047f2c8658b5c56be4839a5dd43353")
-        end
-    else
-        add_urls("https://doxygen.nl/files/doxygen-$(version).src.tar.gz")
-        add_versions("1.9.1", "67aeae1be4e1565519898f46f1f7092f1973cce8a767e93101ee0111717091d1")
+    add_urls("https://doxygen.nl/files/doxygen-$(version).src.tar.gz", {alias="archive"})
```
waruqi, line 7: add spaces `alias = "archive"`

packages/d/doxygen/xmake.lua
```diff
@@ -4,28 +4,17 @@ package("doxygen")
...
+    add_versions("archive:1.9.3", "f352dbc3221af7012b7b00935f2dfdc9fb67a97d43287d2f6c81c50449d254e0")
+    add_versions("github:1.9.3", "Release_1_9_3")
+    add_versions("github:1.9.2", "Release_1_9_2")
+    add_versions("github:1.9.1", "Release_1_9_1")
 
-        add_deps("cmake", "bison", "flex")
-        add_deps("python 3.x", {kind = "binary"})
-    end
+    add_deps("cmake", "bison", "flex", {private=true})
+    add_deps("python 3.x", {kind = "binary", private=true})
```
waruqi, line 15: and here

PR #1017, opened 2022-02-23
libunwind: add version v1.6.2, new options: minidebuginfo, zlibdebuginfo

cyfdecyf:
libunwind package on Arch Linux depends on `xz` and `zlib`.

I'm not sure whether minidebuginfo and zlibdebuginfo is widely used. To keep the package behaves the same as before, these two options are disable by default.


packages/l/libunwind/xmake.lua
```diff
@@ -4,18 +4,35 @@ package("libunwind")
     set_description("A portable and efficient C programming interface (API) to determine the call-chain of a program.")
 
     add_urls("https://github.com/libunwind/libunwind/releases/download/$(version).tar.gz", {version = function (version)
-        return version .. "/libunwind-" .. (version:gsub("v", "")) .. ".0"
+        if version:eq("v1.5") then
+            return "v1.5/libunwind-1.5.0"
+        else
```
waruqi, line 9: 直接下面版本号改成 v1.5.0 应该就能完全统一了，这里硬编码区分 不好维护

PR #1046, opened 2022-03-04
Add nodeeditor

SirLynix:
now that qt is a package, we can provide libraries using it


SirLynix:
https://github.com/miurahr/aqtinstall/issues/497

PR #1057, opened 2022-03-13
Add qt5network

SirLynix:


waruqi:
Weird, why does it link to /usr/lib64/libk5crypto.so.3 on fedora

SirMishaa:
I helped a lot SirLynix, it should be fixed now !

waruqi:
I think there should be built-in ssl library in qt sdk, like qtlibpng, zlib, qtharfbuzz ..

waruqi:
> /usr/bin/ld: /usr/lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_ctrl@OPENSSL_1_1_1b'
/usr/bin/ld: /usr/lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_CTX_free@OPENSSL_1_1_1b'
/usr/bin/ld: /usr/lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_CTX_new_id@OPENSSL_1_1_1b'

It seems to depend on a modified version of openssl (with evp-kdf)

https://github.com/sailfishos/openssl/blob/master/rpm/openssl-1.1.1-evp-kdf.patch

SirLynix:
It's possible but I find it weird, how does it work when you install Qt on Fedora normally?

waruqi:
> It's possible but I find it weird, how does it work when you install Qt on Fedora normally?

It works if the system openssl (with evp-kdf) is installed on fedora.

```console
sudo dnf install openssl-devel
```

SirLynix:
> > It's possible but I find it weird, how does it work when you install Qt on Fedora normally?
> 
> It works if the system openssl (with evp-kdf) is installed on fedora.
> 
> ```
> sudo dnf install openssl-devel
> $ nm -D /usr/lib64/libcrypto.so | grep EVP_KDF_ctrl
> 0000000000126190 T EVP_KDF_ctrl@@OPENSSL_1_1_1b
> 0000000000126280 T EVP_KDF_ctrl_str@@OPENSSL_1_1_1b
> ```

Thanks for checking, what's the best way to handle this here? Change openssl package to include KDF on Fedora? Or require qt5network to have system-wide openssl on Fedora?

waruqi:
> Or require qt5network to have system-wide openssl on Fedora?

right, we use this solution now. In the future we can also improve openssl to support evp-kdf

PR #1081, opened 2022-03-21
Add support for Qt6

SirLynix:


SirLynix:
I'm a bit stuck at this point

waruqi:
> I'm a bit stuck at this point

What problem?

SirLynix:
> > I'm a bit stuck at this point
> 
> What problem?

Missing libicui18n.so dependency on Android, missing pcr2 on iOS (I didn't find any precompiled library file like for Qt5), and libfontconfig which isn't found on Archlinux (but is a dependency).

waruqi:
> > > I'm a bit stuck at this point
> > 
> > 
> > What problem?
> 
> Missing libicui18n.so dependency on Android, missing pcr2 on iOS (I didn't find any precompiled library file like for Qt5), and libfontconfig which isn't found on Archlinux (but is a dependency).

we can disable android/ios first for qt6

SirLynix:
> and add_deps("fontconfig") for linux

Well, that's the problem, it's already there:
https://github.com/xmake-io/xmake-repo/pull/1081/files#diff-7aa4b14e36d1159e91dde5eeb2ad987d929eb282d012174f264caa92f50efb1aR10

waruqi:
> > and add_deps("fontconfig") for linux
> 
> Well, that's the problem, it's already there: https://github.com/xmake-io/xmake-repo/pull/1081/files#diff-7aa4b14e36d1159e91dde5eeb2ad987d929eb282d012174f264caa92f50efb1aR10

maybe it only support ubuntu.

packages/q/qt6base/xmake.lua
```diff
@@ -0,0 +1,13 @@
+package("qt6base")
+    set_kind("phony")
+    set_base("qtbase")
+
+    add_versions("6.3.0", "dummy")
+
+    on_load(function (package)
+        package:set("kind", "phony")
```
waruqi, line 8: Why set it twice?

SirLynix: I can't recall, we can probably remove it

waruqi:
I think you should split this patch.

patch 1: update aqt
patch 2: improve qt5
patch 3: support qt6

SirLynix:
I think you're right, I'll do that.

SirLynix:
Superseded by #2281

PR #1097, opened 2022-03-24
nlohmann_json: use c++11 for on_test.

cyfdecyf:
nlohmann_json is written in vanilla C++11. It compile with gcc-4.8 which don't support `-std=c++14` option.


waruqi:
重新开个 pr 提交到 dev 分支，触发下 ci

PR #1100, opened 2022-03-25
libevent: fix build for shared or static lib.

cyfdecyf:
`DEVENT__LIBRARY_TYPE` defaults to `BOTH`, so the original version build both static and dynamic lib.

### DIFF
--- a/packages/l/libevent/xmake.lua
+++ b/packages/l/libevent/xmake.lua
@@ -25,7 +25,7 @@ package("libevent")
         io.replace("CMakeLists.txt", "advapi32", "advapi32 crypt32", {plain = true})
         local configs = {"-DEVENT__DISABLE_TESTS=ON", "-DEVENT__DISABLE_REGRESS=ON", "-DEVENT__DISABLE_SAMPLES=ON"}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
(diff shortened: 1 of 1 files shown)


cyfdecyf:
libevent [CMakeLists.txt](https://github.com/azat/libevent/blob/877f2355e9b1cee5190cc25c446b0effb1108989/CMakeLists.txt#L212-L217) says build both static and dynamic library is not supported for MSVC. But seems it does not support static lib.

Should I add?

```cmake
    if is_plat("windows", "mingw") then
        add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = true})
        add_configs("vs_runtime", {description = "Set vs compiler runtime.", default = "MD", readonly = true})
    end
```

waruqi:
> libevent [CMakeLists.txt](https://github.com/azat/libevent/blob/877f2355e9b1cee5190cc25c446b0effb1108989/CMakeLists.txt#L212-L217) says build both static and dynamic library is not supported for MSVC. But seems it does not support static lib.
> 
> Should I add following code:
> 
> ```lua
>     if is_plat("windows", "mingw") then
>         add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = true})
>         add_configs("vs_runtime", {description = "Set vs compiler runtime.", default = "MD", readonly = true})
>     end
> ```

跟着没关系吧，我看编译都过了，只是 on_test 对 static 缺库而已，`add_syslinks("ws2_32")` 加上就好了

waruqi:
还缺 advapi32

PR #1102, opened 2022-03-25
fix(packages/a/arrow): typo in ARROW_BUILD_SHARED

leodido:
The current `arrow` package contains a definition with a duplicated D.

### DIFF
--- a/packages/a/arrow/xmake.lua
+++ b/packages/a/arrow/xmake.lua
@@ -75,7 +75,7 @@ ${yellow}In case of boost dependency conflicts, please use following code (order
 
         local shared = package:config("shared")
         table.insert(configs, "-DARROW_BUILD_STATIC=" .. (shared and "OFF" or "ON"))
-        table.insert(configs, "-DDARROW_BUILD_SHARED=" .. (shared and "ON" or "OFF"))
+        table.insert(configs, "-DARROW_BUILD_SHARED=" .. (shared and "ON" or "OFF"))
         table.insert(configs, "-DARROW_DEPENDENCY_USE_SHARED=" .. (shared and "ON" or "OFF"))
 
         for config, enabled in pairs(package:configs()) do
--- a/packages/a/arrow/xmake.lua
+++ b/packages/a/arrow/xmake.lua
@@ -31,6 +31,12 @@ package("arrow")
 
     add_deps("cmake", "boost")
 
+    if is_plat("bsd") then
+        add_syslinks("pthread", "execinfo")
+    elseif is_plat("linux") then
+        add_syslinks("pthread")
+    end
+
     on_load(function (package)
         for name, dep in pairs(configdeps) do
             if package:config(name) then


cyfdecyf:
@leodido PR #1105 has lots of fixes and improvements for the arrow package. Please run `xrepo update` to get the latest changes.

I've also made a [project](https://github.com/cyfdecyf/xrepo-apache-arrow) to show how to use xmake to install Arrow with more features enabled, along with compiling PyArrow package.

leodido:
Thank you both!

I was about to send almost the same changes, but then I got stopped by `jemalloc` (which is giving me issues at the linking step, after successful compilation).

cyfdecyf:
I actually didn't test with `jemalloc` as `mimalloc` is default.

I guess the link problem arises because arrow is building vendored jemalloc and collecting the object file in `libarrow_bundled_dependencies.a`. I have to use `package:add("links", xxx)` to control link order in order for `on_test` to pass.

cyfdecyf:
@leodido Just tested with `jemalloc=true` and it works.

The PyArrow package build in my project still needs some work.

leodido:
@cyfdecyf I confirm that `jemalloc=true` works now.

But, I had to disable `lz4`. In fact, with `lz4=true` I keep getting a CMake error:

```
-- Checking for module 'liblz4'
--   Found liblz4, version 1.9.3
CMake Error at /usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Lz4 (missing: LZ4_LIB)
```

At first, I thought it was something similar to `boost` (which configs need to have `system = true`). 
But, configuring `lz4` dep as `lz4 = {version = "v1.9.3", configs = {system = true}},` doesn't help anyway.

Any idea?

To me it seems the error is [here](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L78).

cyfdecyf:
@leodido Do you have `pkg-config` installed on your system?

The whold build process should be able to use all non-system installed packages, which is how I did this in my helper project.

lz4 packages provides only pkg-config .pc files, so the `FindLz4.cmake` cmake module is relying on `PkgConfig` module to find lz4.

If you already have `pkg-config` installed. Please clone my [project](https://github.com/cyfdecyf/xrepo-apache-arrow) and then add a file `pkg/lz4.lua` with following content:

```lua
includes("common.lua")

add_requires("lz4", get_configs("lz4"))
```

Then run `xrepo fetch pkg/lz4.lua` and post the output.

cyfdecyf:
> At first, I thought it was something similar to `boost` (which configs need to have `system = true`).
> But, configuring `lz4` dep as `lz4 = {version = "v1.9.3", configs = {system = true}},` doesn't help anyway.

To use system provided package, the config should be:

```lua
lz4 = {version = "v1.9.3", system = true}
boost = {version = "1.78.0", system = true, configs = {system = true, ...}}
```

Note for boost: `configs = {system = true}` means enable the Boost System Library, this is not for using system installed boost.

leodido:
> @leodido Do you have `pkg-config` installed on your system?

Indeed I have it :)

> Then run `xrepo fetch pkg/lz4.lua` and post the output.

```
{ 
  { 
    links = { 
      "lz4" 
    },
    version = "v1.9.3",
    includedirs = { 
      "~/.xmake/packages/l/lz4/v1.9.3/55e9aa7ed4014be287ab80a488dfae47/include" 
    },
    libfiles = { 
      "~/.xmake/packages/l/lz4/v1.9.3/55e9aa7ed4014be287ab80a488dfae47/lib/liblz4.a" 
    },
    linkdirs = { 
      "~/.xmake/packages/l/lz4/v1.9.3/55e9aa7ed4014be287ab80a488dfae47/lib" 
    },
    static = true 
  } 
}
```

To me, it seems correct. But when I enable such a dep (`lz4`) I get the error mentioned above... 🤷 😕 

Anyways, I took a look at your example project. Very useful, thank you!
On a side note: you probably forgot snappy dependency for snappy compression, but overall it's very well done!

cyfdecyf:
It's weird. Xrepo adds all the depenedent packages' install paths to `CMAKE_PREFIX_PATH`, the [`PkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module would then search under those paths.

> ```
> -- Checking for module 'liblz4'
> --   Found liblz4, version 1.9.3
> CMake Error at /usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Lz4 (missing: LZ4_LIB)
> ```

Not sure why this happened. As I can't reproduce the problem, I can only give you some advices on possible ways to debug this:

1. `cd ~/.xmake/cache/packages/2203/a/arrow/7.0.0/source/arrow-apache-arrow-7.0.0/cpp`
2. Modify `cmake_modules/FindLz4.cmake` and add `message` to help identify which branch is taken
3. Run `xrepo install -vD -f xxx arrow`, look for the `cmake` configure command in the output
5. Run the `cmake configure` command use by `xrepo install`

> On a side note: you probably forgot snappy dependency for snappy compression, but overall it's very well done!

Snappy is now added. For fight rpc (which I don't use currently), it relies on `grpc` which is not currently in `xmake-repo` and requires more effort to add.

@waruqi The arrow build code uses some patched or pre-release version (for unreleased fixes) thirdparty libraries. e.g. [lz4](https://github.com/apache/arrow/blob/e90472e35b40f58b17d408438bb8de1641bfe6ef/cpp/thirdparty/versions.txt#L61) for ubsan, [snappy](https://github.com/apache/arrow/blob/e90472e35b40f58b17d408438bb8de1641bfe6ef/cpp/thirdparty/versions.txt#L84) for ubsan, [rapidjson](https://github.com/apache/arrow/blob/e90472e35b40f58b17d408438bb8de1641bfe6ef/cpp/thirdparty/versions.txt#L80) the last release is in 2016.

For those patches which fixes UBSan problems, can we submit PR to include thoses patches in `xmake-repo`?

For those pre-release version, can we create PR to create a virtual version to refer to speicfic commit?

waruqi:
> It's weird. Xrepo adds all the depenedent packages' install paths to `CMAKE_PREFIX_PATH`, the [`PkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module would then search under those paths.
> 
> > ```
> > -- Checking for module 'liblz4'
> > --   Found liblz4, version 1.9.3
> > CMake Error at /usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
> >   Could NOT find Lz4 (missing: LZ4_LIB)
> > ```
> 
> Not sure why this happened. As I can't reproduce the problem, I can only give you some advices on possible ways to debug this:

maybe we need add `add_rules("utils.install.pkgconfig_importfiles", {filename = "liblz4.pc"})` to fix .pc filename in xmake.lua 

[lz4 has been ported use xm
](https://github.com/xmake-io/xmake-repo/blob/4367726d063ce6a933a4af94ce7bab343f37e55a/packages/l/lz4/xmake.lua#L22-L36)

> 
> For those patches which fixes UBSan problems, can we submit PR to include thoses patches in `xmake-repo`?
> For those pre-release version, can we create PR to create a virtual version to refer to speicfic commit?


sure, you can try.

cyfdecyf:
> maybe we need add add_rules("utils.install.pkgconfig_importfiles", {filename = "liblz4.pc"}) to fix .pc filename in xmake.lua

Thanks. Guess this should be the root cause. I'll have a send a PR today.

cyfdecyf:
@leodido The fix for lz4 pkg-config file name has been merged. Arrow's `FindLz.cmake` is looking for .pc file named `liblz4.pc`, but xrepo installed one provides `lz4.pc` which I guess the cause of the problem. (I'm still not clear why I'm not affected as I don't have lz4 dev packages installed on system.)

Please run following command to see if the problem is fixed.

```
xrepo update-repo
xrepo remove lz4
xrepo install -f lz4=true arrow
```

leodido:
Thank you for looking into this @cyfdecyf 

I've investigated this another bit because it keeps **not** working even with your improvements for the location of the `.pc`.

I'm convinced the error has something to do with `find_package_handle_standard_args` (line [76](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L76)) of my CMake version, when the first level [else](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L54)-branch evaluates.

```
-- Checking for module 'liblz4'
--   Found liblz4, version 1.9.3
CMake Error at /usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Lz4 (missing: LZ4_LIB)
Call Stack (most recent call first):
  /usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  cmake_modules/FindLz4.cmake:76 (find_package_handle_standard_args)
  cmake_modules/ThirdpartyToolchain.cmake:241 (find_package)
  cmake_modules/ThirdpartyToolchain.cmake:2143 (resolve_dependency)
  CMakeLists.txt:546 (include)
```

If I specify `LZ4_ROOT` like so `-DLZ4_ROOT=/<REDACTED>/.xmake/packages/l/lz4/v1.9.3/55e9aa7ed4014be287ab80a488dfae47`, it configures correctly.

```
-- Found Lz4: /<REDACTED>/.xmake/packages/l/lz4/v1.9.3/55e9aa7ed4014be287ab80a488dfae47/lib/liblz4.a  
-- Providing CMake module for Lz4
```

Now, providing `LZ4_ROOT=` is an easy fix, even tho a more generic solution would be preferable.

For this reason, I kept looking...

And I've discovered that when **not** using `LZ4_ROOT`, after entering the `else` (and thanks to your last fix) it enters the `if(LZ4_PC_FOUND)` [branch](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L57).

At this point `find_library` fails to provide `LZ4_LIB` which is **why** after at line 76 `find_package_handle_standard_args` complains so much. So at least we know where the actual problem is...

waruqi:
> And I've discovered that when not using LZ4_ROOT, after entering the else (and thanks to your last fix) it enters the if(LZ4_PC_FOUND) [branch](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L57).
At this point find_library fails to provide LZ4_LIB which is why after at line 76 find_package_handle_standard_args complains so much. So at least we know where the actual problem is...

Is @cyfdecyf  latest fix not working either? try reinstalling lz4 first.

```console
xrepo remove --all -y lz4
```

cyfdecyf:
@leodido If you have already tried reinstalling lz4 and still got the problem. Take a look of this fix https://github.com/microsoft/vcpkg/pull/14805/files#diff-2949f257677b5b14a643f87e15b452f2931b0d7e50cdf153b339a930f299270aR18

I encounter similar problems for libevent after fixing the `_INSTALL_PREFIX`, but then the `NO_DEFAULT_PATH` argument  causes problems for finding header and lib in the **correct** path (I added debug message to check the path).

It's can be frustrating to debug cmake problems. And I still don't know why the `NO_DEFAULT_PATH` need to be removed after I read the cmake [document](https://cmake.org/cmake/help/latest/command/find_library.html).

I'll send fix for libevent tonight and that fix can be a reference for changing cmake files before install.

leodido:
> > And I've discovered that when not using LZ4_ROOT, after entering the else (and thanks to your last fix) it enters the if(LZ4_PC_FOUND) [branch](https://github.com/apache/arrow/blob/9aaf61c044189d4d15bd145380fb4e2a2a65138f/cpp/cmake_modules/FindLz4.cmake#L57).
> > At this point find_library fails to provide LZ4_LIB which is why after at line 76 find_package_handle_standard_args complains so much. So at least we know where the actual problem is...
> 
> Is @cyfdecyf latest fix not working either? try reinstalling lz4 first.
> 
> ```
> xrepo remove --all -y lz4
> xrepo install lz4
> ```

I did of course :)

PR #1124, opened 2022-03-30
rapidjson: add virtual version v1.1.0-arrow

cyfdecyf:


packages/r/rapidjson/xmake.lua
```diff
@@ -7,6 +7,8 @@ package("rapidjson")
              "https://github.com/Tencent/rapidjson.git")
 
     add_versions("v1.1.0", "8e00c38829d6785a2dfb951bb87c6974fa07dfe488aa5b25deec4b8bc0f6a3ab")
+    -- This commit is used in arrow 7.0.0 https://github.com/apache/arrow/blob/release-7.0.0/cpp/thirdparty/versions.txt#L80
+    add_versions("v1.1.0-arrow", "1a803826f1197b5e30703afe4b9c0e7dd48074f5")
```
leodido, line 11: The second argument of `add_versions` shouldn't be the sha256 checksum or it can also be a commit?

leodido, line 11: The second argument of `add_versions` shouldn't be the sha256 checksum or it can also be a commit?

waruqi: right, it is git commit or sha256

PR #1129, opened 2022-04-01
Add Fmi4cpp

PucklaMotzer09:


PucklaMotzer09:
On windows the cmake build of the library itself fails because it doesn't link to zlib.

packages/f/fmi4cpp/xmake.lua
```diff
@@ -0,0 +1,73 @@
...
+    add_versions("0.8.0", "78616e9c86a23137a8d3a113fe6420207c3f9ea46442e1c75a01215eb2693bb7")
+
+    add_patches("0.8.0", path.join(os.scriptdir(), "patches", "0.8.0", "clang_fix.patch"), "dacd893e90298763223b21b0054dad6d6a82c7c36ab0d3d0cc1984a342c01f9f")
+    add_patches("0.8.0", path.join(os.scriptdir(), "patches", "0.8.0", "win32_zlib.patch"), "99d14ebf2f1d7b848ab5fc5b659826d50429e59810f13b25953fddfc8f4313b7")
+
+    add_deps("cmake", "boost", "libzip")
+
+    on_install("linux", function (package)
+        local configs = {
+            "-DFMI4CPP_BUILD_EXAMPLES=OFF"
```
waruqi, line 16: please add shared/static and debug configs

PucklaMotzer09: I implemented this using the `BUILD_SHARED_LIBS` and `CMAKE_BUILD_TYPE` variables. Those seem to work for any cmake project. So I think that these could be utilised by xmake itself to configure this.

PR #1130, opened 2022-04-01
Update xmake.lua

Force67:
Sentry native on windows requires crashpad by default. This can be enabled thorugh the flag ive added in the build script, (see here: https://docs.sentry.io/platforms/native/configuration/backends/crashpad/) but im not sure if this is how its supposed to be implemented in xmake. Please gimme some input on how this is supposed to work :D.


waruqi:
maybe you need add frameworks and syslinks for unresolved external symbol

yamashi:
This can be closed, I have fixed that in another merged PR

PR #1134, opened 2022-04-03
Add missing package_end() in sfml xmake script

2nafish117:
* Add missing package_end() in sfml xmake script


waruqi:
it is optional, we need not it in xmake-repo

2nafish117:
I copied this package script to build sfml as package from local cmake source, and it was required for that. 
This change could be useful for others copying the script as well.

PR #1182, opened 2022-04-23
[Volk] update to latest version andd add header only support

Arthapz:


Arthapz:
iphoneos checks should be ok after https://github.com/xmake-io/xmake-repo/pull/1183

PR #1185, opened 2022-04-25
[Volk] Fix test when use header_only config

Arthapz:


packages/v/volk/xmake.lua
```diff
@@ -51,5 +51,16 @@ package("volk")
...
+            header_only_implementation = "#define VOLK_IMPLEMENTATION"
+        end
+
+        assert(package:check_csnippets({test = format([[
+            %s
+            #include <volk.h>
+            void test() {
+                volkInitialize();
+            }
+        ]], header_only_implementation)}))
```
waruqi, line 65:
```lua
        local defines
        if package:config("header_only") then 
            defines = "VOLK_IMPLEMENTATION"
        end

        assert(package:check_csnippets({test = format([[
            #include <volk.h>
            void test() {
                volkInitialize();
            }
        ]], {configs = {defines = defines}})}))
```

Arthapz:
done

PR #1187, opened 2022-04-26
Adds vulkan headers version

zheng95z:
Adds a new version (v1.3.212) of vulkan headers.

### DIFF
--- a/packages/v/vulkan-headers/xmake.lua
+++ b/packages/v/vulkan-headers/xmake.lua
@@ -6,7 +6,7 @@ package("vulkan-headers")
     set_license("Apache-2.0")
 
     add_urls("https://github.com/KhronosGroup/Vulkan-Headers/archive/$(version).tar.gz", {version = function (version) return version:startswith("v") and version or "sdk-" .. version:gsub("%+", ".") end})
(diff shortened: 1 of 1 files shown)


packages/v/vulkan-headers/xmake.lua
```diff
@@ -6,6 +6,7 @@ package("vulkan-headers")
     set_license("Apache-2.0")
 
     add_urls("https://github.com/KhronosGroup/Vulkan-Headers/archive/$(version).tar.gz", {version = function (version) return version:startswith("v") and version or "sdk-" .. version:gsub("%+", ".") end})
+    add_versions("1.3.212", "deea32809940711be36258e17ac887818f6261c96be27b24a1b82ac6718f0b2b")
```
waruqi, line 9: we only support version with `sdk-` .

[sdk-1.3.211.0](https://github.com/KhronosGroup/Vulkan-Headers/releases/tag/sdk-1.3.211.0)

>     add_urls("https://github.com/KhronosGroup/Vulkan-Headers/archive/$(version).tar.gz", {version = function (version) return version:startswith("v") and version or "sdk-" .. version:gsub("%+", ".") end})

PR #1195, opened 2022-04-28
Adds support for pcg32 library

zheng95z:


packages/p/pcg32/xmake.lua
```diff
@@ -0,0 +1,22 @@
+package("pcg32")
+    set_homepage("https://github.com/wjakob/pcg32")
+    set_description("Tiny self-contained C++ version of the PCG32 pseudorandom number generator")
+    set_license("Apache-2.0")
+
+    add_urls("https://github.com/wjakob/pcg32.git")
```
waruqi, line 6: please add a fakse version if no any releases/tags

```lua
add_versions("2016.06.07", "70099eadb86d3999c38cf69d2c55f8adc1f7fe34")
```

zheng95z:
I also changed the way of adding a head-only library since a lot of people prefer `pcg32/pcg32.h` instead of `pcg32.h`.

packages/p/pcg32/xmake.lua
zheng95z: Done.

waruqi:
> I also changed the way of adding a head-only library since a lot of people prefer `pcg32/pcg32.h` instead of `pcg32.h`.
> 
> And I'm wondering whether there is a way can do that inside:
> 
> ```lua
> io.writefile("xmake.lua", [[
>     target("pcg32")
>         set_kind("headeronly")
>         add_headerfiles("pcg32.h")
> ]])
> import("package.tools.xmake").install(package)
> ```

need only os.cp

PR #1197, opened 2022-04-28
Add mingw support to tmxparser

PucklaMotzer09:


PucklaMotzer09:
The shared windows builds are failing because the `*.dll` files are moved into `bin` not like `*.so` files which are moved into `lib`. Therefore xmake doesn't detect them, because they need to be inside `lib`.

waruqi:
> The shared windows builds are failing because the `*.dll` files are moved into `bin` not like `*.so` files which are moved into `lib`. Therefore xmake doesn't detect them, because they need to be inside `lib`.

you can add `add_rules("utils.symbols.export_all")` in xmake.lua

https://github.com/xmake-io/xmake-repo/blob/196c6317c83b55755ad9b825ef9534769326c4ad/packages/n/nanovg/xmake.lua#L18-L20

PR #1199, opened 2022-04-28
Adds support for CLI11 library

zheng95z:


packages/c/cli11/xmake.lua
```diff
@@ -0,0 +1,19 @@
...
+    set_description("CLI11 is a command line parser for C++11 and beyond that provides a rich feature set with a simple and intuitive interface.")
+    set_license("BSD")
+
+    add_urls("https://github.com/CLIUtils/CLI11/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/CLIUtils/CLI11.git")
+
+    add_versions("v2.2.0", "d60440dc4d43255f872d174e416705f56ba40589f6eb07727f76376fb8378fd6")
+
+    on_install("windows", "linux", "macosx", function (package)
+        os.cp("include/CLI", package:installdir("include"))
```
leodido, line 12:
```suggestion
        os.cp("include/**.hpp", package:installdir(path.join("include", "CLI")))
```

I suggest doing it like this (because there are inner headers the user may want to use). Feel free to pick up and commit my suggestion.

leodido (commented):
Please see my suggestion

packages/c/cli11/xmake.lua
zheng95z: It seems more promising. Thanks!

packages/c/cli11/xmake.lua
```diff
@@ -0,0 +1,19 @@
...
+    add_versions("v2.2.0", "d60440dc4d43255f872d174e416705f56ba40589f6eb07727f76376fb8378fd6")
+
+    on_install("windows", "linux", "macosx", function (package)
+        os.cp("include/**.hpp", package:installdir(path.join("include", "CLI")))
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            CLI::App app{"Test", "test"};
+        ]]}, {configs = {languages = "cxx17"}, includes = "CLI/CLI.hpp"}))
```
leodido, line 18: Sorry for the other nitpick but I believe it would be better to use `cxx11` (or equivalent) here., being the library build for C++11 (onwards).
```suggestion
        ]]}, {configs = {languages = "cxx11"}, includes = "CLI/CLI.hpp"}))
```

leodido (commented):
Another comment :)

packages/c/cli11/xmake.lua
zheng95z: Thanks.

leodido:
LGTM

leodido (approved):

PR #1201, opened 2022-04-30
Update workflow and libwfkafka to v0.10.1

Barenboim:


packages/l/libwfkafka/xmake.lua
```diff
@@ -5,7 +5,7 @@ package("libwfkafka")
 
     add_urls("https://github.com/sogou/workflow/archive/refs/tags/$(version).tar.gz",
              "https://github.com/sogou/workflow.git")
-    add_versions("v0.9.11", "71b5531728d6b4f3666176dbc45d680350518af8")
+    add_versions("v0.10.1", "315eb1b1b5411e807e5ecc45ba5aa7db1d4f7c28")
```
waruqi, line 8: you need only add versions. please do not remove old versions

packages/w/workflow/xmake.lua
```diff
@@ -5,7 +5,7 @@ package("workflow")
 
     add_urls("https://github.com/sogou/workflow/archive/refs/tags/$(version).tar.gz",
              "https://github.com/sogou/workflow.git")
-    add_versions("v0.9.11", "71b5531728d6b4f3666176dbc45d680350518af8")
+    add_versions("v0.10.1", "315eb1b1b5411e807e5ecc45ba5aa7db1d4f7c28")
```
waruqi, line 8: and here

packages/l/libwfkafka/xmake.lua
Barenboim: Done!

PR #1207, opened 2022-05-05
new(packages/a/aws-checksums)

leodido:
Create `aws-checksums` package

### DIFF
--- a/packages/a/aws-checksums/xmake.lua
+++ b/packages/a/aws-checksums/xmake.lua
@@ -3,23 +3,29 @@ package("aws-checksums")
     set_description("Cross platform HW accelerated CRC32c and CRC32 with fallback to efficient SW implementations - C interface with language bindings for AWS SDKs")
     set_license("Apache-2.0")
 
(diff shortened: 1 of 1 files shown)


packages/a/aws-checksums/xmake.lua
```diff
@@ -0,0 +1,28 @@
...
+    if is_plat("linux") then
+      add_deps("cmake", "aws-c-common")
+    end
+
+    on_install("linux", function (package)
+        local configs = {}
+
+        local common_cmake_files = package:dep("aws-c-common"):installdir("lib", "cmake")
+        if os.isdir(common_cmake_files) then
+            os.cp(common_cmake_files .. "/*.cmake", package:installdir("lib", "cmake"))
```
waruqi, line 18: why?

leodido: It doesn't build otherwise (at least locally). It seems to need the content of the cmake/ directory of `aws-c-common`

PR #1222, opened 2022-05-17
Add dpp package

SirLynix:


packages/d/dpp/xmake.lua
```diff
@@ -0,0 +1,48 @@
+package("dpp")
+    set_homepage("https://github.com/brainboxdotcc/DPP")
+    set_description("D++ Extremely Lightweight C++ Discord Library")
+    set_license("Apache-2.0")
+
+    add_urls("https://github.com/brainboxdotcc/DPP/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/brainboxdotcc/DPP.git")
+
+    add_versions("v10.0.8", "7a16d31841fc67fabcafbc33abb1a6b2ac472202df7e8c48542f77e089de08e3")
+    add_patches("v10.0.8", path.join(os.scriptdir(), "patches", "v10.0.8", "dpp_dependencies.patch"), "3e4b943cca885624dd2f7fe103b4c3abf031e8963b9e9f0e96b90b96188d298e")
```
waruqi, line 10: oh, it's a too large patch. ~1.38M

you can set http url to add_patches and remove this patch file in xmake-repo

SirLynix: It's large because it's removing fmt and nlohmann headers from include, I can let them in to reduce patch size.

SirLynix:
I improved it, be sure to squash the commits so we won't have the previous file in history.

PR #1237, opened 2022-05-23
new(packages/p): create prometheus-cpp package

leodido:
This PR creates the package for the `prometheus-cpp` library as per the title.


packages/p/prometheus-cpp/xmake.lua
```diff
@@ -11,16 +11,19 @@ package("prometheus-cpp")
...
+            links = {"prometheus-cpp-push"},
             default = true,
         },
         compression = {
             description = "Enable compression.",
             deps = {["zlib"] = {}},
+            links = nil,
             default = true,
         },
     }
```
waruqi, line 29: Please use add_configs directly to ensure a consistent style

PR #1244, opened 2022-05-30
qt5: fix datarace

SirLynix:
When using already installed Qt package, I noticed that calling find_qt would pause the coroutine and allows another one to perform the fetch, causing issue:

- qt5base: fetch
- qt5base: <pause because of find_qt>
- qt5core: fetch (before qt5base finished)
- qt5core: tries to retrieve Qt info before qt5base finished - error

The problem doesn't happen when in verbose mode (because no concurrency happens)

To fix this I made it so qt5core fetches qt5base, but it was causing multiple simultaneous find_qt calls.

To fix this, I added some kind of "mutex" around the find_qt call in qt5base, so only one is made then the result is cached.

I think this is an issue inside of xmake itself, as it shouldn't try to fetch packages before their dependencies are fetched, but this is a working temporary fix.

### DIFF
--- a/packages/q/qt5base/xmake.lua
+++ b/packages/q/qt5base/xmake.lua
@@ -13,6 +13,8 @@ local function qt_table(sdkdir, version)
     }
 end
 
+local qt_fetch = nil
+
 package("qt5base")
     set_kind("phony")
     set_homepage("https://www.qt.io")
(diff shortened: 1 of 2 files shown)


waruqi:
> I think this is an issue inside of xmake itself, as it shouldn't try to fetch packages before their dependencies are fetched, but this is a working temporary fix.

I agree that a better fix is needed here, but most of the packages that have dependencies, they fetch without any dependencies as well. If the global fetch is based on dependencies, this will affect the speed of the fetch.

I think there should be a configuration policy to disable parallel fetch for specific packages

SirLynix:
I can confirm the PR works, thanks.
So yeah, no need to merge this PR.

PR #1252, opened 2022-06-05
add libargh

myuanz:


SirLynix:
You should rebase the target branch to `dev`, also please add some tests

myuanz:
Is it appropriate to run the tests from the official repository?

SirLynix:
It would be better to add a `on_test` which checks if a function exists, or a code compiles, check other packages for examples

myuanz:
Thanks

PR #1253, opened 2022-06-05
Add argh

myuanz:
I didn't figure out how to run the official binary test in `on_test`, but it's a very refreshing header-only project, so if it gets compiled successfully it will run fine.

### DIFF
--- a/packages/a/argh/xmake.lua
+++ b/packages/a/argh/xmake.lua
@@ -3,44 +3,38 @@ package("argh")
     set_description("Argh! A minimalist argument handler.")
     set_license("BSD-3-Clause")
 
-    set_urls("https://github.com/adishavit/argh.git")
+    set_kind("library", {headeronly = true})
+
+    add_urls(
+        "https://github.com/adishavit/argh/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/adishavit/argh.git"
+    )
+    add_versions("v1.3.2", "4b76d8c55e97cc0752feee4f00b99dc58464dd030dea9ba257c0a7d24a84f9dd")
+
     add_deps("cmake")
 
     on_install(function (package)
         import("package.tools.cmake").install(package)
     end)
+
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
+            // Modified from https://github.com/adishavit/argh/blob/master/example.cpp
             #include <iostream>
-
             using namespace std;
-
             #include "argh.h"
-
-            int test(int argc, char* argv[])
-            {
+            int test(int argc, char* argv[]) {
                 argh::parser cmdl;
                 cmdl.parse(argc, argv, argh::parser::PREFER_PARAM_FOR_UNREG_OPTION);
-
-                if (cmdl["-v"])
-                    cout << "Verbose, I am." << endl;
-
+                if (cmdl["-v"]) cout << "Verbose, I am." << endl;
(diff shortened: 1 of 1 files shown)


packages/a/argh/xmake.lua
```diff
@@ -0,0 +1,48 @@
...
+
+                cout << "\nFlags:\n";
+                for (auto& flag : cmdl.flags())
+                    cout << '\t' << flag << endl;
+
+                cout << "\nParameters:\n";
+                for (auto& param : cmdl.params())
+                    cout << '\t' << param.first << " : " << param.second << endl;
+
+                return EXIT_SUCCESS;
```
waruqi, line 44: oh, too large test code, please simplify it.

packages/a/argh/xmake.lua
```diff
@@ -0,0 +1,48 @@
+package("argh")
```
waruqi, line 1: headeroly? please add `set_kind("library", {headeronly = true})`

packages/a/argh/xmake.lua
```diff
@@ -0,0 +1,48 @@
...
+    set_homepage("https://github.com/adishavit/argh")
+    set_description("Argh! A minimalist argument handler.")
+    set_license("BSD-3-Clause")
+
+    set_urls("https://github.com/adishavit/argh.git")
```
waruqi, line 6:
```lua
add_urls("https://github.com/adishavit/argh/archive/refs/tags/$(version).tar.gz",
    "https://github.com/adishavit/argh.git")

set_versions("v1.3.2", "<sha256 of tar.gz>")
```

packages/a/argh/xmake.lua
```diff
@@ -0,0 +1,48 @@
...
+    add_deps("cmake")
+
+    on_install(function (package)
+        import("package.tools.cmake").install(package)
+    end)
```
waruqi, line 11: add a newline.

packages/a/argh/xmake.lua
```diff
@@ -0,0 +1,48 @@
+package("argh")
+    set_homepage("https://github.com/adishavit/argh")
+    set_description("Argh! A minimalist argument handler.")
+    set_license("BSD-3-Clause")
+
+    set_urls("https://github.com/adishavit/argh.git")
+    add_deps("cmake")
+
+    on_install(function (package)
+        import("package.tools.cmake").install(package)
```
waruqi, line 10: add configs to disable test and examples

BUILD_TESTS=OFF, BUILD_EXAMPLES=OFF

or only use `os.cp("args.h", package:installdir("include"))`

packages/a/argh/xmake.lua
myuanz: I agree, but it is from the official origin example (https://github.com/adishavit/argh/blob/master/example.cpp). Do I need to remove blank lines?

SirLynix: on_test purpose is mainly to check if the library can be properly included/used (and linked if not header only), its code is not even executed. Its role is not to test the library but to test the package has been properly built

packages/a/argh/xmake.lua
myuanz: It's nice to get rid of cmake

packages/a/argh/xmake.lua
myuanz: Got

PR #1277, opened 2022-06-17
add package: zycore-c

yszhou2019:
add package: zycore-c, using as the dependency for package "zydis"

i've already run ci in my fork version. it works fine.


waruqi:
thanks!

PR #1291, opened 2022-06-23
Add fast_float package

SirLynix:


packages/f/fast_float/xmake.lua
```diff
@@ -0,0 +1,22 @@
+package("fast_float")
```
waruqi, line 1: add headeronly

waruqi:
c++11

PR #1317, opened 2022-07-01
fix openssl package nil crash on windows.

Zmmfly:
Issue: https://github.com/xmake-io/xmake-repo/issues/1315


waruqi:
你得 push 到 dev 分支才会触发 ci tests，可以重新 force push 下，或者重开个 pr

PR #1345, opened 2022-07-12
Add extsources to brotli

ImperatorS79:
@waruqi on ubuntu, the executable `brotli` is in its own package separated from the rest (`libbrotli-dev`). I do not know how the xmake repository handles this.


waruqi:
If it is already installed in the system bin directory, we don't need to do any additional processing.

PR #1350, opened 2022-07-15
Add extsources to ffmpeg

ImperatorS79:
**Question**: ffmpeg dev files on ubuntu are splitted into multiple packages. So instead of installing one pacman package named `ffmpeg` (on msys2, brew, arch linux for example) that will contain the executable and all the dev files, you need to install multiple packages on ubuntu:
```
ffmpeg, libavcodec-dev, libavdevice-dev, libavfilter-dev, libavformat-dev, libavutil-dev, libpostproc-dev, libswresample-dev, libswscale-dev
```
That means it should be possible, using `add_extsources`, to require multiple packages to be installed to consider the dependency ffmpeg installed, something that could look like:

```lua
add_extsources("pacman::ffmpeg", {"apt::ffmpeg", "apt::libavcodec-dev", "apt::libavdevice-dev", "apt::libavfilter-dev", "apt::libavfilter-dev", "apt::libavformat-dev", "apt::libavutil-dev", "apt::lib-postproc-dev", "apt::libswresample-dev", "apt::libswscale-dev"})
```

Which means: either you have the pacman package ffmpeg installed (arch linux and the likes), or the following deb packages (debian, mint, ubuntu, ...). Is it something possible with xmake ?

### DIFF
--- a/packages/f/ffmpeg/xmake.lua
+++ b/packages/f/ffmpeg/xmake.lua
@@ -60,7 +60,7 @@ package("ffmpeg")
                             result = result .. pkginfo
                         end
                     else
-                        return {}
+                        return nil
                     end
                 end
                 local ffmpeg = find_tool("ffmpeg", {check = "-help", version = true, command = "-version", parse = "%d+%.?%d+%.?%d+", force = true})
--- a/packages/f/ffmpeg/xmake.lua
+++ b/packages/f/ffmpeg/xmake.lua
@@ -60,7 +60,7 @@ package("ffmpeg")
                             result = result .. pkginfo
                         end
                     else
-                        return nil
+                        return
                     end
                 end
                 local ffmpeg = find_tool("ffmpeg", {check = "-help", version = true, command = "-version", parse = "%d+%.?%d+%.?%d+", force = true})
--- a/packages/f/ffmpeg/xmake.lua
+++ b/packages/f/ffmpeg/xmake.lua
@@ -75,15 +75,13 @@ package("ffmpeg")
                 end
                 result.includedirs = table.unique(result.includedirs)
                 result.linkdirs = table.unique(result.linkdirs)
-                import("lib.detect.find_programver")
                 
+                import("lib.detect.find_programver")
                 local version = find_programver("ffmpeg", {command = "-version"})
                 if(version == nil) then
                     return
                 end
-                    
                 result.version = version:gsub("(.*)%-.*$","%1") --remove -0ubuntu0.22.04.1 on ubuntu
-                
                 return result
             end
         end)
--- a/packages/f/ffmpeg/xmake.lua
+++ b/packages/f/ffmpeg/xmake.lua
@@ -52,6 +52,31 @@ package("ffmpeg")
     elseif is_plat("macosx") then
         add_extsources("brew::ffmpeg")
     end
+    
+    if on_fetch then
+        on_fetch("linux", function (package, opt)
+            if opt.system then
+                local result
+                for _, name in ipairs({"libavcodec", "libavdevice", "libavfilter", "libavformat", "libavutil", "libpostproc", "libswresample", "libswscale"}) do
+                    local pkginfo = package.find_package and package:find_package("pkgconfig::" .. name, opt)
+                    if pkginfo then
+                        if not result then
+                            result = table.copy(pkginfo)
+                        else
+                            local includedirs = pkginfo.sysincludedirs or pkginfo.includedirs
+                            result.links = table.wrap(result.links)
+                            result.linkdirs = table.wrap(result.linkdirs)
+                            result.includedirs = table.wrap(result.includedirs)
+                            table.join2(result.includedirs, includedirs)
+                            table.join2(result.linkdirs, pkginfo.linkdirs)
+                            table.join2(result.links, pkginfo.links)
+                        end
+                    end
+                end
+                return result
+            end
+        end)
+    end
 
     on_load("linux", "macos", "android", function (package)
         local configdeps = {zlib    = "zlib",
--- a/packages/f/ffmpeg/xmake.lua
+++ b/packages/f/ffmpeg/xmake.lua
@@ -45,6 +45,14 @@ package("ffmpeg")
         add_deps("yasm")
     end
 
+    if is_plat("mingw") and is_subhost("msys") then
+        add_extsources("pacman::ffmpeg")
+    elseif is_plat("linux") then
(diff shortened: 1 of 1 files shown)


waruqi:
you should use on_fetch to find multiple libs

https://github.com/xmake-io/xmake-repo/blob/70534ac88fc90306edda841add4b860338eeb482/packages/b/brotli/xmake.lua#L38

ImperatorS79:
What is the content of `opt` in the function passed to `on_fetch` ? How can I limit that to apt based system ?

waruqi:
> What is the content of `opt` in the function passed to `on_fetch` ? How can I limit that to apt based system ?

package:find_package("apt:libavcodec-dev", ..)

but you should use `pkgconfig::xxx`, it is more common.

ImperatorS79:
Okay so I adapted the code from `brotli`, I hope it will work.

waruqi:
> Okay so I adapted the code from `brotli`, I hope it will work.

You can test it on your host first.

waruqi:
use `package:find_tool`

> ```lua
> import("lib.detect.find_tool")
> ```

It works for me, and you can use `package:find_tool`

Also, we should be looking for the library version rather than the ffmpeg binary version.

ImperatorS79:
There is no ffmpeg library. All the librairies for which I look for the .pc files have different versions, the version added in `add_version` corresponds to the version of the binary.

packages/f/ffmpeg/xmake.lua
```diff
@@ -45,6 +45,50 @@ package("ffmpeg")
...
+                    end
+                end
+                result.includedirs = table.unique(result.includedirs)
+                result.linkdirs = table.unique(result.linkdirs)
+                import("lib.detect.find_programver")
+                
+                local version = find_programver("ffmpeg", {command = "-version"})
+                if(version == nil) then
+                    return
+                end
```
waruqi, line 83: we should use find_tool instead of find_programver

packages/f/ffmpeg/xmake.lua
```diff
@@ -45,6 +45,50 @@ package("ffmpeg")
...
+                    
+                result.version = version:gsub("(.*)%-.*$","%1") --remove -0ubuntu0.22.04.1 on ubuntu
+
```
waruqi, line 86: please remove empty lines

packages/f/ffmpeg/xmake.lua
ImperatorS79: find_tool does not work...

ImperatorS79: How can I override the version command like in `find_programver`  with `find_tool`?

waruqi:
```lua
 find_tool("ffmpeg", {check = "-help", version = true, command = "-version", parse = "%d+%.?%d+%.?%d+", force = true})
```

waruqi:
```lua

    on_fetch("linux", function (package, opt)
        import("lib.detect.find_tool")
        if opt.system then
            local result
            for _, name in ipairs({"libavcodec", "libavdevice", "libavfilter", "libavformat", "libavutil", "libpostproc", "libswresample", "libswscale"}) do
                local pkginfo = package:find_package("pkgconfig::" .. name, opt)
                if pkginfo then
                    pkginfo.version = nil
                    if not result then
                        result = pkginfo
                    else
                        result = result .. pkginfo
                    end
                end
            end
            if result then
                local ffmpeg = find_tool("ffmpeg", {check = "-help", version = true, command = "-version", parse = "%d+%.?%d+%.?%d+", force = true})
                if ffmpeg then
                    result.version = ffmpeg.version
                end
            end
            return result
        end
    end)
```

```console
{ 
  version = "4.3.2",
  sysincludedirs = { 
    "/usr/include/x86_64-linux-gnu" 
  },
  links = { 
    "avcodec",
    "avformat",
    "avutil",
    "swresample" 
  },
  linkdirs = { 
    "/usr/lib/x86_64-linux-gnu" 
  } 
}
```

ImperatorS79: Done. I switched to using `on_fetch` for every platform to have the right ffmpeg version.

I also return nothing if one of the lib is not installed.

packages/f/ffmpeg/xmake.lua
```diff
@@ -45,6 +45,33 @@ package("ffmpeg")
...
+                        if not result then
+                            result = pkginfo
+                        else
+                            result = result .. pkginfo
+                        end
+                    else
+                        return {}
+                    end
+                end
+                local ffmpeg = find_tool("ffmpeg", {check = "-help", version = true, command = "-version", parse = "%d+%.?%d+%.?%d+", force = true})
```
waruqi, line 66:
```lua
if result then
    local ffmpeg = ...
end

File: packages/f/ffmpeg/xmake.lua [hunk 4]
```diff
@@ -45,6 +45,33 @@ package("ffmpeg")
...
+                    local pkginfo = package:find_package("pkgconfig::" .. name, opt)
+                    if pkginfo then
+                        pkginfo.version = nil
+                        if not result then
+                            result = pkginfo
+                        else
+                            result = result .. pkginfo
+                        end
+                    else
+                        return {}
```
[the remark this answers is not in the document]
ImperatorS79: Ok for nil but if the `else` is removed you could end up with a partially filled `result` (because some libs would be installed and some not) that will be considered valid, if I understand the code correctly.

packages/f/ffmpeg/xmake.lua
waruqi: > Because of the else inside the loop, result is always defined when the code here is executed, so no need of the if here right ?

if all sub-libraries are not found, pkginfo is always nil, so result is nil too.

ImperatorS79: As soon as one pkginfo is nil, the return at line 63 will be triggered and we will never arrives here.

packages/f/ffmpeg/xmake.lua
```diff
@@ -45,6 +45,33 @@ package("ffmpeg")
...
+                    local pkginfo = package:find_package("pkgconfig::" .. name, opt)
+                    if pkginfo then
+                        pkginfo.version = nil
+                        if not result then
+                            result = pkginfo
+                        else
+                            result = result .. pkginfo
+                        end
+                    else
+                        return nil
```
waruqi, line 63: use `return` instead of `return nil`

PR #1358, opened 2022-07-20
add new version to flecs

happyxianyu:
As title.


waruqi:
```lua
if is_plat("linux") then
    add_syslinks("pthread")
end
```

waruqi:
see https://github.com/xmake-io/xmake-repo/pull/1381

PR #1367, opened 2022-07-27
Update freetype and use cmake on Linux/macosx

SirLynix:
The makefile can't find brotli when installed by xmake: <https://github.com/NazaraEngine/xmake-repo/runs/7544869084?check_suite_focus=true#step:5:1026>

I think it's same to use cmake to build it (as this is what vcpkg does)


SirLynix:
Well I wanted to see on which platform freetype failed to build but it seems it builds on all of them 😄

PR #1371, opened 2022-07-29
Added 1.12.0 and 1.12.1 version of gTest

Chupakabra0:
New versions of googletests


xq114:
@waruqi The environment setup failed on archlinux. Seems it's not a matter with the package

PR #1381, opened 2022-08-02
new stable flecs version

FranekStratovarius:


waruqi:
see https://github.com/xmake-io/xmake-repo/blob/a4b95529d90d053c5291ca3d373b1b7e452a46ad/packages/l/libargon2/xmake.lua#L11

FranekStratovarius:
ok, i found the problem. the library itself compiles without a problem, but the test needs to be linked to pthread and also include `pthread.h`. how can i link the test to pthread?

waruqi:
> see
> 
> https://github.com/xmake-io/xmake-repo/blob/a4b95529d90d053c5291ca3d373b1b7e452a46ad/packages/l/libargon2/xmake.lua#L11

just link issues, try this

FranekStratovarius:
i would call that good enough. i unfortunately have no mac to test it there and no clue what the error is

packages/f/flecs/xmake.lua
```diff
@@ -19,7 +23,7 @@ package("flecs")
 
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
-			#inlcude <pthread.h>
+			#include <pthread.h>
```
waruqi: please remove it

packages/f/flecs/xmake.lua
```diff
@@ -10,6 +10,10 @@ package("flecs")
 
     add_deps("cmake")
 
+	if is_plat("linux") then
```
waruqi, line 13: incorrect indent

waruqi:
include monitor.h is missing in rest.c

https://github.com/SanderMertens/flecs/blob/b0b6bb7892eaffa36555c616893f1c9891998600/include/flecs/addons/monitor.h

FranekStratovarius:
okay, that looks good. every platform compiles without errors.

i have just one question: i have a different sha256 hash when i download https://github.com/SanderMertens/flecs/archive/refs/tags/v3.0.0.tar.gz than the one you changed. do i have the wrong file or what hash is here needed?

and why does it fail on mac when downloading the correct .tar.gz, but works without an error when just downloading the repo itself?

waruqi:
I updated 3.0.0 to the latest git commit number because the flecs author fixed the mac link issue yesterday.

xmake will try to download the tar.gz first, and although it will fail because of the incorrect sha256, xmake will continue to try to download it from the git url and checkout to the specified git commit.

This does not affect the installation of the package, but may result in one more download.

waruqi:
When flecs releases a new version, we'll go back and update and fix the tar.gz

PR #1413, opened 2022-08-17
Add unordered_dense and deprecate robin-hood-hasing

Arthapz:


packages/r/robin-hood-hashing/xmake.lua
```diff
@@ -10,7 +10,12 @@ package("robin-hood-hashing")
     add_versions("3.11.3", "dcf2b7fa9ef9dd0c67102d94c28e8df3effbe1845e0ed1f31f4772ca5e857fc4")
     add_versions("3.11.5", "7aa183252527ded7f46186c1e2f4efe7d6139a3b7c0869c1b6051bd7260587ed")
 
-    add_deps("cmake")
+    add_deps("meson")
```
SirLynix, line 13: ?

Arthapz: c'etait un fail =P

Arthapz:
i don't understand why this doesn't work :/

waruqi:
> i don't understand why this doesn't work :/

it's cmake and meson scripts have not install header configuration. 

In addition, using meson introduces additional python/openssl/ninja dependencies and builds a lot of tests, which is just a headeronly library. It is faster to use cmake to install it here.

PR #1444, opened 2022-08-29
Add github link and update libsdl*

SFGrenade:


SFGrenade:
both tested and worked locally, with and without `--shallow`

packages/l/libsdl_image/xmake.lua
```diff
@@ -50,23 +52,17 @@ package("libsdl_image")
...
+        table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes"))
+        if package:is_plat("linux") and package:config("pic") ~= false then
+            table.insert(configs, "--with-pic")
+        end
         local libsdl = package:dep("libsdl")
         if libsdl and not libsdl:is_system() then
             table.insert(configs, "--with-sdl-prefix=" .. libsdl:installdir())
         end
-        if package:version():eq("2.0.5") then
-            io.replace("Makefile.am", "noinst_PROGRAMS = showimage", "")
```
waruqi: We need to be compatible with older versions, 2.0.5 need this patch

SFGrenade: how do i tell the test script to specifically install the 2.0.5 version to test it?

waruqi: test.lua "libsdl_image 2.0.5"

SFGrenade: so shallow and non-shallow both work at 2.0.5, but i can add it back

waruqi: It will fix `link showanim binary error`, it still does not work on macos ci now.

Maybe we need also disable it in new version.

SFGrenade: i'm kinda stumped why it fails with `configure: error: C compiler cannot create executables` for linux, or `.libs/libSDL2_image.a(IMG_tga.o) has no symbols` for mac

waruqi: I fixed it, but has an anther error on macOS.


```lua
Undefined symbols for architecture x86_64:
  "_IMG_Init", referenced from:
      _main in _C2F4F90BE61F471084D12FE0A7112520.o
```

packages/l/libsdl_image/xmake.lua
```diff
@@ -43,19 +52,16 @@ package("libsdl_image")
...
-        end
+        table.insert(configs, "--enable-shared=" .. (package:config("shared") and "yes" or "no"))
+        table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes"))
         local libsdl = package:dep("libsdl")
         if libsdl and not libsdl:is_system() then
             table.insert(configs, "--with-sdl-prefix=" .. libsdl:installdir())
         end
-        io.replace("Makefile.am", "noinst_PROGRAMS = showimage", "")
+        if libsdl and package:is_plat("macosx") then
+            table.insert(configs, "-DSDL2_DIR=" .. libsdl:installdir())
```
waruqi, line 62: configure: error: unrecognized option: `-DSDL2_DIR=/Users/runner/.xmake/packages/l/libsdl/2.24.0/452ad75fc0274e7eb90336efbef13ba2'
Try `./configure --help' for more information

packages/l/libsdl_image/xmake.lua
```diff
@@ -43,19 +52,15 @@ package("libsdl_image")
...
-        if package:is_plat("linux") and package:config("pic") ~= false then
-            table.insert(configs, "--with-pic")
-        end
+        table.insert(configs, "--enable-shared=" .. (package:config("shared") and "yes" or "no"))
+        table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes"))
         local libsdl = package:dep("libsdl")
         if libsdl and not libsdl:is_system() then
             table.insert(configs, "--with-sdl-prefix=" .. libsdl:installdir())
+        elseif libsdl and package:is_plat("macosx") then
+            table.insert(configs, "--with-sdl-prefix=" .. libsdl:installdir())
```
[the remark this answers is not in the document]
SFGrenade: though that didn't work before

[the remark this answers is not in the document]
SFGrenade: hm, weird that it still can't find the method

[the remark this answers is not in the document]
SFGrenade: especially since it says `building for macOS-x86_64 but attempting to link with file built for macOS-x86_64`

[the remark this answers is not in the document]
waruqi: After I reinstalled libsdl and use 2.24.0, I was able to reproduce it locally, but I don't know how to fix it.

It works if it use libsdl 2.0.22, but it does not work for libsdl 2.24.0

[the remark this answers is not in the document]
waruqi: > especially since it says `building for macOS-x86_64 but attempting to link with file built for macOS-x86_64`

then libSDL2_image.a will be ignore when doing link.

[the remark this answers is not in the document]
waruqi: It should be configure's bug. I have used cmake to fix it, but other libsdl project need use cmake too.

PR #1446, opened 2022-08-30
Making sure that debug builds work

talhasaruhan:
on_test now calls appropriate opencv_version(d) executable upon installing. Previously the debug builds failed because the executable couldn't be found.

### DIFF
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -202,7 +202,7 @@ package("opencv")
             runtest = false
         end
         if runtest then
-            os.vrun("opencv_version")
+            os.vrun((package:debug() and "opencv_version" or "opencv_versiond"))
         end
         assert(package:check_cxxsnippets({test = [[
             #include <iostream>


packages/o/opencv/xmake.lua
```diff
@@ -202,7 +202,7 @@ package("opencv")
             runtest = false
         end
         if runtest then
-            os.vrun("opencv_version")
+            os.vrun((package:debug() and "opencv_version" or "opencv_versiond"))
```
SirLynix, line 205: this doesn't seems right, as it select "opencv_version" if package:debug() is true, and "opencv_versiond" otherwise.
`os.vrun(package:debug() and "opencv_versiond" or "opencv_version")`

talhasaruhan: Yes that's the whole idea. Because opencv adds "d" at the end of executable and library file names. Actually I realized there needs to be another change, [in this for loop](https://github.com/xmake-io/xmake-repo/blob/master/packages/o/opencv/xmake.lua#L149) and "d" needs to be added to "link" variable as well.

For your information I'm trying to build a project that has a opencv dependency with the following configuration:
```
add_requires("opencv 4.6.0", {debug = true, configs = {vs_runtime = "MTd"}})
xmake f --vs_runtime="MTd" -m "debug"
```

SirLynix: I think you got me wrong, I was saying that you were adding the "d" suffix on **non-**debug builds with your code.

talhasaruhan: This is the only way I could get my build to work with `debug = true` and `vs_runtime="MTd"`. Since otherwise:

1- opencv_version executable can not be found since there's only opencv_versiond executable
2- Even when I bypassed that manually, the installation doesn't copy any libraries so linking fails (since the filenames are wrong)

SirLynix: I'm not discussing the need for this, your adding "d" in release build here

talhasaruhan: Sorry my mistake. Will open another PR with all the correct changes.

PR #1447, opened 2022-08-30
(Kinda) Fixing debug = true, vc_runtime = "MTd" builds

talhasaruhan:
Note that the build is still broken because the check_cxxsnippets doesn't get linked to the right libraries, so I'm hoping I could get some help here. I tried adding `debug = package:debug()` and `vs_runtime = package:config("vc_runtime")` but it didn't help :/


waruqi:
please push to dev branch to trigger ci next.

PR #1448, opened 2022-08-30
Fixed "debug=true" and "vs_runtime="MTd" builds

talhasaruhan:
This allows successful compilation in debug modes (for windows targets)


waruqi:
You don't need to reopen a new pr every time, just keep committing + pushing to an existing branch.

PR #1467, opened 2022-09-07
Added config for CPR_USE_SYSTEM_CURL

talhasaruhan:


packages/c/cpr/xmake.lua
```diff
@@ -10,12 +10,18 @@ package("cpr")
     add_versions("1.7.2", "aa38a414fe2ffc49af13a08b6ab34df825fdd2e7a1213d032d835a779e14176f")
     add_versions("1.8.3", "0784d4c2dbb93a0d3009820b7858976424c56578ce23dcd89d06a1d0bf5fd8e2")
 
+	add_configs("CPR_USE_SYSTEM_CURL", {description = "Sets CPR_USE_SYSTEM_CURL for CMake.", default = true, type = "boolean"})
```
waruqi, line 13: incorrect indent

packages/c/cpr/xmake.lua
```diff
@@ -10,12 +10,18 @@ package("cpr")
...
     if is_plat("mingw") then
         add_syslinks("pthread")
     end
     on_install("linux", "macosx", "windows", "mingw@windows", function (package)
-        local configs = {"-DCPR_BUILD_TESTS=OFF", "-DCPR_ENABLE_SSL=ON", "-DCPR_FORCE_USE_SYSTEM_CURL=ON"}
+        local configs = {
+			"-DCPR_BUILD_TESTS=OFF",
+			"-DCPR_ENABLE_SSL=ON",
+			"-DCPR_USE_SYSTEM_CURL="..(package:config("CPR_USE_SYSTEM_CURL") and "ON" or "OFF")
+		}
```
waruqi, line 24: use space 4 instead of tab

packages/c/cpr/xmake.lua
waruqi: and please rename `CPR_USE_SYSTEM_CURL ` to `system_curl`

talhasaruhan:
Not sure why "Archlinux / build (ubuntu-latest, static) (pull_request)" is failing :/ Perhaps adding -lssh2 to linker flags would help?

waruqi:
We should always enable `CPR_USE_SYSTEM_CURL=ON` and need not add any configuration, because we need use `add_deps("libcurl")` instead of `cmake/fetchcontent curl`

PR #1475, opened 2022-09-12
gRPC 1.48.1

yamashi:


yszhou2019:
Hi, @yamashi I noticed that you've added a new version to grpc in xmake-repo. It seems compilation encountered some problems in msvc 2019. Can you figure out the problem?

waruqi:
it need higher winsdk and gcc now.

https://github.com/xmake-io/xmake-repo/pull/1912

PR #1511, opened 2022-10-02
Fix SFML Freetype links on Windows

SirLynix:


SirLynix:
I don't understand that Linux error, but this isn't caused by this commit

waruqi:
> I don't understand that Linux error, but this isn't caused by this commit

>checking for c++ includes(SFML/Graphics.hpp) 

The new test examples test more interface links and therefore trigger more hidden issues.

```lua
        if package:config("graphics") then
                void test(int args, char** argv) {
                    sf::Text text;
                    text.setString("Hello world");
                }
            ]]}, {includes = "SFML/Graphics.hpp"}))
        end
```

SirLynix:
Yes, I understood that it's a hidden issue my additional tests uncovered. But I don't understand why it's triggered on Linux with GCC but not with Clang, also libXext is linked.

waruqi:
I don't know, it should work.

waruqi:
> /usr/sbin/ld: cannot find -lXrender: No such file or directory


add Xrender dep?

packages/s/sfml/xmake.lua
```diff
@@ -61,31 +61,31 @@ package("sfml")
...
         end
         if package:config("window") or package:config("graphics") then
             package:add("links", a .. "window" .. e)
-            if package:is_plat("windows", "mingw") then
+            if package:is_plat("windows", "mingw") and not package:config("shared") then
                 package:add("syslinks", "opengl32", "gdi32", "user32", "advapi32")
             end
             if package:is_plat("linux") then
-                package:add("deps", "libx11", "libxrandr", "freetype", "eudev")
+                package:add("deps", "libx11", "libxext", "libxrandr", "freetype", "eudev")
```
waruqi, line 74: add libxrender

PR #1519, opened 2022-10-10
add aws-sdk-cpp

igavintang:
* By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.

------

* 默认情况下，构建所有的SDK，如果只需要AWS S3，那么设置build_only=" s3 "，多个SDK用逗号分隔。

### DIFF
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -6,7 +6,7 @@ package("aws-sdk-cpp")
     add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
 
     if is_plat("macosx") then
-        add_frameworks("SystemConfiguration")
+        add_frameworks("Foundation", "CoreFoundation", "SystemConfiguration")
     end
 
     add_configs("build_only",  {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -6,11 +6,11 @@ package("aws-sdk-cpp")
     add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
 
     add_configs("build_only",  {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
-    add_configs("http_client", {description = 'If disabled, no platform-default http client will be included in the library.', default = false, type = "boolean"})
+    add_configs("http_client", {description = 'If disabled, no platform-default http client will be included in the library.', default = true, type = "boolean"})
     add_configs("encryption",  {description = 'If disabled, no platform-default encryption will be included in the library.', default = false, type = "boolean"})
 
     add_deps("zlib")
-    add_deps("cmake")
+    add_deps("cmake", "ninja")
 
     on_load(function (package)
         if package:config("http_client") then
@@ -31,7 +31,7 @@ package("aws-sdk-cpp")
         if package:config("build_only") then
             table.insert(configs, "-DBUILD_ONLY=" .. package:config("build_only"))
         end
-        import("package.tools.cmake").install(package, configs)
+        import("package.tools.cmake").install(package, configs, {cmake_generator = "Ninja"})
     end)
 
     on_test(function (package)
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -7,7 +7,7 @@ package("aws-sdk-cpp")
 
     add_configs("build_only", {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
     add_deps("cmake")
-    add_deps("curl")
+    add_deps("libcurl")
 
     on_install("linux", "macosx", function (package)
         local configs = {}
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -10,7 +10,7 @@ package("aws-sdk-cpp")
     add_configs("encryption",  {description = 'If disabled, no platform-default encryption will be included in the library.', default = false, type = "boolean"})
 
     add_deps("zlib")
-    add_deps("cmake", "ninja")
+    add_deps("cmake")
 
     on_load(function (package)
         if package:config("http_client") then
@@ -31,7 +31,7 @@ package("aws-sdk-cpp")
         if package:config("build_only") then
             table.insert(configs, "-DBUILD_ONLY=" .. package:config("build_only"))
         end
-        import("package.tools.cmake").install(package, configs, {cmake_generator = "Ninja"})
+        import("package.tools.cmake").install(package, configs)
     end)
 
     on_test(function (package)
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -7,6 +7,7 @@ package("aws-sdk-cpp")
 
     add_configs("build_only", {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
     add_deps("cmake")
+    add_deps("curl")
 
     on_install("linux", "macosx", function (package)
         local configs = {}
--- a/packages/a/aws-sdk-cpp/xmake.lua
+++ b/packages/a/aws-sdk-cpp/xmake.lua
@@ -5,13 +5,9 @@ package("aws-sdk-cpp")
     add_urls("https://github.com/aws/aws-sdk-cpp.git")
     add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
 
-    if is_plat("macosx") then
-        add_frameworks("Foundation", "CoreFoundation", "Security", "SystemConfiguration")
-    end
-
     add_configs("build_only",  {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
     add_configs("http_client", {description = 'If disabled, no platform-default http client will be included in the library.', default = true, type = "boolean"})
-    add_configs("encryption",  {description = 'If disabled, no platform-default encryption will be included in the library.', default = true, type = "boolean"})
+    add_configs("encryption",  {description = 'If disabled, no platform-default encryption will be included in the library.', default = false, type = "boolean"})
 
     add_deps("zlib")
     add_deps("cmake")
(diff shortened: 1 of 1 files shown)


packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -0,0 +1,16 @@
...
+
+    add_configs("build_only", {description = 'By default, all SDKS are built, if AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
+    add_deps("cmake")
+    on_install("linux", "macosx", function (package)
+        local configs = {}
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        table.insert(configs, "-DBUILD_ONLY=" .. (package:config("build_only") or ""))
+        import("package.tools.cmake").install(package, configs)
+    end)
```
waruqi, line 16: missing on_test

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -0,0 +1,30 @@
+package("aws-sdk-cpp")
+    set_homepage("https://github.com/aws/aws-sdk-cpp")
+    set_description("AWS SDK for C++")
+
+    add_urls("https://github.com/aws/aws-sdk-cpp.git")
+    add_versions("1.9.333", "6ec095fa11379174da20bf1523c93c92503eb8e2")
+
+    add_configs("build_only", {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
+    add_deps("cmake")
+
```
waruqi, line 10: add_deps("curl")

igavintang:
```shell
error: @programdir/core/main.lua:280: @programdir/modules/private/action/require/impl/package.lua:758: package(curl) not found!
```

packages/a/aws-sdk-cpp/xmake.lua
igavintang:
```shell
error: @programdir/core/main.lua:280: @programdir/modules/private/action/require/impl/package.lua:758: package(curl) not found!
```

SirLynix: it's actually called `libcurl`, `add_deps("libcurl")`

igavintang: > it's actually called `libcurl`, `add_deps("libcurl")`

Thx

waruqi:
>   Could not find zlib

maybe you need add zlib dep too.

igavintang:
> > Could not find zlib
> 
> maybe you need add zlib dep too.
Yes, added

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -0,0 +1,32 @@
...
+    add_urls("https://github.com/aws/aws-sdk-cpp.git")
+    add_versions("1.9.333", "6ec095fa11379174da20bf1523c93c92503eb8e2")
+
+    add_configs("build_only", {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
+    add_deps("zlib")
+    add_deps("libcurl")
+    add_deps("cmake")
+
+    on_install("linux", "macosx", function (package)
+        local configs = {}
```
waruqi, line 14: please disable some test options. 

https://github.com/aws/aws-sdk-cpp/blob/e9372218a2c8fab756ecaa6e4fefcdb33c3670c1/CMakeLists.txt#L42-L43

waruqi:
maybe it also need openssl

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -3,14 +3,16 @@ package("aws-sdk-cpp")
...
+    add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
 
     add_configs("build_only", {description = 'By default, all SDKS are built, if only AWS S3 is required, then set build_only="s3", with multiple SDKS separated by commas.'})
     add_deps("libcurl", "openssl", "zlib")
     add_deps("cmake")
 
     on_install("linux", "macosx", function (package)
         local configs = {}
+        table.insert(configs, "-DMINIMIZE_SIZE=ON")
+        table.insert(configs, "-DCMAKE_PREFIX_PATH=" .. package:installdir())
```
waruqi, line 15: why? xmake will pass all deps path to CMAKE_PREFIX_PATH in cmake.install

igavintang: openssl already included, but has "missing: crypto_LIBRARY" error.
Okay, Let me see again:)

waruqi: so you can patch it to disable find crypto and set packagedeps, it will force pass all flags about openssl

igavintang: > so you can patch it to disable find crypto and set packagedeps, it will force to pass all flags about openssl

Ok, I will patch Findcrypto.cmake file

igavintang:
https://github.com/aws/aws-sdk-cpp/blob/main/cmake/Findcrypto.cmake#L42-L68

```shell
if (NOT crypto_LIBRARY)
    if (BUILD_SHARED_LIBS)
        set(crypto_LIBRARY ${crypto_SHARED_LIBRARY})
    else()
        if (crypto_STATIC_LIBRARY)
           set(crypto_LIBRARY ${crypto_STATIC_LIBRARY})
        else()
           set(crypto_LIBRARY ${crypto_SHARED_LIBRARY})
        endif()
    endif()
endif()
```
Later, I will try patch code

waruqi:
ok, you can try

waruqi:
I now have ssl and curl disabled and it should work on linux, archlinux, macos but there are not enough memory resources on the ci and it will be forced to break.

But on fedora, one of its a sub library still looks for crypto and find fails.

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -0,0 +1,61 @@
...
+        io.replace("cmake/Findcrypto.cmake",
+            "if (BUILD_SHARED_LIBS)\n            set(crypto_LIBRARY ${crypto_SHARED_LIBRARY})",
+            [[
+                if (BUILD_SHARED_LIBS)
+                    if (crypto_SHARED_LIBRARY)
+                        set(crypto_LIBRARY ${crypto_SHARED_LIBRARY})
+                    else()
+                        set(crypto_LIBRARY ${crypto_STATIC_LIBRARY})
+                    endif()
+            ]], {plain = true})
```
waruqi, line 48: you can use `add_patches` to add large patch.

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -18,7 +18,7 @@ package("aws-sdk-cpp")
 
     on_load(function (package)
         if package:config("http_client") then
-            package:add("deps", "libcurl")
+            package:add("deps", "libcurl", {configs = {openssl = false}})
```
waruqi, line 21: why?

igavintang: [CMakeLists.txt](https://github.com/aws/aws-sdk-cpp/blob/ecf0320fefbd57ef1744c2b5c539e4466cd252ed/CMakeLists.txt#L195)
> libcurl and aws-sdk-cpp-core may link to different libcrypto, which leads to some issues for shared build.

I would like to rule out if this causes compilation errors😅

waruqi: By default, they should all be using xmake-repo or openssl on the system, unless it uses the system libcurl library.

Also disabling libcurl's openssl may cause http to lack ssl features.

waruqi:
The few failed ci tasks were probably just link failures due to insufficient ci memory.

SirLynix:
You have to add the missing system framework.

```lua
if is_plat("macosx") then
    add_frameworks("SystemConfiguration")
end
```

waruqi:
> You have to add the missing system framework.
> 
> ```lua
> if is_plat("macosx") then
>     add_frameworks("SystemConfiguration")
> end
> ```

it happen in cmake/link. we should pass frameworks flags to cmakelists.

try 

https://github.com/xmake-io/xmake-repo/blob/2dcf09d97482223cf5a24deef0467475c8c33124/packages/p/proj/xmake.lua#L51-L53

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -5,13 +5,9 @@ package("aws-sdk-cpp")
     add_urls("https://github.com/aws/aws-sdk-cpp.git")
     add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
 
-    if is_plat("macosx") then
-        add_frameworks("Foundation", "CoreFoundation", "Security", "SystemConfiguration")
```
waruqi: we need also them if http_client is enabled.

igavintang: 👌

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -15,6 +15,9 @@ package("aws-sdk-cpp")
     on_load(function (package)
         if package:config("http_client") then
             package:add("deps", "libcurl")
+            if is_plat("macosx") then
```
waruqi, line 18: package:is_plat

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -15,6 +15,9 @@ package("aws-sdk-cpp")
...
+                add_frameworks("Foundation", "CoreFoundation", "Security", "SystemConfiguration")
```
waruqi, line 19: package:add("frameworks", "xxxx")

waruqi: and add_frameworks

packages/a/aws-sdk-cpp/xmake.lua
igavintang: Oh, sorry😅, this is the first time to write xmake-repo, thanks for your help

packages/a/aws-sdk-cpp/xmake.lua
```diff
@@ -0,0 +1,68 @@
+package("aws-sdk-cpp")
+    set_homepage("https://github.com/aws/aws-sdk-cpp")
+    set_description("AWS SDK for C++")
+
+    add_urls("https://github.com/aws/aws-sdk-cpp.git")
+    add_versions("1.9.362", "e9372218a2c8fab756ecaa6e4fefcdb33c3670c1")
+
+    if is_plat("macosx") then
+        add_frameworks("Foundation", "CoreFoundation", "Security", "SystemConfiguration")
+    end
```
waruqi, line 10: Why add it repeatedly? Need it even if http is not enabled?

igavintang: > Why add it repeatedly? Need it even if http is not enabled?

deleted. 🧠🔫 😂

waruqi:
I have merged it, thanks!

PR #1541, opened 2022-10-16
brpc 1.3.0 add patch

igavintang:
* brpc 1.3.0 add patch


packages/b/brpc/xmake.lua
```diff
@@ -4,6 +4,7 @@ package("brpc")
 
     add_urls("https://github.com/apache/incubator-brpc.git")
     add_versions("1.3.0", "a90cf60714941632b2986826336a7c50cbd3d530")
+    add_patches("1.3.0", path.join(os.scriptdir(), "patches", "1.3.0.diff"), "a71bf46a4a6038a89da3ee9057dea5f452155a2da1f1c9bdcae7ecd0bb5e0510")
```
waruqi, line 7: 加一层版本目录 patches/1.3.0/cmake.patch

waruqi: 参考 https://github.com/xmake-io/xmake-repo/blob/master/packages/f/ftgl/patches/2.4.0/cmake.patch 路径格式

igavintang: 👌

PR #1577, opened 2022-10-26
Support to control `lto` flag in boost

garfieldkwong:
Allow to control whether turn on `lto` for the package via `add_requireconfg`

### DIFF
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -166,6 +166,10 @@ package("boost")
             "debug-symbols=" .. (package:debug() and "on" or "off"),
             "link=" .. (package:config("shared") and "shared" or "static")
         }
+
+        if package:config("lto") then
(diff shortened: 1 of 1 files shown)


packages/b/boost/xmake.lua
```diff
@@ -70,6 +70,7 @@ package("boost")
 
     add_configs("all",          { description = "Enable all library modules support.",  default = false, type = "boolean"})
     add_configs("multi",        { description = "Enable multi-thread support.",  default = true, type = "boolean"})
+    add_configs("lto",          { description = "Enable lto for compiling.", default = false, type = "boolean"})
```
waruqi, line 73: lto is builtin config, you need not add it.

garfieldkwong: I force-pushed for deleting the line `add_configs("lto"...`

PR #1623, opened 2022-11-11
add cargs

haneefmubarak:
Adds a package used for parsing flags easily in C and C++


waruqi:
> error C2220: the following warning is treated as an error 

use io.replace or add_patches to disable this flags, https://github.com/likle/cargs/blob/f422cb2ceb40aeff440a63813a7d6bc8817dea81/cmake/EnableWarnings.cmake#L6

haneefmubarak:
fixed and pushed, thanks!

PR #1627, opened 2022-11-14
Python: Dismiss local python installation if pip is not present

SirLynix:
Python can be installed without pip/setuptools/wheel that python package otherwise guarantees.

This PR requires https://github.com/xmake-io/xmake/pull/3059 to work

### DIFF
--- a/packages/p/python/xmake.lua
+++ b/packages/p/python/xmake.lua
@@ -103,7 +103,7 @@ package("python")
         end
     end)
 
-    on_install("@windows", "@msys", "@cygwin", function (package)
+    on_install("@windows|x86", "@windows|x64", "@msys", "@cygwin", function (package)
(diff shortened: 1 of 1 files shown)


waruqi:
> Hello there, `pip` is the tool used in the python ecosystem to download and install package from online repositories, the best known being https://pypi.org. The two other packages: setuptools and wheel, are two tools used to install the packages downloaded by pip. They usually come bundled with pip and are required to install built packages (wheel) or to build packages from sources (setuptools).

I don't understand what you mean? Does this have anything to do with this current patch?

SirLynix:
I don't know why it still tries to build on ARM even though I selected x86 and x64.

About Windows ARM, Python 3.11 will support it so this package can be updated then

PR #1640, opened 2022-11-19
Update libsdl

SirLynix:
and make with_x config only on Linux

### DIFF
--- a/packages/l/libsdl/xmake.lua
+++ b/packages/l/libsdl/xmake.lua
@@ -44,6 +44,7 @@ package("libsdl")
         add_syslinks("iconv")
     elseif is_plat("linux", "bsd") then
         if is_plat("bsd") then
+	    add_deps("libusb")
             add_syslinks("usbhid")
         end
         add_syslinks("pthread", "dl")
--- a/packages/l/libsdl/xmake.lua
+++ b/packages/l/libsdl/xmake.lua
@@ -44,7 +44,7 @@ package("libsdl")
         add_syslinks("iconv")
     elseif is_plat("linux", "bsd") then
         if is_plat("bsd") then
-	    add_deps("libusb")
+	        add_deps("libusb")
             add_syslinks("usbhid")
         end
         add_syslinks("pthread", "dl")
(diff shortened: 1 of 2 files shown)


waruqi:
missing libusb dep for new version. or you can disable it

waruqi:
maybe missing include libusb.h in sdl?

it's inline api.

https://github.com/libusb/libusb/blob/f3619c4078a921c3d6bf3238ff7a8dc70d60e40e/libusb/libusb.h#L1837

SirLynix:
See https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi.c#L803

I think we need libsdl build logs to understand which usb library it uses, as it builds correctly.

waruqi:
> See https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi.c#L803
> 
> I think we need libsdl build logs to understand which usb library it uses, as it builds correctly.

you can add `-vD` in ci

SirLynix:
```
2022-11-20T12:30:22.9860810Z checking for usbhid.h... yes
2022-11-20T12:30:23.0427180Z checking for libusbhid.h... no
2022-11-20T12:30:23.1097400Z checking for usbhid... yes
2022-11-20T12:30:23.2281770Z checking for libusb-1.0... no
2022-11-20T12:30:23.2882290Z checking for libusb.h... yes
```

it looks like it doesn't find libusb-1.0, which is the version installed by xmake.

packages/l/libsdl/xmake.lua
```diff
@@ -154,6 +154,16 @@ package("libsdl")
...
+        elseif package:is_plat("bsd") then
+            local dep = package:dep("libusb")
+            if dep then
+                local depfetch = dep:fetch()
+                if depfetch then
+                    for _, includedir in ipairs(depfetch.includedirs or depfetch.sysincludedirs) do
+                        table.join2(cflags, "-I" .. includedir)
+                    end
+                end
+            end
```
waruqi, line 166: try `packagedeps.= "libusb"` for autoconf.install

PR #1656, opened 2022-11-24
Improve quickcpplib

Arthapz:
Convert quickcpplib to XMake


packages/q/quickcpplib/xmake.lua
```diff
@@ -13,11 +13,12 @@ package("quickcpplib")
     end)
 
     on_test(function (package)
+        local cxxflags = is_toolchain("clang") and {"-fsized-deallocation"} or {}
```
waruqi: package:has_tool("cxx", "clang", "clangxx")

PR #1663, opened 2022-11-29
Change mentions of PucklaMotzer09 -> PucklaJ, due to change of username

PucklaJ:
I just changed my username to **PucklaJ**. And therefore I wanted to reflect that in xmake-repo. This also completely removes the mention of my name in the patch of `libsdl_gfx`


waruqi:
you need fix patch checksum

PR #1686, opened 2022-12-21
update sol2 -> v3.3.0

UrNightmaree:


packages/s/sol2/xmake.lua
```diff
@@ -7,6 +7,7 @@ package("sol2")
     set_urls("https://github.com/ThePhD/sol2/archive/$(version).tar.gz",
              "https://github.com/ThePhD/sol2.git")
 
+	add_versions("v3.3.0", "b82c5de030e18cb2bcbcefcd5f45afd526920c517a96413f0b59b4332d752a1e")
```
waruqi, line 10: incorrect indent

packages/s/sol2/xmake.lua
```diff
@@ -28,8 +29,7 @@ package("sol2")
     end)
 
     on_install(function (package)
-        local configs = {}
-        import("package.tools.cmake").install(package, configs)
+        os.cp("include/sol",package:installdir("include"))
```
waruqi, line 32: please revert it, cmake.install installs an additional *.cmake import file, which is useful for other packages that depend on sol2 and helps to find sol2 in other packages using cmake's find_package.

UrNightmaree:
that's weird, the CI only failed if its Windows Arm and iPhoneOS

waruqi:
> that's weird, the CI only failed if its Windows Arm and iPhoneOS

we need patch it to disable system call on ios, or you can disable iphoneos platform.

> /vendor/lua-5.4.4/src/loslib.c:146:10: error: 'system' is unavailable: not available on iOS
>  stat = system(cmd);

packages/s/sol2/xmake.lua
```diff
@@ -28,8 +29,7 @@ package("sol2")
     end)
 
     on_install(function (package)
-        local configs = {}
-        import("package.tools.cmake").install(package, configs)
+        import("package.tools.cmake").install(package, {})
```
waruqi, line 32: remove `, {})`

waruqi: `import("package.tools.cmake").install(package)`

packages/s/sol2/xmake.lua
```diff
@@ -29,7 +29,9 @@ package("sol2")
     end)
 
     on_install(function (package)
-        import("package.tools.cmake").install(package)
+		if not package:is_plat("iphoneos") then
+        	import("package.tools.cmake").install(package)
+		end
```
waruqi: incorrect fix, you should not do it.

packages/s/sol2/xmake.lua
UrNightmaree: changed in commit `exclude iphoneos attempt 2`

packages/s/sol2/xmake.lua
UrNightmaree: changed in commit `exclude iphoneos attempt 2`

SirLynix:
shouldn't it be add("deps", "lua >=5.4") so it supports the future Lua versions?

packages/s/sol2/xmake.lua
```diff
@@ -23,12 +24,26 @@ package("sol2")
...
+                package:add("deps", "lua")
+            end
         end
     end)
 
     on_install(function (package)
         local configs = {}
+        if package:version() and package:version():ge("3.3") then
+            table.insert(configs, "-DSOL2_BUILD_LUA=FALSE")
+            local lua = package:dep("lua"):fetch()
```
SirLynix, line 39: won't package:dep("lua") be nil if "includes_lua" is false?

UrNightmaree:
this should be able to merge

PR #1689, opened 2022-12-22
add missing stb_vorbis.c

likun123687:
add missing stb_vorbis.c

### DIFF
--- a/packages/s/stb/xmake.lua
+++ b/packages/s/stb/xmake.lua
@@ -16,6 +16,7 @@ package("stb")
 
     on_install(function (package)
         os.cp("*.h", package:installdir("include"))
+        os.cp("*.c", package:installdir("include"))
     end)
 
     on_test(function (package)


SirLynix:
stb_vorbis.c sounds like a mystery.

it says in https://github.com/nothings/stb/issues/282 that's it's not meant to be included but compiled and linked, but it's the only file which has stb_vorbis_* prototypes.

I don't know what's the best thing to do here, rename it stb_vorbis.h (as suggested in https://github.com/nothings/stb/issues/25)? Copy stb_vorbis.c in the include dirs (which is very counterintuitive)? Split it ourselves and make this a static library?

Copying stb_vorbis.c in the include dirs seems like the easiest and right thing to be done until this is fixed in the main repository, but that's really ugly.

navossoc:
@SirLynix

The best way I found to use this lib from vcpkg is:

> create 'stb_vorbis.h'
> #define STB_VORBIS_HEADER_ONLY
> #include "stb_vorbis.c"

This way I can use the latest package available on vcpkg and don't need to worry in copy, pasting, etc...
More info here: https://nothings.org/stb_vorbis/#:~:text=You%20can%20also%20avoid,%23include%20%22stb_vorbis.c%22

But yeah, it is pretty weird the behavior of this file if you look at the others files.

PR #1695, opened 2022-12-23
libsdl: Fix static compilation

SirLynix:
It was compiling shared and static at the same time causing issue when linking.


SirLynix:
It's just a matter of adding the missing syslinks and frameworks in static configuration, I will fix this in a couple of day (have to go for christmas)

PR #1698, opened 2022-12-25
Adding libcurl version 7.85, 7.86, 7.87

IanBullard:
Updating libcurl package to include versions 7.85, 7.86, 7.87.


waruqi:
you need modify libcurl/xmake.lua to trigger correct package test.

IanBullard:
@waruqi 

Apologies, this is my first difficult change to xmake-repo so I'm still trying to understand the system.  From what I see, libcurl added a call `BCryptGenRandom` to `rand.c`, and that's causing a linker error on non-windows MingW static builds.  I do not have a system configured to iterate on this problem, so I've been trying to debug via commits.  If I cannot fix it tonight, I'll install msys2 and iterate on the problem there.

If this is not correct, please let me know.  I don't really understand your comment.

packages/l/libcurl/xmake.lua
```diff
@@ -72,7 +72,7 @@ package("libcurl")
         table.insert(configs, (package:version():ge("7.80") and "-DCURL_USE_SCHANNEL=ON" or "-DCMAKE_USE_SCHANNEL=ON"))
         local version = package:version()
 
-        if is_plat("mingw") and (version:ge("7.85.0")) and not (package:config("shared")) then
+        if package:is_plat("mingw") and version:ge("7.85") then
             package:add("links", "Bcrypt")
```
waruqi, line 76: links -> syslinks

PR #1706, opened 2022-12-31
Add Blake 1.3.3

IanBullard:
Add blake 1.3.3

### DIFF
--- a/packages/b/blake3/xmake.lua
+++ b/packages/b/blake3/xmake.lua
@@ -30,6 +30,7 @@ package("blake3")
                         add_files("c/*x86-64_unix.S")
                     end
                 elseif is_arch("x86", "i386", "i686") then
+                    add_cxflags("-msse4.1")
                     add_files("c/blake3_portable.c")
                     add_files("c/blake3_sse2.c")
                     add_files("c/blake3_sse41.c")
--- a/packages/b/blake3/xmake.lua
+++ b/packages/b/blake3/xmake.lua
@@ -12,6 +12,7 @@ package("blake3")
         local configs = {}
         if (package:is_plat("msys2") and package:is_arch("i686")) then
             io.replace("blake3_avx2.c", "return _mm256_loadu_si256((const __m256i *)src);", "return _mm256_loadu_si256((const __m256i_u *)src);", {plain = true})
+            io.replace("blake3_sse41.c", "x, _mm_set_epi8(", "x, (__m128i)_mm_set_epi8(")
         end
 
         io.writefile("xmake.lua", [[
@@ -33,7 +34,7 @@ package("blake3")
                     else
                         add_files("c/*x86-64_unix.S")
                     end
-                elseif is_arch("x86", "i386", "i686") then
+                elseif is_arch("x86", "i386") then
                     add_files("c/blake3_portable.c")
                     add_files("c/blake3_sse2.c")
                     add_files("c/blake3_sse41.c")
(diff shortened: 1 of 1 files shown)


packages/b/blake3/xmake.lua
```diff
@@ -28,6 +29,12 @@ package("blake3")
                     else
                         add_files("c/*x86-64_unix.S")
                     end
+                elseif is_arch("x86") then
```
waruqi, line 32: `is_arch("x86", "i386")`

waruqi:
error: @programdir\modules\private\async\runjobs.lua:256: @programdir\modules\private\action\build\object.lua:91: @programdir\modules\core\tools\gcc.lua:709: D:/a/_temp/msys64/mingw32/lib/gcc/i686-w64-mingw32/12.2.0/include/avxintrin.h:927:1: error: inlining failed in call to 'always_inline' '_mm256_loadu_si256': target specific option mismatch
  927 | _mm256_loadu_si256 (__m256i_u const *__P)
      | ^~~~~~~~~~~~~~~~~~
c\blake3_avx2.c:8:10: note: called from here
    8 |   return _mm256_loadu_si256((const __m256i *)src);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  > in c\blake3_avx2.c

maybe it does not support mingw/i386?

you can disable it `on_install("mingw|x86_64")`

packages/b/blake3/xmake.lua
```diff
@@ -28,6 +33,12 @@ package("blake3")
                     else
                         add_files("c/*x86-64_unix.S")
                     end
+                elseif is_arch("x86", "i386", "i686") then
```
waruqi, line 36: we need not i686, it is also i386 for mingw platform.

IanBullard: Right, I wasn't sure, I'll correct this with my next commit.

IanBullard:
I have to stop for tonight, even if the checks complete, please do not merge.

IanBullard:
@waruqi How do I disable MingW (Msys2) static/shared i686?  Those are the only two failed builds and it's become obvious they're not supported.

waruqi:
> @waruqi How do I disable MingW (Msys2) static/shared i686? Those are the only two failed builds and it's become obvious they're not supported.

> you can disable it on_install("mingw|x86_64")

packages/b/blake3/xmake.lua
```diff
@@ -8,11 +8,11 @@ package("blake3")
     add_versions("1.3.3", "27d2bc4ee5945ba75434859521042c949463ee7514ff17aaef328e23ef83fec0")
     add_versions("1.3.1", "112becf0983b5c83efff07f20b458f2dbcdbd768fd46502e7ddd831b83550109")
 
-    on_install("msys2|i686")
+    on_install("mingw|x86_64")
```
waruqi, line 11: no

```
    on_install("windows", "linux", "macosx", "bsd", "mingw|x86_64", "android", "iphoneos", "cross", function (package)
```

packages/b/blake3/xmake.lua
```diff
@@ -8,11 +8,11 @@ package("blake3")
...
 
     on_install(function (package)
```
waruqi, line 13: you need modify this

packages/b/blake3/xmake.lua
IanBullard: Thank you.

PR #1723, opened 2023-01-06
port pdfhummus to windows

jingkaimori:

------


packages/p/pdfhummus/xmake.lua
```diff
@@ -36,14 +38,24 @@ package("pdfhummus")
...
                 add_files("PDFWriter/*.cpp")
                 add_headerfiles("(PDFWriter/*.h)")
                 add_packages("freetype")
                 add_packages("libtiff", "libpng", "libjpeg")
-                add_packages("libaesgm", "zlib")
+                add_packages("zlib")
+                if is_plat("windows","mingw") then
+                    add_packages("zlib")
+                    add_files("LibAesgm/*.c")
+                    add_includedirs("LibAesgm")
```
waruqi, line 55: why not improve libaesgm package? it is from `pdfhummus/LibAesgm`.

jingkaimori: libaesgm is imported by pdfhummus at 2017, but its first release is earlier than 2009. It's hard to determine newer libaesgm can be compiled on windows.

waruqi: But now the libaesgm is exactly the same as the pdfhummus/LibAesgm, in addition, even if it does not compile on windows. We should also improve the package code of libaesgm. You can make libaesgm support Windows with add_patches

jingkaimori:
why dynamic linking on windows fails?

waruqi:
> why dynamic linking on windows fails?

it did not export c++ class symbols in shared library.

you can try adding this code in ported xmake.lua

```lua
if is_plat("windows") and is_kind("shared") then
    add_rules("utils.symbols.export_all", {export_classes = true})
end
```

jingkaimori:
test script try to install libaesgm on windows|arm, even if pdfhummus is not supported on this platform. Is it a bug?

waruqi:
> test script try to install libaesgm on windows|arm, even if pdfhummus is not supported on this platform. Is it a bug?

you need not disable window arm, It is experimental. I will merge it.

PR #1729, opened 2023-01-09
Dump drogon to v1.8.2 and trantor to v1.5.8 and c-ares to 1.18.1

TES286:
Dump drogon to v1.8.2 and trantor to v1.5.8


TES286:
I've looked at the checks that fail, and some of the checks that fail on Windows are interesting: file 'C:/Users/runneradmin/AppData/Local' not found
Windows uses '\' to separate paths, but here it becomes '/', which is the usage of Linux and Mac. But in fact, this path will almost always exist, and many programming languages or tools will automatically convert '/' to '\' , but there doesn't seem to be one here, the OS doesn't recognize it.
So I don't think it's a problem with this PR
But why is it okay on Windows 2019? Metaphysics?

waruqi:
> CUSTOMBUILD : error : file 'C:/Users/runneradmin/AppData/Local' not found 

It should be trantor/CMakeLists.txt issue, maybe it has a custom build command, and it does not support path with space better.

you can check it's CMakeLists.txt and add patch to fix it.

waruqi:
or you can try use cmake to build trantor on vs2022 directly and debug it.

TES286:
Trantor can be successfully installed using vcpkg, see https://github.com/TES286-org/test-build/actions/runs/3871779193/jobs/6599887573

But the project itself does not have any patch in vcpkg, only its dependencies have newer versions.

I'll test that updating its dependencies succeeds.

packages/c/c-ares/xmake.lua
```diff
@@ -26,24 +36,5 @@ package("c-ares")
...
-            table.insert(configs, "--enable-debug")
-        else
-            table.insert(configs, "--disable-debug")
-        end
-        import("package.tools.autoconf").install(package, configs)
-    end)
-
-    on_test(function (package)
-        assert(package:has_cfuncs("ares_library_init", {includes = "ares.h"}))
-    end)
```
waruqi, line 49: why?

TES286: It's my wrong... sorry....

PR #1734, opened 2023-01-11
update xmake.lua

freestriker:
To solve the problem #1733  that vulkan-headers conflicts with the local vulkansdk, add new logic in xmake.lua to first search the local vulkansdk and use the local vulkansdk‘s header files.

### DIFF
--- a/packages/v/vulkan-headers/xmake.lua
+++ b/packages/v/vulkan-headers/xmake.lua
@@ -24,6 +24,17 @@ package("vulkan-headers")
         import("package.tools.cmake").install(package)
     end)
 
+    on_fetch(function (package, opt)
+        if opt.system then
+            import("detect.sdks.find_vulkansdk")
+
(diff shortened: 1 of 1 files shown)


packages/v/vulkan-headers/xmake.lua
```diff
@@ -24,6 +24,17 @@ package("vulkan-headers")
         import("package.tools.cmake").install(package)
     end)
 
+    on_fetch(function (package, opt)
+        if opt.system then
+            import("detect.sdks.find_vulkansdk")
+
+            local vulkansdk = find_vulkansdk()
+            if vulkansdk then
+                return {includedirs = vulkansdk.includedirs, linkdirs = vulkansdk.linkdirs, links = {}}
```
waruqi, line 33: 这个包只是获取头文件，不要加 linkdirs /links 。。要用 links ，直接走 vulkansdk 包。

PR #1737, opened 2023-01-12
Add configurations to Raylib

Akselmo:
USE_EXTERNAL_GLFW, PLATFORM and OPENGL_VERSION

### DIFF
--- a/packages/r/raylib/xmake.lua
+++ b/packages/r/raylib/xmake.lua
@@ -42,6 +42,10 @@ package("raylib")
     end
     add_deps("opengl", {optional = true})
 
+    add_configs("use_external_glfw", {description = "Link raylib against system GLFW instead of embedded one.", default = "OFF", values = {"OFF", "ON", "IF_POSSIBLE"}})
+    add_configs("platform", {description = "Platform to build for.", default = "Desktop", values = {"Desktop", "Web", "Android", "Raspberry Pi"}})
+    add_configs("opengl_version", {description = "Force a specific OpenGL Version.", default = "OFF", values = {"OFF", "3.3", "2.1", "1.1", "ES 2.0"}})
+
     on_install("macosx|x86_64", function (package)
         os.cp("include/*.h", package:installdir("include"))
         os.cp("lib/libraylib.a", package:installdir("lib"))
@@ -51,6 +55,10 @@ package("raylib")
         local configs = {"-DBUILD_EXAMPLES=OFF"}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        table.insert(configs, "-DCUSTOMIZE_BUILD=ON") -- Needed for the below flags
+        table.insert(configs, "USE_EXTERNAL_GLFW=" .. (package:config("use_external_glfw")))
+        table.insert(configs, "PLATFORM=" .. (package:config("platform")))
+        table.insert(configs, "OPENGL_VERSION=" .. (package:config("opengl_version")))
         import("package.tools.cmake").install(package, configs, {packagedeps = {"libx11", "libxrender", "libxrandr", "libxinerama", "libxcursor", "libxi", "libxfixes", "libxext"}})
     end)
 
--- a/packages/r/raylib/xmake.lua
+++ b/packages/r/raylib/xmake.lua
@@ -40,54 +40,36 @@ package("raylib")
         add_syslinks("pthread", "dl", "m")
         add_deps("libx11", "libxrandr", "libxrender", "libxinerama", "libxcursor", "libxi", "libxfixes", "libxext")
     end
+
     add_deps("opengl", {optional = true})
 
-    add_configs("external_glfw", {description = "Link raylib against system GLFW instead of embedded one.", default = false, type = "boolean"})
-    add_configs("opengl_version", {description = "Force a specific OpenGL Version.", values = {"3.3", "2.1", "1.1", "ES 2.0"}})
-
-    on_load(function (package)
-        if package:config("external_glfw") then
-            package:add("deps", "glfw")
-        end
-    end)
+    local configs = {"-DBUILD_EXAMPLES=OFF"}
 
     on_install("macosx|x86_64", function (package)
         os.cp("include/*.h", package:installdir("include"))
         os.cp("lib/libraylib.a", package:installdir("lib"))
+        table.insert(configs, "-DPLATFORM=Desktop")
     end)
 
     on_install("windows", "linux", "macosx|arm64", "mingw", function (package)
-        local configs = {"-DBUILD_EXAMPLES=OFF"}
+        table.insert(configs, "-DPLATFORM=Desktop")
+    end)
+
+    on_install("android", function (package)
+        table.insert(configs, "-DPLATFORM=Android")
+    end)
+
+    on_install("web", function (package)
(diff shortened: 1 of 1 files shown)


packages/r/raylib/xmake.lua
```diff
@@ -42,6 +42,10 @@ package("raylib")
     end
     add_deps("opengl", {optional = true})
 
+    add_configs("use_external_glfw", {description = "Link raylib against system GLFW instead of embedded one.", default = "OFF", values = {"OFF", "ON", "IF_POSSIBLE"}})
```
waruqi, line 45: please do not use OFF/ON, we should use `true/false`, and remove `use_` prefix

```lua
    add_configs("external_glfw", {description = "Link raylib against system GLFW instead of embedded one.", default = false, type = "boolean"})
```

and it is enabled, you need add `add_deps("glfw")`

packages/r/raylib/xmake.lua
```diff
@@ -42,6 +42,10 @@ package("raylib")
...
+    add_configs("platform", {description = "Platform to build for.", default = "Desktop", values = {"Desktop", "Web", "Android", "Raspberry Pi"}})
```
waruqi, line 46: please remove it, we should use `package:is_plat()` to get platform from `xmake f -p/--plat`

packages/r/raylib/xmake.lua
```diff
@@ -42,6 +42,10 @@ package("raylib")
...
+    add_configs("opengl_version", {description = "Force a specific OpenGL Version.", default = "OFF", values = {"OFF", "3.3", "2.1", "1.1", "ES 2.0"}})
```
waruqi, line 47: please do not OFF, 

```
    add_configs("opengl_version", {description = "Force a specific OpenGL Version.", values = {"3.3", "2.1", "1.1", "ES 2.0"}})
```

packages/r/raylib/xmake.lua
```diff
@@ -51,6 +55,10 @@ package("raylib")
         local configs = {"-DBUILD_EXAMPLES=OFF"}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        table.insert(configs, "-DCUSTOMIZE_BUILD=ON") -- Needed for the below flags
+        table.insert(configs, "USE_EXTERNAL_GLFW=" .. (package:config("use_external_glfw")))
```
waruqi, line 59:
```
table.insert(configs, "USE_EXTERNAL_GLFW=" .. (package:config("external_glfw") and "ON" or "OFF"))
```

and you need `package:add("deps", "glfw")` in on_load when it is enabled

packages/r/raylib/xmake.lua
```diff
@@ -51,6 +55,10 @@ package("raylib")
...
+        table.insert(configs, "PLATFORM=" .. (package:config("platform")))
```
waruqi, line 60:
```
local platform
if package:is_plat("windows", "macosx", "linux") then
    platform = "Desktop"
elseif package:is_plat("android", "iphoneos") then
   ...
end
if platform then
    table.insert(configs, "PLATFORM=" .. platform)
end
```

packages/r/raylib/xmake.lua
```diff
@@ -51,6 +55,10 @@ package("raylib")
...
+        table.insert(configs, "OPENGL_VERSION=" .. (package:config("opengl_version")))
```
waruqi, line 61: same

packages/r/raylib/xmake.lua
Akselmo: Ah got it. I was following how it was mentioned in raylib wiki. Will fix these.

Akselmo: I have now changed the file according to feedback. Did I understand correctly?

Also does the `    on_install("windows", "linux", "macosx|arm64", "mingw", function (package)` need to have also `android` `web` and `raspberry-pi` ?

Akselmo: Seems that if I set `-DCUSTOMIZE_BUILD=ON` then none of my builds actually run. I think I'm missing something.

I will look into this more, will keep you updated. 

Of course if you or anyone else who sees this has something proper working already I'd appreciate any help.

Akselmo: Okay, I had approached this in completely wrong way. I removed the custom opengl and glfw setting.

Instead, I'm checking now the architecture and platform, then setting `-DPLATFORM` correctly depending on it.

Please check and tell me if there's anything I could've done better :)

Akselmo: Well, it works on my linux pc but not on my raspberry pi.

I don't know what to do :(

packages/r/raylib/xmake.lua
```diff
@@ -55,10 +60,27 @@ package("raylib")
...
+        table.insert(configs, "-DCUSTOMIZE_BUILD=ON")
+
+        table.insert(configs, "USE_EXTERNAL_GLFW=" .. (package:config("external_glfw") and "ON" or "OFF"))
+
+        local platform
+        if package:is_plat("windows", "macosx", "linux") then
+            platform = "Desktop"
+        elseif package:is_plat("android") then
+            platform = "Android"
+        elseif package:is_plat("web") then
```
waruqi: xmake has not `web`, it is only for raylib

is_plat("wasm")

packages/r/raylib/xmake.lua
```diff
@@ -55,10 +60,27 @@ package("raylib")
...
+            platform = "Web"
+        elseif package:is_plat("raspberry-pi") then
```
waruqi: has not this, remove it

packages/r/raylib/xmake.lua
```diff
@@ -40,15 +40,34 @@ package("raylib")
...
-        local configs = {"-DBUILD_EXAMPLES=OFF"}
+        table.insert(configs, "-DPLATFORM=Desktop")
+    end)
+
+    on_install("android", function (package)
+        table.insert(configs, "-DPLATFORM=Android")
+    end)
+
+    on_install("web", function (package)
+        table.insert(configs, "-DPLATFORM=Web")
```
waruqi, line 63: oh, no... please revert it

packages/r/raylib/xmake.lua
waruqi: > I have now changed the file according to feedback. Did I understand correctly?
> 
> Also does the ` on_install("windows", "linux", "macosx|arm64", "mingw", function (package)` need to have also `android` `web` and `raspberry-pi` ?

they are only raylib, not for xmake. you can see platform list of xmake. `xmake f -h`

waruqi: > Well, it works on my linux pc but not on my raspberry pi.
> 
> I don't know what to do :(

raspberry pi is also linux platform in xmake. You will need to fix it based on the actual compilation error.

waruqi:
> [Set correct platform for raylib](https://github.com/xmake-io/xmake-repo/pull/1737/commits/7087045b6f2e071f8f6ffd7883fc60bd243eda2b)

Completely wrong patch, please revert it

packages/r/raylib/xmake.lua
waruqi: or you should open an issue in raylib to fix your errors on raspberry pi. or use `add_patches` to fix it.

Akselmo: I'll remake whole PR at some point when i have more time

Thank you for being patient with me

PR #1754, opened 2023-01-25
add custom main option for Catch2

dingjr27:
Add custom main option for Catch2. 
If set main == "catch2", it will link Catch2Main, which is the default option for catch2. 
If set main == "custom", it will link your custom main.

### DIFF
--- a/packages/c/catch2/xmake.lua
+++ b/packages/c/catch2/xmake.lua
@@ -19,7 +19,7 @@ package("catch2")
     add_versions("v2.9.2", "dc486300de22b0d36ddba1705abb07b9e5780639d824ba172ddf7062b2a1bf8f")
 
     add_configs("cxx17", {description = "Compiles Catch as a C++17 library (requires a C++17 compiler).", default = true, type = "boolean"})
-    add_configs("catch2main",{description = "Use Catch2 entry point", default = true, type = "boolean"})
+    add_configs("catch2main", {description = "Use Catch2 entry point", default = true, type = "boolean"})
 
     if is_plat("mingw") and is_subhost("msys") then
         add_extsources("pacman::catch")
--- a/packages/c/catch2/xmake.lua
+++ b/packages/c/catch2/xmake.lua
@@ -30,25 +30,25 @@ package("catch2")
     end
 
     on_load(function (package)
+        package:add("deps", "cmake")
         if package:config("catch2main") then
             package:add("components", "main")
         end
         package:add("components", "lib")
+        if package:version():lt("3.0") then
+            package:set("kind", "library", {headeronly = true})
+        end
     end)
 
     on_component("main", function (package, component)
         if package:version():ge("3.0") then
-            component:add("deps", "cmake")
             component:add("links", "Catch2Main")
         end
     end)
 
     on_component("lib", function (package, component)
         if package:version():ge("3.0") then
-            component:add("deps", "cmake")
             component:add("links", "Catch2")
-        else
-            package:set("kind", "library", {headeronly = true})
         end
     end)
 
--- a/packages/c/catch2/xmake.lua
+++ b/packages/c/catch2/xmake.lua
@@ -30,26 +30,23 @@ package("catch2")
     end
 
     on_load(function (package)
-        package:add("deps", "cmake")
-        if package:config("catch2main") then
-            package:add("components", "main")
-        end
-        package:add("components", "lib")
-        if package:version():lt("3.0") then
+        if package:version():ge("3.0") then
+            package:add("deps", "cmake")
+            if package:config("catch2main") then
+                package:add("components", "main")
+            end
+            package:add("components", "lib")
+        else
             package:set("kind", "library", {headeronly = true})
         end
     end)
 
     on_component("main", function (package, component)
-        if package:version():ge("3.0") then
-            component:add("links", "Catch2Main")
-        end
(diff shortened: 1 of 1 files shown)


SirLynix:
This should be a package component instead of an option I think

dingjr27:
For version less than 3.0, i am not sure whether to use
`component:set("kind", "library", {headeronly = true})`
or
`package:set("kind", "library", {headeronly = true})`
All these two pass test.

packages/c/catch2/xmake.lua
```diff
@@ -29,11 +30,23 @@ package("catch2")
...
+
+    on_component("main", function (package, component)
+        if package:version():ge("3.0") then
+            component:add("links", "Catch2Main")
+        end
+    end)
+
+    on_component("lib", function (package, component)
         if package:version():ge("3.0") then
-            package:add("deps", "cmake")
```
waruqi, line 33: why.remove it?

packages/c/catch2/xmake.lua
```diff
@@ -29,11 +30,23 @@ package("catch2")
...
-            package:add("links", "Catch2Main", "Catch2")
+            component:add("links", "Catch2")
         else
-            package:set("kind", "library", {headeronly = true})
+            component:set("kind", "library", {headeronly = true})
```
waruqi, line 49: set_kind only for package, not component:set

packages/c/catch2/xmake.lua
```diff
@@ -19,6 +19,7 @@ package("catch2")
     add_versions("v2.9.2", "dc486300de22b0d36ddba1705abb07b9e5780639d824ba172ddf7062b2a1bf8f")
 
     add_configs("cxx17", {description = "Compiles Catch as a C++17 library (requires a C++17 compiler).", default = true, type = "boolean"})
+    add_configs("use_catch2main",{description = "Use Catch2 entry point", default = true, type = "boolean"})
```
waruqi, line 22: please rename to `catch2main`

packages/c/catch2/xmake.lua
waruqi: you can add package deps in on_load

packages/c/catch2/xmake.lua
SirLynix: This should be removed and replaced by components

packages/c/catch2/xmake.lua
dingjr27: Sorry, that is fixed in the latest commit

packages/c/catch2/xmake.lua
dingjr27: fixed

packages/c/catch2/xmake.lua
```diff
@@ -30,14 +30,23 @@ package("catch2")
     end
 
     on_load(function (package)
+       if package:config("use_catch2main") then
+            package:add("components", "main")
```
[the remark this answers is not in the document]
dingjr27: The main component is always added since "catch2main" is default to be true.

packages/c/catch2/xmake.lua
```diff
@@ -30,23 +30,25 @@ package("catch2")
...
-       if package:config("use_catch2main") then
+        if package:config("catch2main") then
             package:add("components", "main")
         end
         package:add("components", "lib")
     end)
 
     on_component("main", function (package, component)
         if package:version():ge("3.0") then
+            component:add("deps", "cmake")
```
waruqi, line 41: please move to on_load, and use `package:add("deps")`

packages/c/catch2/xmake.lua
```diff
@@ -30,23 +30,25 @@ package("catch2")
...
             component:add("links", "Catch2Main")
         end
     end)
 
     on_component("lib", function (package, component)
         if package:version():ge("3.0") then
+            component:add("deps", "cmake")
```
waruqi, line 48: and here

packages/c/catch2/xmake.lua
```diff
@@ -30,23 +30,25 @@ package("catch2")
...
             component:add("links", "Catch2")
         else
-            component:set("kind", "library", {headeronly = true})
+            package:set("kind", "library", {headeronly = true})
```
waruqi, line 51: please move to on_load

packages/c/catch2/xmake.lua
SirLynix: Yeah, but libsdl has a config for compatibility purposes because it existed before components, if you introduce this you don't have to add a config

packages/c/catch2/xmake.lua
```diff
@@ -30,25 +30,25 @@ package("catch2")
     end
 
     on_load(function (package)
+        package:add("deps", "cmake")
```
waruqi, line 33: only `catch3 >= 3.0` need it.

packages/c/catch2/xmake.lua
waruqi: The use of config is indeed optional, and we can disable catch2main in the following way 

```lua
add_requires("catch2")
target("test")
    -- ..
    add_packages("catch2", {components = "lib"})
```

waruqi:
if you do not want use catch2main, you can try:

```lua
add_requires("catch2")
target("test")
    -- ..
    add_packages("catch2", {components = "lib"})
```

PR #1770, opened 2023-01-31
Add msdfgen

SirLynix:
https://github.com/Chlumsky/msdfgen


SirLynix:
It looks like a limitation of MinGW/msys2 when exporting all symbols:
```
unrecognized file format in 'CMakeFiles/msdfgen-core.dir/core/Contour.cpp.obj, 332'
```

I can make the shared option read-only on MinGW/msys2, as the only way to fix this I'm aware of is to patch the library to make use of declspec dllexport/dllimport instead of relying on cmake export all symbols feature.

waruqi:
> It looks like a limitation of MinGW/msys2 when exporting all symbols:
> 
> ```
> unrecognized file format in 'CMakeFiles/msdfgen-core.dir/core/Contour.cpp.obj, 332'
> ```
> 
> I can make the shared option read-only on MinGW/msys2, as the only way to fix this I'm aware of is to patch the library to make use of declspec dllexport/dllimport instead of relying on cmake export all symbols feature.

you can disable shared first.

PR #1771, opened 2023-01-31
update lzma and xz

xq114:


packages/l/lzma/xmake.lua
```diff
@@ -3,19 +3,20 @@ package("lzma")
...
         io.writefile("xmake.lua", [[
             add_rules("mode.debug", "mode.release")
             target("lzma")
                 set_kind("$(kind)")
-                add_files("Alloc.c", "LzFind.c",  "Lzma2Dec.c", "Lzma2Enc.c", "LzmaDec.c", "LzmaEnc.c", "LzmaLib.c")
+                add_files("Alloc.c", "LzFind.c",  "Lzma2Dec.c", "Lzma2Enc.c", "LzmaDec.c", "LzmaEnc.c", "LzmaLib.c", "CpuArch.c")
                 add_headerfiles("7zTypes.h", "LzFind.h", "LzHash.h", "Lzma2Dec.h", "Lzma2Enc.h", "LzmaDec.h", "LzmaEnc.h", "LzmaLib.h")
                 if is_plat("windows") then
-                    add_files("LzFindMt.c", "MtCoder.c", "MtDec.c", "Threads.c", "DllSecur.c", "Lzma2DecMt.c")
+                    add_files("LzFindMt.c", "LzFindOpt.c", "MtCoder.c", "MtDec.c", "Threads.c", "DllSecur.c", "Lzma2DecMt.c")
```
SirLynix, line 19: Won't that break with version 19.00?

xq114: I just forgot to add this file when I wrote the package description for version 19.00. It contains an implementation of a decompress function, but it seems nobody is using this package or the missing function would be already reported. Anyway, since liblzma from `xz` provides more functionality than this package `lzma sdk`, directly switching to xz could be a better idea.

PR #1784, opened 2023-02-04
update libxml2

xq114:


xq114:
这里下载失败了，需要重跑下

darcy-shen:
```
note: the following packages are unsupported on macosx/x86_64:
  -> libiconv 1.17
```
Is it caused by this commit?
https://github.com/XmacsLabs/mogan/actions/runs/4096346496/jobs/7063919459

packages/l/libiconv/xmake.lua
```diff
@@ -42,7 +46,7 @@ package("libiconv")
         })
     end)
 
-    on_install("macosx", "linux", "cross", "android", function (package)
```
darcy-shen, line 45: why `macosx` is removed here?

PR #1827, opened 2023-02-14
add ut

star-hengxing:
UT: C++20 μ(micro)/Unit Testing Framework.

I don't how to deal with header only modules package.


waruqi:
> UT: C++20 μ(micro)/Unit Testing Framework.
> 
> I don't how to deal with header only modules package.

还得再来个 patch 改下，这个包名来通用了，容易冲突，改成 boost_ut

PR #1832, opened 2023-02-15
Update libdeflate to v1.17

jj683:


jj683:
Windows 2019 + ARM fails like a previous PR with libdeflate: #1674

PR #1837, opened 2023-02-15
wayland: ensure ninja >= 1.8.2 is present

SirLynix:
Should fix this:
```
  => download https://wayland.freedesktop.org/releases/wayland-1.19.0.tar.xz .. ok
downloading resource(protocols: https://wayland.freedesktop.org/releases/wayland-protocols-1.21.tar.xz) to wayland-1.19.0 ..
/usr/bin/curl -SL -A "Xmake/2.7.6+HEAD.7fd2cec12 (Linux;5.15.0-1031-azure) curl/7.81.0" https://wayland.freedesktop.org/releases/wayland-protocols-1.21.tar.xz -o /home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/resources/protocols/wayland-protocols-1.21.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  114k  100  114k    0     0   236k      0 --:--:-- --:--:-- --:--:--  236k
/usr/bin/tar -xf /home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/resources/protocols/wayland-protocols-1.21.tar.xz -C /home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/resources/protocols/wayland-protocols-1.21.tar.xz.dir.tmp
meson -Ddocumentation=false -Dc_link_args=-lm --libdir=lib --prefix=/home/runner/work/NazaraEngine/xmake-global/.xmake/packages/w/wayland/1.19.0/d4813a414d234036bb8606e1e07c959d --libdir=lib -Dbuildtype=release -Db_staticpic=true build_d4813a41
The Meson build system
Version: 1.0.0
Source dir: /home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/source
Build dir: /home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/source/build_d4813a41
Build type: native build
Project name: wayland
Project version: 1.19.0
C compiler for the host machine: cc (gcc 11.3.0 "cc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0")
C linker for the host machine: cc ld.bfd 2.38
C++ compiler for the host machine: c++ (gcc 11.3.0 "c++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0")
C++ linker for the host machine: c++ ld.bfd 2.38
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments -Wno-unused-parameter: YES 
Compiler for C supports arguments -Wstrict-prototypes: YES 
Compiler for C supports arguments -Wmissing-prototypes: YES 
Compiler for C supports arguments -fvisibility=hidden: YES 
Has header "sys/prctl.h" : YES 
Checking for function "accept4" : YES 
Checking for function "mkostemp" : YES 
Checking for function "posix_fallocate" : YES 
Checking for function "prctl" : YES 
Checking for function "memfd_create" : YES 
Checking for function "strndup" : YES 
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Run-time dependency libffi found: YES 3.4.2
Header "sys/signalfd.h" has symbol "SFD_CLOEXEC" : YES 
Header "sys/timerfd.h" has symbol "TFD_CLOEXEC" : YES 
Header "time.h" has symbol "CLOCK_MONOTONIC" : YES 
Checking for function "clock_gettime" : YES 
Run-time dependency expat found: YES 2.4.7
Run-time dependency libxml-2.0 found: YES 2.9.13
Configuring config.h using configuration
Configuring wayland-version.h using configuration
Library m found: YES
Run-time dependency threads found: YES
Program nm found: YES (/usr/bin/nm)
Program wayland-egl-symbols-check found: YES (/home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/source/egl/wayland-egl-symbols-check)
Library dl found: YES
Dependency threads found: YES unknown (cached)
Dependency threads found: YES unknown (cached)
Library dl found: YES
Program sed found: YES (/usr/bin/sed)
Program scanner-test.sh found: YES (/home/runner/work/NazaraEngine/xmake-global/.xmake/cache/packages/2302/w/wayland/1.19.0/source/tests/scanner-test.sh)
Build targets in project: 40
NOTICE: Future-deprecated features used:
 * 0.55.0: {'ExternalProgram.path'}
 * 0.62.0: {'pkgconfig.generate variable for builtin directories'}
 * 0.64.0: {'copy arg in configure_file'}

wayland 1.19.0

  User defined options
    buildtype    : release
    libdir       : lib
    prefix       : /home/runner/work/NazaraEngine/xmake-global/.xmake/packages/w/wayland/1.19.0/d4813a414d234036bb8606e1e07c959d
    b_staticpic  : true
    c_link_args  : -lm
    documentation: false


ERROR: Could not detect Ninja v1.8.2 or newer
```

### DIFF
--- a/packages/w/wayland/xmake.lua
+++ b/packages/w/wayland/xmake.lua
@@ -14,7 +14,7 @@ package("wayland")
         add_extsources("apt::libwayland-dev", "pacman::wayland")
     end
 
-    add_deps("meson", "libxml2", "libffi", "expat", "bison", "pkg-config")
+    add_deps("meson", "ninja >=1.8.2", "libxml2", "libffi", "expat", "bison", "pkg-config")
 
     on_install("linux", function (package)
         -- imports


eli-schwartz:
This seems fundamentally wrong. Ninja 1.8.2 is required by meson itself, not the wayland package.

waruqi:
> This seems fundamentally wrong. Ninja 1.8.2 is required by meson itself, not the wayland package.

ninja has been removed in meson, https://github.com/xmake-io/xmake-repo/commit/ec493cd3b36bfc562865584df988c69e5d7dfade#diff-5e32e9463bd58ad9a172a24d5d3c5a1f1fd229abe0f7ed853f52ee49dc33ffa5

@xq114

eli-schwartz:
While it is technically true that Meson can use xcode on macOS or msbuild on Windows, that's not the default and you absolutely do need ninja if you're going to use the ninja backend for Meson. How you get ninja is up to you... adding the dependencies on individual meson-using packages seems quite strange.

waruqi:
> While it is technically true that Meson can use xcode on macOS or msbuild on Windows, that's not the default and you absolutely do need ninja if you're going to use the ninja backend for Meson. How you get ninja is up to you... adding the dependencies on individual meson-using packages seems quite strange.
If ninja is built into meson, the user cannot choose another generator (msbuild/make) and ninja will always be installed, even if it is not used.

Although perhaps meson uses ninja by default, we can control it more flexibly inside the package itself with add_deps("meson", "ninja"). That's how most packages do it now.

https://github.com/xmake-io/xmake-repo/blob/45ce8c5c7f52180318261b6a312d2c747c2260b0/packages/l/libpsl/xmake.lua#L10

eli-schwartz:
A very interesting suggestion, but... xmake itself, in package.tools.meson, explicitly imports package.tools.ninja and then directly runs it.

waruqi:
> A very interesting suggestion, but... xmake itself, in package.tools.meson, explicitly imports package.tools.ninja and then directly runs it.

This is only because I haven't had time to improve it yet, and I haven't encountered a need to switch to another generator package yet. But there is no guarantee that other packages do not have this need.

In the future I will improve xmake so that tools.meson supports to select more generators, but this doesn't conflict with the current problem, we just need to keep the `add_deps("meson", "ninja")` setting, it doesn't affect anything.

But if the built-in ninja is built into meson, if I later make tools.meson support generator switching, then I'll need to bulk modify all the existing package configurations, which is a problem instead.

eli-schwartz:
Is it really the responsibility of individual recipes to choose the generator to use (and the dependencies to install in order to use that generator)?

waruqi:
> Is it really the responsibility of individual recipes to choose the generator to use (and the dependencies to install in order to use that generator)?

Yes, for example with cmake, we usually just need the default generator (which is not ninja), but for some larger packages we need to switch to ninja to speed up the compilation.

Therefore, we need to choose the best generator for each package.
https://github.com/xmake-io/xmake-repo/blob/45ce8c5c7f52180318261b6a312d2c747c2260b0/packages/l/libtorch/xmake.lua#L172

PR #1850, opened 2023-02-20
[spdlog] Fix package issues

Qudix:
- Fixed typo: `SPDLOG_COMPILE_LIB` -> `SPDLOG_COMPILED_LIB`.
- Fixed `fmt_external` option
- Added `fmt_external_ho` option
- Added `SPDLOG_NO_EXCEPTIONS` define for `noexcept` option.
- Added `wchar` option.

### DIFF
--- a/packages/s/spdlog/xmake.lua
+++ b/packages/s/spdlog/xmake.lua
@@ -30,9 +30,7 @@ package("spdlog")
     on_load(function (package)
         if not package:config("header_only") then
             package:add("defines", "SPDLOG_COMPILED_LIB")
-            if package:version():ge("1.4.0") then
-                package:add("deps", "cmake")
-            end
+            package:add("deps", "cmake")
         end
         if package:config("fmt_external") then
             package:add("defines", "SPDLOG_FMT_EXTERNAL")
@@ -47,7 +45,7 @@ package("spdlog")
     end)
 
     on_install(function (package)
-        if package:version():lt("1.4.0") or package:config("header_only") then
+        if (not package:gitref() and package:version():lt("1.4.0")) or package:config("header_only") then
             os.cp("include", package:installdir())
             return
         end
--- a/packages/s/spdlog/xmake.lua
+++ b/packages/s/spdlog/xmake.lua
@@ -18,9 +18,10 @@ package("spdlog")
     add_versions("v1.4.2", "56b90f0bd5b126cf1b623eeb19bf4369516fa68f036bbc22d9729d2da511fb5a")
     add_versions("v1.3.1", "db6986d0141546d4fba5220944cc1f251bd8afdfc434bda173b4b0b6406e3cd0")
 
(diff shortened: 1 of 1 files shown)


packages/s/spdlog/xmake.lua
```diff
@@ -24,25 +24,30 @@ package("spdlog")
...
         end
-        if package:config("fmt_external") then
-            package:add("deps", "fmt", {configs = {header_only = true}})
+        if package:config("wchar") then
+            package:add("defines", "SPDLOG_WCHAR_TO_UTF8_SUPPORT")
         end
     end)
 
     on_install(function (package)
-        if (package:version() and package:version():lt("1.4.0")) or package:config("header_only") then
```
waruqi, line 45: package:version() will be nil if it is from git branch, e.g. `add_requires("spdlog master")`

Qudix: Interesting, I was following what other packages do. Does that mean that all packages that can accept a branch name must check whether version is nil?

waruqi: yes, if it provides git url.

and we can also use `package:gitref()` to check it's from git branch.

Qudix: If I understand this correctly, does this mean that even if you specify a commit or tag or branch that also satisfies the condition on the right, being earlier than or equal to 2.5.9, it will fail the `if check`?

waruqi: ? if you use `add_requires("xxx branch")` to set branch, xmake cannot get package version. the branch name may be any string.

xmake just can select versions from `add_versions()`

waruqi:
Is it ready to be merged? I see it is still in draft status.

Qudix:
I still haven't had a response to my two TODO questions above.

waruqi:
> Determine if the option can be set to default to false. This seems unlikely at this point.header_only

We can't modify it, it will break the package the user is using now. Also, xmake-repo has its own independent configuration specification and option status, and is not guaranteed to be consistent with cmake's original options.

> Determine if the option can have fmt's be moved to a separate option to follow spdlog's cmake.fmt_externalheader_only = truefmt_external_ho

I don't understand what this is to solve

Qudix:
> I don't understand what this is to solve
At the moment it is adding `fmt` as a dependency with `header_only` turned on but it is passing the non-header-only cmake option `SPDLOG_FMT_EXTERNAL` to ON rather than `SPDLOG_FMT_EXTERNAL_HO`

waruqi:
> > I don't understand what this is to solve
> 
> At the moment it is adding `fmt` as a dependency with `header_only` turned on but it is passing the non-header-only cmake option `SPDLOG_FMT_EXTERNAL` to ON rather than `SPDLOG_FMT_EXTERNAL_HO`

You can fix it and add fmt_external_ho.

PR #1854, opened 2023-02-22
[commonlibsse-ng] Edit `spdlog` dependency

Qudix:
Set `spdlog`'s `header_only` option to false

### DIFF
--- a/packages/c/commonlibsse-ng/port/xmake.lua
+++ b/packages/c/commonlibsse-ng/port/xmake.lua
@@ -42,7 +42,8 @@ option("skse_xbyak")
 option_end()
 
 -- add packages
-add_requires("fmt", "spdlog", "rapidcsv")
+add_requires("fmt", "rapidcsv")
+add_requires("spdlog", { configs = { header_only = false } })
 
 if has_config("skse_xbyak") then
     add_requires("xbyak")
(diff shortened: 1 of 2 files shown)


waruqi:
why?

Qudix:
Header only brings in `Windows.h` and we had some issues caused by this requiring us to create workarounds. With this we can remove them.

waruqi:
you can use `add_requireconfs("commonlibsse-ng.spdlog", { configs = { header_only = false } })` to override it in user side.

and if Windows.h will bring some issues, we can add patches to fix it.

waruqi:
Switching the default dependency configuration doesn't solve the problem at all, it just hides it. And there is a risk of breaking existing users' projects.

Qudix:
I can count the users of this package on one hand as they all use my project and there should be no risk of breakage. Also commonlibsse-ng already made similar changes for the vcpkg port.

waruqi:
> I can count the users of this package on one hand as they all use my project and there should be no risk of breakage. Also commonlibsse-ng already made similar changes for the vcpkg port.

? I'm not sure how you counted, commonlibsse-ng is your project? so you can count it?

Qudix:
No, people using commonlibsse-ng directly use vcpkg, I made the xmake port for use in my template project
https://github.com/Qudix/template-commonlibsse-ng

waruqi:
> No, people using commonlibsse-ng directly use vcpkg, I made the xmake port for use in my template project https://github.com/Qudix/template-commonlibsse-ng

so you can add `add_requireconfs("commonlibsse-ng.spdlog", { configs = { header_only = false } })` in your xmake.lua

Also, it is possible that other users are using this package and I am not sure if it will have an effect.

Qudix:
I don't want to do that, this should be a simple change with no unwanted effects. I know exactly who else is using this and they all go through me.

waruqi:
ok, but please fix some errors on ci.

PR #1878, opened 2023-03-01
Add GDExtension library

PucklaJ:
This is the Godot 4 version of `godotcpp`, but since the API changed drastically I created a new package for it. I don't use `import("package.tools.scons")` directly, because this version of godotcpp uses different arguments. The different arguments such as `plat`, `arch`, `use_mingw` and so on are not universal to scons. They are unique to godot. Therefore the build function for scons needs to be reworked accordingly.
In the mean time I am just using `import("lib.detect.find_tool")` directly and then generate the arguments manually.

### DIFF
--- a/packages/g/gdextension/xmake.lua
+++ b/packages/g/gdextension/xmake.lua
@@ -9,7 +9,10 @@ package("gdextension")
   add_deps("scons")
   add_includedirs("gen/include", "include")
  
-  on_install("linux", "windows", "macosx", "mingw", "iphoneos", "android", function(package)
+  on_install("linux", "windows", "macosx", "mingw", "iphoneos", "android", "wasm", function(package)
+    import("core.base.option")
+    import("lib.detect.find_tool")
+  
     local platform = package:plat()
     if package:is_plat("mingw") then
         platform = "windows"
@@ -22,17 +25,24 @@ package("gdextension")
     local arch = package:arch()
     if package:is_arch("x64") then
         arch = "x86_64"
+    elseif package:is_arch("x86") then
+        arch = "x86_32"
+    elseif package:is_arch("arm64-v8a") then
+        arch = "arm64"
+    elseif package:is_arch("arm") then
+        arch = "arm32"
+    elseif package:is_arch("ppc") then
+        arch = "ppc32"
     end
     
     local configs = {
-      "-j " .. tostring(os.default_njob()),
+      "-j " .. (option.get("jobs") or tostring(os.default_njob())),
       "use_mingw=" .. (package:is_plat("mingw") and "yes" or "no"),
       "target=" .. (package:debug() and "template_debug" or "template_release"),
       "platform=" .. platform,
       "arch=" .. arch,
     }
     
-    import("lib.detect.find_tool")
     local scons = assert(find_tool("scons"), "scons not found")
     
     os.execv(scons.program, configs)
--- a/packages/g/gdextension/xmake.lua
+++ b/packages/g/gdextension/xmake.lua
@@ -1,140 +1,147 @@
 package("gdextension")
 
-  set_homepage("https://godotengine.org/")
-  set_description("C++ bindings for the Godot 4 script API")
-  
-  set_urls("https://github.com/godotengine/godot-cpp.git")
-  add_versions("4.0", "9d1c396c54fc3bdfcc7da4f3abcb52b14f6cce8f")
-  
-  add_deps("scons")
-  add_includedirs("gen/include", "include")
+    set_homepage("https://godotengine.org/")
+    set_description("C++ bindings for the Godot 4 script API")
+    
+    set_urls("https://github.com/godotengine/godot-cpp.git")
+    add_versions("4.0", "9d1c396c54fc3bdfcc7da4f3abcb52b14f6cce8f")
+    
+    add_deps("scons")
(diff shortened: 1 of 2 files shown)


packages/g/gdextension/xmake.lua
```diff
@@ -0,0 +1,140 @@
...
+  set_description("C++ bindings for the Godot 4 script API")
+  
+  set_urls("https://github.com/godotengine/godot-cpp.git")
+  add_versions("4.0", "9d1c396c54fc3bdfcc7da4f3abcb52b14f6cce8f")
+  
+  add_deps("scons")
+  add_includedirs("gen/include", "include")
+
+  on_load(function(package)
+    assert(not package:is_arch(
```
waruqi, line 13: 4 spaces indent

waruqi:
> They are unique to godot. Therefore the build function for scons needs to be reworked accordingly.
In the mean time I am just using import("lib.detect.find_tool") directly and then generate the arguments manually.

What are the differences? Please use `package.tools.scons` and if it doesn't support some settings yet, you can mention pr to improve it.

package.tools.scons should be able to handle the common settings for all scons.

PucklaJ:
scons is highly dynamic. Some of the differences between the current settings and godotcpp4 are:
+ `bits` does not exist, it sets them in `arch` directly.
+ `platform=osx` should be `platform=macos`
+ `freebsd` is not another platform, it's combined with `linux`
+ `target` should be set to `template_debug` and `template_release` instead of just `debug` and `release`.

Therefore the platform, architecture and mode can not be set automatically by `package.tools.scons`. They need to be set manually in each package that uses scons (which are currently only the godot related ones).

> package.tools.scons should be able to handle the common settings for all scons.

Yes I agree that it should, but because of the dynamic nature of scons this is hardly possible.

waruqi:
> platform=osx should be platform=macos

 Does scons not provide standard platform values? Or does all configuration rely on the values set by the current project?

PucklaJ:
That's the issue. It does not provide standard platform values. They are unique to every scons project.
From `SConstruct` of `godot-cpp`:
```py
platforms = ("linux", "macos", "windows", "android", "ios", "javascript")
opts.Add(
    EnumVariable(
        "platform",
        "Target platform",
        default_platform,
        allowed_values=platforms,
        ignorecase=2,
    )
)

# CPU architecture options.
architecture_array = ["", "universal", "x86_32", "x86_64", "arm32", "arm64", "rv64", "ppc32", "ppc64", "wasm32"]
architecture_aliases = {
    "x64": "x86_64",
    "amd64": "x86_64",
    "armv7": "arm32",
    "armv8": "arm64",
    "arm64v8": "arm64",
    "aarch64": "arm64",
    "rv": "rv64",
    "riscv": "rv64",
    "riscv64": "rv64",
    "ppcle": "ppc32",
    "ppc": "ppc32",
    "ppc64le": "ppc64",
}
opts.Add(EnumVariable("arch", "CPU architecture", "", architecture_array, architecture_aliases))
```

[Here is a post on StackOverflow](https://stackoverflow.com/questions/60164709/how-to-select-build-architecture-with-scons) that says that scons does not support cross compiling by default.

waruqi:
I have improved tools.scons, you can use it. And add any configs to override the builtin configs.

Later we can also improve the original godotcpp package and gradually remove some of the built-in configs that are no longer used.

https://github.com/xmake-io/xmake/commit/d469035fc49b6565a2ba7366d148d7f239bc352e

packages/g/godotcpp4/xmake.lua
```diff
@@ -0,0 +1,145 @@
...
+    end)
+ 
+    on_install("linux", "windows|x64", "windows|x86", "macosx", "mingw", "iphoneos", "android", function(package)
+        if package:is_plat("windows") then
+            io.replace("tools/targets.py", "/MD", "/" .. package:config("vs_runtime"), {plain = true})
+        end
+    
+        local platform = package:plat()
+        if package:is_plat("mingw") then
+                platform = "windows"
```
waruqi, line 38: 4 spaces indent

packages/g/godotcpp4/xmake.lua
```diff
@@ -0,0 +1,145 @@
...
+
+        void uninitialize_example_module(ModuleInitializationLevel p_level) {
+          if (p_level != MODULE_INITIALIZATION_LEVEL_SCENE) {
+            return;
+          }
+        }
+
+        extern "C" {
+        // Initialization.
+        GDExtensionBool GDE_EXPORT
```
waruqi, line 129: too long test,

waruqi:
> g++ -o gen/src/classes/stream_peer_gzip.windows.template_release.x86_64.o -c -std=c++17 -O3 -Igdextension -Iinclude -Igen/include gen/src/classes/stream_peer_gzip.cpp

mingw gcc does not seem to be used

PucklaJ:
This is probably because in MSYS2 the normal gcc is the mingw gcc.

waruqi:
> This is probably because in MSYS2 the normal gcc is the mingw gcc.

if it's cmake/autoconf/meson, xmake will pass mingw toolchains to them, but scons does not now.

waruqi:
If it doesn't support it, we can only disable the mingw platform for now

waruqi:
I disable mingw first, you can improve it later.

PR #1884, opened 2023-03-02
Fix ImGui v1.89.3-docking version

WSSDude:
Commit sha previously pointed to master branch tag 1.89.3 instead of equivalent commit in docking branch.


Qudix:
My mistake, this is correct

PR #1897, opened 2023-03-07
add package rocksdb (#1894)

byronhe:


packages/r/rocksdb/xmake.lua
```diff
@@ -0,0 +1,50 @@
...
+            table.insert(configs, "DEBUG_LEVEL=2")
+        end
+
+        if package:config("shared") then
+            table.insert(configs, "shared_lib")
+        else
+            table.insert(configs, "static_lib")
+        end
+        table.insert(configs, "install")
+        import("package.tools.make").build(package, configs)
```
waruqi, line 33: it has cmakelists.txt, you should use `tools.cmake`.  tools.cmake does not support cross-compilation and switch toolchains.

packages/r/rocksdb/xmake.lua
```diff
@@ -0,0 +1,50 @@
...
+    add_versions("v7.10.2", "4619ae7308cd3d11cdd36f0bfad3fb03a1ad399ca333f192b77b6b95b08e2f78")
+
+    add_deps("snappy")
+    add_deps("lz4")
+    add_deps("zlib")
+    add_deps("zstd")
+    add_deps("jemalloc")
+
+    if is_plat("linux") then
+        add_syslinks("pthread")
```
waruqi, line 16: add_syslinks("pthread", "rt", "dl")

packages/r/rocksdb/xmake.lua
byronhe: rocksdb/CMakeLists.txt is only for Windows platform.

i tried use cmake on linux , it fails.

waruqi: if you use makefile, you need pass mingw toolchains, and dependent packages configuraion to makefile. maybe you need add much patches.

or you can use xmake.lua to port it.

waruqi:
https://github.com/xmake-io/xmake-repo/pull/1904

PR #1900, opened 2023-03-08
RmlUi: add SVG and Lottie plugins

zziger:


zziger:
Force pushed to fix indentation issue without creating an additional commit just for that

PR #1901, opened 2023-03-08
Add cppast, debug_assert and type_safe (WIP)

SirLynix:
I'm trying to add [cppast](https://github.com/foonathan/cppast) which is based on LLVM, I couldn't make the llvm package work on Windows (as a library) so I'm working on Linux, however I'm not sure how to link properly LLVM (I only need clang-cpp, clang and clangAST libraries for this example)


SirLynix:
I think this PR is working, failed checks seems to be because of connection issues

Issue #1910, opened 2023-03-12
xrepo 安装grpc 失败


wzy2687:
### Xmake 版本

v2.7.7+20230310,

### 操作系统版本和架构

Linux ubuntu 4.15.0-200-generic #211-Ubuntu SMP Thu Nov 24 18:16:04 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### 描述问题

xrepo install -vD grpc   执行失败

在:~/.xmake/cache/packages/2303/g/grpc/v1.46.3/installdir.failed/include$  目录下没有absl 相关的第三方库.

部分出错信息如下: 

-- Installing: /home/wzy/.xmake/packages/g/grpc/v1.46.3/fa5cb5edd791499cb17b023fd18a6912/share/pkgconfig/zlib.pc
finding grpc from xmake ..
checking for xmake::grpc ... grpc v1.46.3
{ 
  static = true,
  version = "v1.46.3",
  license = "Apache-2.0",
  sysincludedirs = { 
    "/home/wzy/.xmake/packages/g/grpc/v1.46.3/fa5cb5edd791499cb17b023fd18a6912/include" 
  },
  shared = true 
}

checking for gcc ... /usr/bin/gcc
checking for the c++ compiler (cxx) ... gcc
checking for /usr/bin/gcc ... ok
checking for flags (-fPIC) ... ok
> gcc "-fPIC"
> /usr/bin/gcc -c -m64 -std=c++11 -isystem /home/wzy/.xmake/packages/g/grpc/v1.46.3/fa5cb5edd791499cb17b023fd18a6912/include -o /tmp/.xmake1000/230312/_10EB397BA5584A10851603DB27D63920.o /tmp/.xmake1000/230312/_C09B0F55C5CD47AC8879AFB72378A4B6.cpp
checking for flags (-fdiagnostics-color=always) ... ok
> gcc "-fdiagnostics-color=always"
> checking for c++ includes(grpcpp/grpcpp.h)
> checking for c++ snippet(test)
checkinfo: ...amdir/core/sandbox/modules/import/core/tool/compiler.lua:84: @programdir/modules/core/tools/gcc.lua:713: In file included from /home/wzy/.xmake/packages/g/grpc/v1.46.3/fa5cb5edd791499cb17b023fd18a6912/include/grpcpp/impl/codegen/completion_queue.h:45:0,
[...]

PR #1912, opened 2023-03-13
improve grpc

waruqi:
https://github.com/xmake-io/xmake-repo/issues/1910


waruqi:
it need higher version gcc and win sdk.

PR #1930, opened 2023-03-19
Improve JoltPhysics

SirLynix:
fix debug_renderer and profile option and enable Windows ARM support


SirLynix:
CI failing on VS 2019 ARM64 is expected, JoltPhysics requires at least VS 2022 for ARM64

PR #1951, opened 2023-03-25
Update version 4.5.6 of pdfhummus

jingkaimori:


packages/p/pdfhummus/xmake.lua
```diff
@@ -4,8 +4,10 @@ package("pdfhummus")
     set_license("Apache-2.0")
 
     add_urls("https://github.com/galkahana/PDF-Writer/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/galkahana/PDF-Writer/archive/refs/tags/v$(version).tar.gz",
```
waruqi, line 7: you can move this url to first .

PR #1957, opened 2023-03-27
add shaderwriter version 2.6

WubiCookie:


waruqi:
linux ci fails

DragonJoker:
Yes, looks like debug only fail, I'm having a look at it right now.

PR #1968, opened 2023-04-01
Add dotnet hosting for native code

yamashi:


packages/d/dotnet/xmake.lua
```diff
@@ -0,0 +1,59 @@
...
+        if package:is_plat("windows") then
+            out_path = path.join(out_path, "Microsoft.NETCore.App.Host.win-" .. (package:is_arch("x64") and "x64" or "x86"), "7.0.4", "runtimes", "win-" .. (package:is_arch("x64") and "x64" or "x86"), "native")
+            
+            os.cp(path.join(out_path, "nethost.dll"), package:installdir("bin"))
+            os.cp(path.join(out_path, "nethost.lib"), package:installdir("lib"))
+        elseif package:is_plat("linux") then
+            out_path = path.join(out_path, "Microsoft.NETCore.App.Host.linux-" .. (package:is_arch("arm64") and "arm64" or "x64"), "7.0.4", "runtimes", "linux-" .. (package:is_arch("arm64") and "arm64" or "x64"), "native")
+
+            os.cp(path.join(out_path, "*.so"), package:installdir("bin"))
+            os.cp(path.join(out_path, "*.a"), package:installdir("lib"))
```
waruqi, line 44: We should distinguish between static/shared, otherwise the linker may link to the wrong library.

and so and dylib should be in lib directory instead of bin.

```
            os.cp(path.join(out_path, "*.so"), package:installdir("lib"))
else
            os.cp(path.join(out_path, "*.a"), package:installdir("lib"))
end

File: packages/d/dotnet/xmake.lua [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+        elseif package:is_plat("macosx") then
+            out_path = path.join(out_path, "Microsoft.NETCore.App.Host.osx-" .. (package:is_arch("arm64") and "arm64" or "x64"), "7.0.4", "runtimes", "osx-" .. (package:is_arch("arm64") and "arm64" or "x64"), "native")
+
+            os.cp(path.join(out_path, "*.dylib"), package:installdir("bin"))
+            os.cp(path.join(out_path, "*.a"), package:installdir("lib"))
```
waruqi, line 49: and here

packages/d/dotnet/xmake.lua
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    on_install("windows|x86", "windows|x64", "linux|x64", "linux|arm64", "macosx|x86_64", "macosx|arm64", function (package)
+
+        local out_path = "packs"
+        if package:is_plat("windows") then
+            out_path = path.join(out_path, "Microsoft.NETCore.App.Host.win-" .. (package:is_arch("x64") and "x64" or "x86"), "7.0.4", "runtimes", "win-" .. (package:is_arch("x64") and "x64" or "x86"), "native")
+            
+            os.cp(path.join(out_path, "nethost.dll"), package:installdir("bin"))
+            os.cp(path.join(out_path, "nethost.lib"), package:installdir("lib"))
```
waruqi, line 39: and here

packages/d/dotnet/xmake.lua
yamashi: There is no static version, the .a is used to reference the .so

waruqi: Only the windows dll library may rely on .lib as an import library.

The dynamic libraries of linux and macos do not depend on the .a static library at all. There is no relationship between them.

I looked at libnethost.so and libnethost.a for linux and they both contain separate symbolic definitions with no cross-reference.

```
ruki@88b79cd6fd38:/mnt/dotnet-sdk-7.0.202-linux-x64/packs/Microsoft.NETCore.App.Host.linux-x64/7.0.4/runtimes/linux-x64/native$ nm libnethost.a | grep get_hostfxr_path
0000000000000000 T get_hostfxr_path
ruki@88b79cd6fd38:/mnt/dotnet-sdk-7.0.202-linux-x64/packs/Microsoft.NETCore.App.Host.linux-x64/7.0.4/runtimes/linux-x64/native$ nm libnethost.so | grep get_hostfxr_path
00000000000140c0 T get_hostfxr_path
```

So they should be completely separate dynamic and static libraries, and only one of them can be linked and used at the same time. Otherwise there may be link symbol conflicts, as well as runtime loading conflicts.

PR #1972, opened 2023-04-06
[convertutf] Add package

Qudix:
Add [`convertutf`](https://github.com/xmake-mirror/convertutf)

Pre-requisite for `simpleini`

### DIFF
--- a/packages/c/convertutf/xmake.lua
+++ b/packages/c/convertutf/xmake.lua
@@ -13,7 +13,7 @@ package("convertutf")
                 set_languages("c99")
                 add_files("ConvertUTF.c")
                 add_headerfiles("ConvertUTF.h")
-                if is_kind("shared") then
+                if is_plat("windows") and is_kind("shared") then
                     add_rules("utils.symbols.export_all")
                 end
         ]])
--- a/packages/c/convertutf/xmake.lua
+++ b/packages/c/convertutf/xmake.lua
@@ -13,8 +13,13 @@ package("convertutf")
                 set_languages("c99")
                 add_files("ConvertUTF.c")
                 add_headerfiles("ConvertUTF.h")
+                if is_kind("shared") then
+                    add_rules("utils.symbols.export_all")
+                end
         ]])
-        import("package.tools.xmake").install(package)
+        local configs = {}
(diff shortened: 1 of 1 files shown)


packages/c/convertutf/xmake.lua
```diff
@@ -0,0 +1,22 @@
...
+    add_versions("v1.0.1", "ecd79378cc9f24034a2edebf5f198e84c9eb8046261f8208908d2b29d4fa4d99")
+
+    on_install(function (package)
+        io.writefile("xmake.lua", [[
+            target("convertutf")
+                set_kind("$(kind)")
+                set_languages("c99")
+                add_files("ConvertUTF.c")
+                add_headerfiles("ConvertUTF.h")
+        ]])
```
waruqi, line 16: please add shared/static

https://github.com/xmake-io/xmake-repo/blob/47a1bd8a326ba23202adfb78797f2d3b115665f6/packages/z/zlib/xmake.lua#L66

Qudix: I don't believe it exports anything so it may be best to make it static only

waruqi: We can force to export it. see 

https://github.com/xmake-io/xmake-repo/blob/47a1bd8a326ba23202adfb78797f2d3b115665f6/packages/l/libimagequant/xmake.lua#L21-L23

Qudix: Oh, interesting. I'll give that a shot.

packages/c/convertutf/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+    add_versions("v1.0.1", "ecd79378cc9f24034a2edebf5f198e84c9eb8046261f8208908d2b29d4fa4d99")
+
+    on_install(function (package)
+        io.writefile("xmake.lua", [[
+            target("convertutf")
+                set_kind("$(kind)")
+                set_languages("c99")
+                add_files("ConvertUTF.c")
+                add_headerfiles("ConvertUTF.h")
+                if is_kind("shared") then
```
waruqi, line 16: add `is_plat("windows")`

we need only export it for windows.

PR #1975, opened 2023-04-06
[simpleini] Fix `define` to `defines`

Qudix:
Fix a mistake


Qudix:
Why does FreeBSD (static) take so long?

Issue #2013, opened 2023-04-21
添加 yaml_cpp_struct 包


fantasy-peak:
### Is your feature request related to a problem? Please describe.

不是。创建此 issue 以提交 PR

### Describe the solution you'd like

添加 yaml-cpp_struct 包的 xmake.lua

PR #2014, opened 2023-04-21
add yaml_cpp_struct

fantasy-peak:
https://github.com/xmake-io/xmake-repo/issues/2013

### DIFF
--- a/packages/v/visit_struct/xmake.lua
+++ b/packages/v/visit_struct/xmake.lua
@@ -1,0 +1,33 @@
+package("visit_struct")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://github.com/cbeck88/visit_struct")
+    set_description("A miniature library for struct-field reflection in C++")
+    set_license("BSL-1.0")
+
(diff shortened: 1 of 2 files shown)


packages/v/visit_struct/xmake.lua
```diff
@@ -0,0 +1,32 @@
+package("visit_struct")
+    set_homepage("https://github.com/cbeck88/visit_struct")
```
waruqi, line 2: set_kind("library", {headeronly = true})

packages/y/yaml_cpp_struct/xmake.lua
```diff
@@ -0,0 +1,26 @@
+package("yaml_cpp_struct")
```
waruqi, line 1: and here

fantasy-peak: done

packages/v/visit_struct/xmake.lua
fantasy-peak: done

PR #2039, opened 2023-04-26
add nanodbc && odbc

fantasy-peak:


waruqi:
> undefined lt_dlsym

missing libtldl dep? you can try add it first.

PR #2059, opened 2023-05-06
add cpp-peglib

star-hengxing:
A single file C++ header-only PEG (Parsing Expression Grammars) library.

### DIFF
--- a/packages/c/cpp-peglib/xmake.lua
+++ b/packages/c/cpp-peglib/xmake.lua
@@ -11,7 +11,9 @@ package("cpp-peglib")
     add_versions("1.8.3", "3de8aeb44a262f9c2478e2a7e7bc2bb9426a2bdd176cf0654ff5a3d291c77b73")
 
     on_install(function (package)
-        package:add("cxxflags", "/Zc:__cplusplus", {tools = "cl"})
+        if is_plat("windows") then
+            package:add("cxxflags", "/Zc:__cplusplus")
+        end
         os.cp("peglib.h", package:installdir("include"))
     end)
 
--- a/packages/c/cpp-peglib/xmake.lua
+++ b/packages/c/cpp-peglib/xmake.lua
@@ -11,6 +11,9 @@ package("cpp-peglib")
     add_versions("1.8.3", "3de8aeb44a262f9c2478e2a7e7bc2bb9426a2bdd176cf0654ff5a3d291c77b73")
 
     on_install(function (package)
+        if is_plat("windows") then
+            package:add("cxxflags", "/Zc:__cplusplus")
+        end
         os.cp("peglib.h", package:installdir("include"))
     end)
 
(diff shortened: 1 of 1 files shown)


packages/c/cpp-peglib/xmake.lua
```diff
@@ -0,0 +1,37 @@
...
+    set_description("A single file C++ header-only PEG (Parsing Expression Grammars) library")
+    set_license("MIT")
+
+    set_urls("https://github.com/yhirose/cpp-peglib/archive/refs/tags/v$(version).tar.gz",
+             "https://github.com/yhirose/cpp-peglib.git")
+
+    add_versions("1.8.3", "3de8aeb44a262f9c2478e2a7e7bc2bb9426a2bdd176cf0654ff5a3d291c77b73")
+
+    on_install(function (package)
+        package:add("cxxflags", "/Zc:__cplusplus", {tools = "cl"})
```
waruqi, line 14: 这个 flags 必须的？ package 里不支持 tools 区分，可以走 is_plat windows 处理。

star-hengxing:
```c++
// https://github.com/yhirose/cpp-peglib/blob/d9dfc5b4f5c674f16f5d26e75674055b071ce8b0/peglib.h#L40-L42
#if !defined(__cplusplus) || __cplusplus < 201703L
#error "Requires complete C++17 support"
#endif
```

vs2022 不加就报错

waruqi: __cplusplus 这玩意不是 c++ 内置的么。。应该用户自己项目里面设置  `set_languages("c++17")`

star-hengxing: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170
> The __cplusplus preprocessor macro is commonly used to report support for a particular version of the C++ standard. Because a lot of existing code appears to depend on the value of this macro matching 199711L, the compiler doesn't change the value of the macro unless you explicitly opt in by using the /Zc:__cplusplus compiler option.

可以看到 ci 都挂了。。

waruqi: 好吧。

packages/c/cpp-peglib/xmake.lua
```diff
@@ -0,0 +1,39 @@
...
+    set_description("A single file C++ header-only PEG (Parsing Expression Grammars) library")
+    set_license("MIT")
+
+    set_urls("https://github.com/yhirose/cpp-peglib/archive/refs/tags/v$(version).tar.gz",
+             "https://github.com/yhirose/cpp-peglib.git")
+
+    add_versions("1.8.3", "3de8aeb44a262f9c2478e2a7e7bc2bb9426a2bdd176cf0654ff5a3d291c77b73")
+
+    on_install(function (package)
+        if is_plat("windows") then
```
waruqi, line 14: package:is_plat

PR #2061, opened 2023-05-06
add ada

star-hengxing:
WHATWG-compliant and fast URL parser written in modern C++.


star-hengxing:
ci 问题，bsd 在上次已经过了

packages/a/ada/xmake.lua
```diff
@@ -0,0 +1,34 @@
+package("ada")
+
+    set_homepage("https://ada-url.github.io/ada")
```
anonrig, line 3: This needs to be updated with www.ada-url.com

waruqi: you can open a pr to update it.

PR #2072, opened 2023-05-10
sfml now uses git for all platforms

paul-reilly:
Just a draft to see if this passes CI after building locally. 

- don't know if I need to add downloading an archive back
- the tests need sorting out too because of changes and deprecations in the API, need to find some common ground to test.

~~Oh, drafts don't trigger CI.~~ <- just a delay in the start


SirLynix:
Superseded by #2206

PR #2078, opened 2023-05-13
Update icu4c, fix links on Windows

AidanSun05:
- Adds version 73.1.
- Fixes links on Windows (they match those on mingw but not macOS/Linux).


AidanSun05:
I noticed the static mingw builds were failing. This could be because the mingw patch is only applied to v72.1 and not the new v73.1.

waruqi:
> I noticed the static mingw builds were failing. This could be because the mingw patch is only applied to v72.1 and not the new v73.1.

You can attempt to fix it.

AidanSun05:
I added the patch to the new version, and all mingw builds now work as expected.

PR #2087, opened 2023-05-16
[commonlibsse-ng] Add `v3.7.0`

Qudix:
Add `v3.7.0` to `commonlibsse-ng`

### DIFF
--- a/packages/c/commonlibsse-ng/port/xmake.lua
+++ b/packages/c/commonlibsse-ng/port/xmake.lua
@@ -85,42 +85,38 @@ target("CommonLibSSE")
 
     -- add flags
     add_cxxflags("/permissive-")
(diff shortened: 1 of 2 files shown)


Qudix:
Ugh, fmt made a breaking change

Qudix:
I'm unsure if forcing versions is the best solution

waruqi:
> I'm unsure if forcing versions is the best solution

It is not recommended to write a fixed version.

Qudix:
Not sure how else to fix it then, spdlog won't compile with fmt version 10.0.0 due to this: https://github.com/gabime/spdlog/issues/2693

waruqi:
You can wait for spdlog to be fixed and use the next version. Or you can use add_patches to make a temporary fix to the current version to make it compatible.

PR #2089, opened 2023-05-17
[spdlog] Add `fmt10` patch

Qudix:
Add `fmt10` patch


Qudix:
Not sure why the patch checksum fails on every platform except windows?

waruqi:
> Not sure why the patch checksum fails on every platform except windows?

maybe git core.autocrlf issue on windows. you need config to disable autocrlf

PR #2100, opened 2023-05-20
add yy-thunks

star-hengxing:
Fix DecodePointer, EncodePointer,RegDeleteKeyEx etc. APIs not found in Windows XP RTM.

usage:

```lua
add_requires("yy-thunks")
add_packages("yy-thunks")

add_rules("@yy-thunks/xp")
-- add_rules("@yy-thunks/2k")
-- add_rules("@yy-thunks/vista")
```

### DIFF
--- a/packages/y/yy-thunks/rules/link.lua
+++ b/packages/y/yy-thunks/rules/link.lua
@@ -1,30 +1,32 @@
 rule("xp")
     on_config(function (target)
-        import("core.project.project")
-
         local objectfile = "YY_Thunks_for_WinXP.obj"
-        local installdir = project.required_package("yy-thunks"):installdir()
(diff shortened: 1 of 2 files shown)


packages/y/yy-thunks/rules/link.lua
```diff
@@ -0,0 +1,30 @@
+rule("xp")
+    on_config(function (target)
+        import("core.project.project")
+
+        local objectfile = "YY_Thunks_for_WinXP.obj"
+        local installdir = project.required_package("yy-thunks"):installdir()
```
waruqi, line 6:
```lua
local thunks = target:pkg("yy-thunks")
if thunks then
    local installdir = thunks:installdir()
end
```

PR #2105, opened 2023-05-22
improve yy-thunks

star-hengxing:
- Fix bug
- To run the Win2K program, we need to modify the minimum executable operating system version.

### DIFF
--- a/packages/y/yy-thunks/rules/link.lua
+++ b/packages/y/yy-thunks/rules/link.lua
@@ -24,9 +24,27 @@ rule("2k")
             raise("Win2K only supports x86 architecture")
         end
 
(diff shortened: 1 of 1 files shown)


packages/y/yy-thunks/rules/link.lua
```diff
@@ -24,9 +24,27 @@ rule("2k")
...
     end)
+
+    after_link(function (target, opt)
+        import("core.project.depend")
+        import("lib.detect.find_tool")
+        import("core.tool.toolchain")
+        import("utils.progress")
+
+        depend.on_changed(function()
+            local msvc = toolchain.load("msvc", {plat = target:plat(), arch = target:arch()})
```
waruqi, line 42: target:toolchain("msvc")

PR #2127, opened 2023-05-29
update miniaudio

jube:
miniaudio now uses git tags for its versions. This patch adds the currently available versions of miniaudio and removes the old fake version.

### DIFF
--- a/packages/m/miniaudio/xmake.lua
+++ b/packages/m/miniaudio/xmake.lua
@@ -3,7 +3,8 @@ package("miniaudio")
     set_homepage("https://miniaud.io")
     set_description("Single file audio playback and capture library written in C.")
 
-    add_urls("https://github.com/mackron/miniaudio.git")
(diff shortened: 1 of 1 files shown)


waruqi:
if it provide tags, you need also add tar url.

https://github.com/mackron/miniaudio/archive/refs/tags/0.11.17.tar.gz

packages/m/miniaudio/xmake.lua
```diff
@@ -3,7 +3,8 @@ package("miniaudio")
     set_homepage("https://miniaud.io")
     set_description("Single file audio playback and capture library written in C.")
 
-    add_urls("https://github.com/mackron/miniaudio.git")
+    set_urls("https://github.com/mackron/miniaudio/archive/refs/tags/${version}.tar.gz",
+             "https://github.com/mackron/miniaudio.git")
     add_versions("0.11.15", "26b0a9ffc0b2e9d0528106f2f5e4dfc90b0a0d6b")
```
waruqi, line 8: please use sha256 for tar.gz

PR #2146, opened 2023-06-06
Add swisseph v2.10.03

0x2Adr1:


0x2Adr1:
@waruqi for this PR and the one for CSPICE, is it possible to have an include prefix?

So that users of the library import headers like so:

```
#include <swisseph/swisseph.h>
```

Instead of:

```
#include <swisseph.h>
```

PR #2182, opened 2023-06-18
Auto-update libsdl to 2.26.5

SirLynix:


SirLynix:
cross seems broken, maybe because of recent cmake tool changes? (as it worked before and I don't think this could be caused by this new version)

waruqi:
I have fixed it.

PR #2194, opened 2023-06-20
update trantor patch for win(off_t)

sinnren:
in lib trantor, on windows off_t is defined with long, not longlong, it will be error when off_t is more than 4GB .

patch it.


waruqi:
you should push to dev branch to trigger ci test.

https://github.com/xmake-io/xmake-repo/pull/2195

PR #2205, opened 2023-06-21
Auto-update libsdl to 2.28.0

SirLynix:


SirLynix:
same issue with wasm

PR #2212, opened 2023-06-22
Fix JoltPhysics hashes

SirLynix:
Looks like GitHub reverted their changes to .patch, which broke jolt package once again.

If that happens again I will make the patches part of this repository


SirLynix:
FreeBSD network error, otherwise it's fine

PR #2213, opened 2023-06-22
Auto-update lz4 to v1.9.4

SirLynix:


SirLynix:
It created a branch from the wrong one

PR #2218, opened 2023-06-22
Auto-update open3d to v0.17.0

SirLynix:
New version of open3d detected (package version: v0.15.1, last github version: v0.17.0)


SirLynix:
@xq114 pinging you since you made that package, any idea why on Linux it exports libraries to lib/Release instead of lib? Also it has vs_runtime issues on Windows. (maybe GCC is supported now?)

Sorry I don't know open3d, I just noticed it had a new version.

SirLynix:
Superseded by #2358

PR #2222, opened 2023-06-23
Auto-update libpng to v1.6.40

SirLynix:
New version of libpng detected (package version: v1.6.37, last github version: v1.7.0beta89)


SirLynix:
I added a check against tags containing "beta"

SirLynix:
Another freebsd network error

PR #2224, opened 2023-06-24
Fix Catch2 link error when debug is enabled

calebkiage:
When building in debug mode, the linker fails to find Catch2 symbols because the link name is incorrect.


waruqi:
please push to dev branch to trigger ci test next.

PR #2229, opened 2023-06-25
[rsm-autotoml] Add package

Qudix:
Add `rsm-autotoml` header only package


Qudix:
2 CI failures appear unrelated

PR #2232, opened 2023-06-25
Add Qt webview

maximegmd:


SirLynix:
I think webview is not part of Qt base package, you have to install it using aqt

maximegmd:
The doc said to link webkit :'( get me out

SirLynix:
Are you reading Qt 5.15 doc? https://doc.qt.io/qt-5/qtwebview-module.html

PR #2233, opened 2023-06-25
Fix link error when building boost with a different toolset

calebkiage:
When building boost statically without specifying an msvc version in the config file, b2 will call the environment setup for msvc and activate the latest toolset. This works for the default case, but when you have multiple toolsets installed and aren't using the latest one, the libraries will fail because link.exe will not match the user's environment.

An example error when compiling program_options with msvc 14.1 from Visual Studio 2022 build tools:

```
error LNK2019: unresolved external symbol "__std_count_trivial_1" referenced in function...
```

### DIFF
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -220,8 +220,8 @@ package("boost")
             "threading=" .. (package:config("multi") and "multi" or "single"),
             "debug-symbols=" .. (package:debug() and "on" or "off"),
             "link=" .. (package:config("shared") and "shared" or "static"),
-            "variant="..(package:is_debug() and "debug" or "release"),
-            "runtime-debugging="..(package:is_debug() and "on" or "off")
+            "variant=" .. (package:is_debug() and "debug" or "release"),
+            "runtime-debugging=" .. (package:is_debug() and "on" or "off")
         }
 
         if package:config("lto") then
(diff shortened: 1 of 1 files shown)


packages/b/boost/xmake.lua
```diff
@@ -147,7 +154,16 @@ package("boost")
                 end
                 file:print("using darwin : : %s ;", cc)
             elseif package:is_plat("windows") then
-                file:print("using msvc : : \"%s\" ;", (package:build_getenv("cxx"):gsub("\\", "\\\\")))
+                local vs_toolset = msvc:config("vs_toolset")
+                local msvc_ver = ""
+                if vs_toolset then
+                    local i1, i2 = vs_toolset:find("%.")
+                    msvc_ver = string.sub(vs_toolset, 1, i2 + 1)
```
waruqi, line 161: i1 is not used. and please use `vs_toolset:sub()`

packages/b/boost/xmake.lua
```diff
@@ -147,7 +154,16 @@ package("boost")
                 end
                 file:print("using darwin : : %s ;", cc)
             elseif package:is_plat("windows") then
-                file:print("using msvc : : \"%s\" ;", (package:build_getenv("cxx"):gsub("\\", "\\\\")))
+                local vs_toolset = msvc:config("vs_toolset")
+                local msvc_ver = ""
+                if vs_toolset then
+                    local i1, i2 = vs_toolset:find("%.")
```
waruqi, line 160: you can use `vs_toolset:split("%.")[1]`

calebkiage: I've updated this to:

```lua
local i = vs_toolset:find("%.")
msvc_ver = i and vs_toolset:sub(1, i + 1)
```

PR #2247, opened 2023-06-27
cinatra: add package

star-hengxing:


star-hengxing:
The ci compiler does not fully support c++20.

PR #2248, opened 2023-06-27
boost: add boost package clang-cl toolchain support

calebkiage:
Added support for the clang-cl toolchain in boost.

### DIFF
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -140,11 +140,8 @@ package("boost")
         local msvc
         local is_clang_tc = false
         if package:is_plat("windows") then
-            import("core.base.global")
-            import("core.project.config")
-
             -- Boost's bootstrap.bat needs the runenvs from the toolchain to locate cl
-            local tc_name = config.get("toolchain") or global.get("toolchain") or "msvc"
+            local tc_name = package:config("toolchains") or "msvc"
             msvc = toolchain.load(tc_name, {plat = package:plat(), arch = package:arch()})
 
             if tc_name == "clang-cl" then
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -138,11 +138,21 @@ package("boost")
 
         -- get msvc
         local msvc
+        local is_clang_tc = false
         if package:is_plat("windows") then
-            msvc = toolchain.load("msvc", {plat = package:plat(), arch = package:arch()})
+            import("core.base.global")
+            import("core.project.config")
+
+            -- Boost's bootstrap.bat needs the runenvs from the toolchain to locate cl
+            local tc_name = config.get("toolchain") or global.get("toolchain") or "msvc"
+            msvc = toolchain.load(tc_name, {plat = package:plat(), arch = package:arch()})
+
+            if tc_name == "clang-cl" then
+                is_clang_tc = true
+            end
         end
 
-        local is_clang_cl = false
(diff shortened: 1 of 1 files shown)


packages/b/boost/xmake.lua
```diff
@@ -158,22 +168,21 @@ package("boost")
                 file:print("using darwin : : %s ;", cc)
             elseif package:is_plat("windows") then
                 local vs_toolset = msvc:config("vs_toolset")
-                local toolset = "msvc"
                 local msvc_ver = ""
+                win_toolset = "msvc"
 
-                if cxx:find("clang%-cl$") or cxx:find("clang%-cl%.exe$") then
-                    toolset = "clang-win"
+                if is_clang_tc or cxx:find("clang%-cl$") or cxx:find("clang%-cl%.exe$") then
```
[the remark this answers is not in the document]
calebkiage: I wanted to make sure that we're matching on the file name here. If someone had a weird setup that had clang-cl as a folder name, I didn't want that to pass.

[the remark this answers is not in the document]
waruqi: we can use `package:has_tool("cxx", "clang_cl")`, it will be better.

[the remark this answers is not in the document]
calebkiage: I tried that, but it's empty for `xmake f --cxx=clang-cl`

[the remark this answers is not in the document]
waruqi: > I tried that, but it's empty for `xmake f --cxx=clang-cl`

Please do not switch clang-cl in this way, most packages do not support such switching.

Also, the windows platform is currently strictly bound to the msvc toolchain, and only the packages maintained by xmake.lua support switching to clang-cl.

Although you can modify a package to support the clang-cl toolchain, other packages do not currently support it either.

https://github.com/xmake-io/xmake/issues/3878

packages/b/boost/xmake.lua
```diff
@@ -138,11 +138,18 @@ package("boost")
 
         -- get msvc
         local msvc
+        local is_clang_tc = false
         if package:is_plat("windows") then
-            msvc = toolchain.load("msvc", {plat = package:plat(), arch = package:arch()})
+            -- Boost's bootstrap.bat needs the runenvs from the toolchain to locate cl
+            local tc_name = package:config("toolchains") or "msvc"
```
waruqi, line 144: please use always use msvc toolchain, msvc and clang-cl has same runenvs. and the windows platform is currently strictly bound to the msvc toolchain now.

calebkiage: Ok. I can close the PR and keep a local version of the boost package. As I said in a past comment, when using the clang-cl toolchain, `msvc:runenvs()` is empty so the boost bootstrap fails because it can't find the cl command to build b2 with.

waruqi: try it again.

calebkiage: I've tried it. It doesn't seem to work when using the default auto-detected toolchain...

calebkiage: I notice the pipelines also fail with that same error

calebkiage: Thinking about this a little bit, can xmake provide defaulted configs to packages? CMake for example compiles a list of all the used variables in a CMakeCache.txt file (including the auto-detected ones). I've noticed xmake only provides the explicitly defined configurations within many of its APIs.

PR #2265, opened 2023-07-03
benchmark: update url and test

2283572185:
The function `DoNotOptimize` was deprecated in this scope since 1.8.0 and the url was changed.


darcy-shen (approved):
LGTM

PR #2267, opened 2023-07-03
libpng: update url and use tar.gz instead zip

darcy-shen:

------

### DIFF
--- a/packages/l/libpng/xmake.lua
+++ b/packages/l/libpng/xmake.lua
@@ -9,7 +9,7 @@ package("libpng")
     add_versions("v1.6.37", "ca74a0dace179a8422187671aee97dd3892b53e168627145271cad5b5ac81307")
     add_versions("v1.6.36", "5bef5a850a9255365a2dc344671b7e9ef810de491bd479c2506ac3c337e2d84f")
     add_versions("v1.6.35", "6d59d6a154ccbb772ec11772cb8f8beb0d382b61e7ccc62435bf7311c9f4b210")
-    add_versions("vs1.6.34", "e45ce5f68b1d80e2cb9a2b601605b374bdf51e1798ef1c2c2bd62131dfcf9eef")
+    add_versions("v1.6.34", "e45ce5f68b1d80e2cb9a2b601605b374bdf51e1798ef1c2c2bd62131dfcf9eef")
 
     add_deps("zlib")
 
--- a/packages/l/libpng/xmake.lua
+++ b/packages/l/libpng/xmake.lua
@@ -3,13 +3,13 @@ package("libpng")
     set_description("The official PNG reference library")
     set_license("libpng-2.0")
 
-    set_urls("https://github.com/glennrp/libpng/archive/refs/tags/v$(version).tar.gz",
+    set_urls("https://github.com/glennrp/libpng/archive/refs/tags/$(version).tar.gz",
              "https://github.com/glennrp/libpng.git")
-    add_versions("1.6.40", "ab3f88779f0661bbb07c60e778fda782216bff70355d86848fbf6a327084563a")
-    add_versions("1.6.37", "c2c50c13a727af73ecd3fc0167d78592cf5e0bca9611058ca414b6493339c784")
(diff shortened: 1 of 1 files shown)


CyaNinix:
why? this doesn't seem right as the github tags are named with a starting "v"

SirLynix:
why? this doesn't seem right as the github tags are named with a starting "v"

SirLynix:
by the way you don't need to make another pull request to trigger CI again after changing base, just close and reopen the same PR

darcy-shen:
> why? this doesn't seem right as the github tags are named with a starting "v"

Because I want to add this mirror as the alternative url:

http://mirrors.ustc.edu.cn/debian/pool/main/libp/libpng1.6/libpng1.6_1.6.36.orig.tar.xz

For me, downloading from Github is unstable. I think use version without v is a better practice. What's your opnion? @waruqi

SirLynix:
it's not about opinion, xmake uses the version name as git tags if archive download fails.
so with `add_versions("v1.6.40", "hash")` xmake will try to download .tar.gz and then fallback on the git tag "v1.6.40" (see <https://github.com/glennrp/libpng/releases/tag/v1.6.40>)

You can add a filter on your new url to remove that starting v if it's not part of the url

darcy-shen:
> it's not about opinion, xmake uses the version name as git tags if archive download fails.

Let me find out why #2264 works fine.

SirLynix:
> > it's not about opinion, xmake uses the version name as git tags if archive download fails.
> 
> Let me find out why #2264 works fine.

Because it successfully downloaded the archive (and thus didn't test the git tag), but you can see on this CI that xmake tries to use the version as a git name:
```
  => download https://github.com/glennrp/libpng/archive/refs/tags/v1.6.40.tar.gz .. failed
/usr/bin/git clone https://github.com/glennrp/libpng.git -c core.fsmonitor=false source.tmp/libpng
Cloning into 'source.tmp/libpng'...
/usr/bin/git -c core.fsmonitor=false checkout 1.6.40
error: pathspec '1.6.40' did not match any file(s) known to git
```

darcy-shen:
I think it is because I have not update the hash.

SirLynix:
> I think it is because I have not update the hash.

Yes, the hash is wrong so it tried to fallback on git, and failed because you removed the starting "v".

You should put that starting `v` back and update hash, same for autodiff

darcy-shen:
> autodiff url/version naming is wrong and should be updated

You are right! It is not version, but tag.

packages/l/libpng/xmake.lua
```diff
@@ -3,13 +3,13 @@ package("libpng")
...
-    add_versions("1.6.40", "ab3f88779f0661bbb07c60e778fda782216bff70355d86848fbf6a327084563a")
-    add_versions("1.6.37", "c2c50c13a727af73ecd3fc0167d78592cf5e0bca9611058ca414b6493339c784")
-    add_versions("1.6.36", "6274d3f761cc80f7f6e2cde6c07bed10c00bc4ddd24c4f86e25eb51affa1664d")
-    add_versions("1.6.35", "3d22d46c566b1761a0e15ea397589b3a5f36ac09b7c785382e6470156c04247f")
-    add_versions("1.6.34", "7ffa5eb8f9f3ed23cf107042e5fec28699718916668bbce48b968600475208d3")
+    add_versions("v1.6.40", "62d25af25e636454b005c93cae51ddcd5383c40fa14aa3dae8f6576feb5692c2")
+    add_versions("v1.6.37", "ca74a0dace179a8422187671aee97dd3892b53e168627145271cad5b5ac81307")
+    add_versions("v1.6.36", "5bef5a850a9255365a2dc344671b7e9ef810de491bd479c2506ac3c337e2d84f")
+    add_versions("v1.6.35", "6d59d6a154ccbb772ec11772cb8f8beb0d382b61e7ccc62435bf7311c9f4b210")
+    add_versions("vs1.6.34", "e45ce5f68b1d80e2cb9a2b601605b374bdf51e1798ef1c2c2bd62131dfcf9eef")
```
SirLynix, line 12: vs => v

SirLynix (approved):

PR #2269, opened 2023-07-03
libpng: add_urls from deb-src

darcy-shen:
Currently, I'm using mirrors.ustc.edu.cn, and later, we can find a better way, let the user always choose the fastest mirrors.

eg.
```
add_ubuntu_archives("pool/main/libp/libpng1.6/libpng1.6_$(version).orig.tar.gz")
```

I think it would be a good way to improve locality of the package archives.

### DIFF
--- a/packages/l/libpng/xmake.lua
+++ b/packages/l/libpng/xmake.lua
@@ -3,8 +3,17 @@ package("libpng")
     set_description("The official PNG reference library")
     set_license("libpng-2.0")
 
-    set_urls("https://github.com/glennrp/libpng/archive/refs/tags/$(version).tar.gz",
-             "https://github.com/glennrp/libpng.git")
+    
+    add_urls("https://github.com/glennrp/libpng/archive/refs/tags/$(version).tar.gz")
+    add_urls(
+        "http://mirrors.ustc.edu.cn/ubuntu/pool/main/libp/libpng1.6/libpng1.6_$(version).orig.tar.gz",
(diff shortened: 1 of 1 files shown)


packages/l/libpng/xmake.lua
```diff
@@ -3,8 +3,17 @@ package("libpng")
     set_description("The official PNG reference library")
     set_license("libpng-2.0")
 
-    set_urls("https://github.com/glennrp/libpng/archive/refs/tags/$(version).tar.gz",
-             "https://github.com/glennrp/libpng.git")
+    
+    add_urls("https://github.com/glennrp/libpng/archive/refs/tags/$(version).tar.gz")
+    add_urls(
+        "https://mirrors.ustc.edu.cn/ubuntu/pool/main/libp/libpng1.6/libpng1.6_$(version).orig.tar.gz",
+        "https://mirrors.ustc.edu.cn/debian/pool/main/libp/libpng1.6/libpng1.6_$(version).orig.tar.gz",
```
waruqi, line 10: 都在一个 host 下面，下载速度一样的，就没必要加两个 url 了，没必要，保留一个就行了。。

packages/l/libpng/xmake.lua
```diff
@@ -3,8 +3,17 @@ package("libpng")
...
+        {
+            alias = "deb-src",
+            version = function (version) return string.sub(version, 2, string.len(version)) end
```
waruqi, line 13: 直接使用 `version:sub(2)` 就行了

PR #2275, opened 2023-07-05
add package dxsdk

ChrisCatCP:


SirLynix:
This packages misses protected vs_runtime and shared configs

waruqi:
> This packages misses protected vs_runtime and shared configs

has shared configs, just no vs_runtime

waruqi:
@ChrisCatCP Can you improve it again? add vs runtime as readonly and use `is_plat` and `is_arch` instead of is_host

SirLynix:
vs_runtime should be MD in release and MDd in debug, is there a way to do that?

ChrisCatCP:
i think vs_runtime is invalid here

PR #2276, opened 2023-07-05
dxsdk: Remove os check

SirLynix:
See #2275


SirLynix:
`vs_runtime` should be set to MD in release and MDd in debug (`package:debug()`) but I don't think we can do that

ChrisCatCP:
This is just a binary library and it not ref `vs_runtime`, so i dont set `vs_runtime`

SirLynix:
Setting `vs_runtime` readonly prevents xmake from reinstalling the package if the vs_runtime changes, which should be done on precompiled packages like this one.

The difficulty here is that there are two vs_runtime depending on the config, and since you need to know the config to set the vs_runtime, I don't think it's possible to make it readonly

PR #2281, opened 2023-07-05
Add support for Qt6 (again)

SirLynix:
See #1081


SirLynix:
there's an issue with find_qt for cross platforms (windows/arm, android and ios), it finds the host installation.

Fedora should work but requires system OpenSSL, which is why its ci fails.

otherwise, it can be merged and improved later.

PR #2282, opened 2023-07-06
protobuf-cpp: fix debug mode link

ChrisCatCP:
debug模式下会有lib会有后缀d 导致links里没有lib

### DIFF
--- a/packages/p/protobuf-cpp/xmake.lua
+++ b/packages/p/protobuf-cpp/xmake.lua
@@ -39,6 +39,7 @@ package("protobuf-cpp")
 
     on_install("windows", "linux", "macosx", function (package)
         os.cd("cmake")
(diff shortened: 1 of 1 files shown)


packages/p/protobuf-cpp/xmake.lua
```diff
@@ -39,6 +39,7 @@ package("protobuf-cpp")
 
     on_install("windows", "linux", "macosx", function (package)
         os.cd("cmake")
+        io.replace("CMakeLists.txt", "set(protobuf_DEBUG_POSTFIX \"d\"", "set(protobuf_DEBUG_POSTFIX \"\"", {plain = true})
```
SirLynix, line 42: It would be better to link the correct library name (libprotobuff(d)) in debug) than patch CMakeLists

ChrisCatCP: reason is the pkgconfig pc file dont has d
If you want to link the correct name, you need two patch files `protobuf-lite.pc.cmake` and `protobuf.pc.cmake`

ChrisCatCP:
@waruqi 这里还有问题吗  一直没有合并

PR #2287, opened 2023-07-07
Update ICU4C to 73.2

WSSDude:


WSSDude:
What could I do about Windows ARM64 build? It seems it tries to link x86 libs for some reason.

MinGW I think can be resolved by applying 72.1 patch file (as was in 73.1)

WSSDude:
I found an issue with using the lib in debug even on configs this compiles to according to CI.

Library seems to put "d" suffix similar to SDL2

I did the same sort of fix as was done in https://github.com/xmake-io/xmake-repo/pull/1865/files
It required to move some things into on_load()

packages/i/icu4c/xmake.lua
```diff
@@ -24,20 +24,24 @@ package("icu4c")
...
+    on_load(function (package)
+        local libsuffix = package:is_debug() and "d" or ""
+        package:add("links", "icutu"..libsuffix, "icuio"..libsuffix)
+        if is_plat("mingw", "windows") then
+            package:add("links", "icuin"..libsuffix, "icuuc"..libsuffix, "icudt"..libsuffix)
+        else
+            package:add("links", "icui18n"..libsuffix, "icuuc"..libsuffix, "icudata"..libsuffix)
+        end
+    
+        if is_plat("linux") then
```
SirLynix, line 40: you can leave this check out of on_load

WSSDude:
And btw, thank you @SirLynix for helping fix the build for Windows ARM64 🙇

packages/i/icu4c/xmake.lua
WSSDude: moved it back outside 👍

SirLynix:
Unfortunately it's not that easy, I think we'll have to upgrade some .vcxproj too

SirLynix:
> I found an issue with using the lib in debug even on configs this compiles with according to CI.
> 
> Library seems to put "d" suffix similar to SDL2


Are you sure about that d suffix? it breaks Linux debug compilation

WSSDude:
Im sure at least for VS2022 compiler, without it, debug config does not work on that compiler on both, MSVC and Clang-Cl toolchains. Tested it before inclusion.

Maybe it is Windows or otherwise compiler specific?

WSSDude:
To be specific, it died during package install step, didn't even let me build my project.

SirLynix:
@waruqi do you know why vcxproj upgrade seems to fail here?

@WSSDude I can confirm the "d" suffix on Windows, but it doesn't seem to be there on Linux

waruqi:
> @waruqi do you know why vcxproj upgrade seems to fail here?

I don't know. maybe incorrect win sdk version in sln file

> error MSB8036: The Windows SDK version 10.0.16299.0 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".

waruqi:
> @waruqi do you know why vcxproj upgrade seems to fail here?
> 
> @WSSDude I can confirm the "d" suffix on Windows, but it doesn't seem to be there on Linux

I have fixed vcproj upgrade, but you need update xmake to dev version to support for upgrading *.props

PR #2305, opened 2023-07-11
[fxdiv] Formatting

Qudix:
Simplify the [`fxdiv`](https://github.com/Maratyszcza/FXdiv) package. The cmakelists is only really used for building the tests and benchmarks, the library itself is a single header.

### DIFF
--- a/packages/f/fxdiv/xmake.lua
+++ b/packages/f/fxdiv/xmake.lua
@@ -1,5 +1,4 @@
 package("fxdiv")
-
     set_kind("library", {headeronly = true})
     set_homepage("https://github.com/Maratyszcza/FXdiv")
     set_description("C99/C++ header-only library for division via fixed-point multiplication by inverse")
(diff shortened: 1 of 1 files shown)


packages/f/fxdiv/xmake.lua
```diff
@@ -8,12 +7,10 @@ package("fxdiv")
     add_urls("https://github.com/Maratyszcza/FXdiv.git")
     add_versions("2020.12.09", "63058eff77e11aa15bf531df5dd34395ec3017c8")
 
-    add_deps("cmake")
-    on_install("windows", "macosx", "linux", function (package)
-        local configs = {"-DFXDIV_BUILD_TESTS=OFF", "-DFXDIV_BUILD_BENCHMARKS=OFF"}
-        import("package.tools.cmake").install(package, configs)
+    on_install("windows", "macosx", "linux", function(package)
+        os.cp("include", package:installdir())
```
waruqi, line 11: Please use the cmake that comes with the library to install it if possible, even if it is header only. This is because cmake also installs an additional xxx.cmake import file so that other cmake projects that depend on it can find it automatically.

packages/f/fxdiv/xmake.lua
```diff
@@ -9,11 +8,12 @@ package("fxdiv")
     add_versions("2020.12.09", "63058eff77e11aa15bf531df5dd34395ec3017c8")
 
     add_deps("cmake")
-    on_install("windows", "macosx", "linux", function (package)
+
+    on_install("windows", "macosx", "linux", function(package)
```
waruqi, line 12: please add space `function (`

Qudix: Why?

waruqi: Keeping the style consistent, you can see other packages

https://github.com/xmake-io/xmake-repo/blob/c53b49dbae4c330ae2cdf96ade2d63c08cec4150/packages/c/cairo/xmake.lua#L36

Qudix: They're not all consistent

PR #2310, opened 2023-07-12
[cimg] Add package

Qudix:
Add [`cimg`](https://github.com/greyclab/cimg) header-only image processing package. This one has a strange versioning quirk.

### DIFF
--- a/packages/c/cimg/xmake.lua
+++ b/packages/c/cimg/xmake.lua
@@ -10,7 +10,7 @@ package("cimg")
     add_urls("https://github.com/greyclab/cimg.git")
     add_versions("v3.2.6", "1fcca9a7a453aa278660c10d54c6db9b4c614b6a29250adeb231e95a0be209e7")
 
-    on_install(function (package)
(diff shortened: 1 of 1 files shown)


packages/c/cimg/xmake.lua
```diff
@@ -0,0 +1,21 @@
...
+    end})
+    add_urls("https://github.com/greyclab/cimg.git")
+    add_versions("v3.2.6", "1fcca9a7a453aa278660c10d54c6db9b4c614b6a29250adeb231e95a0be209e7")
+
+    on_install("windows", "linux", "macosx", "android", "mingw", "cygwin", "bsd", "cross", function (package)
+        os.cp("CImg.h", package:installdir("include"))
+    end)
+
+    on_test(function (package)
+        assert(package:has_cxxtypes("cimg_library::CImg<>", {
```
waruqi, line 18: you should use check_cxxsnippets to check links,  has_cxxtypes will not check links.

PR #2327, opened 2023-07-15
Fix ICU4C compilation for 64-bit Windows targets

WSSDude:
I dont know how tests passed, but the library after build on Windows 64-bit MSVC target was not copied correctly into bin and lib (folders were left empty)

There was platform being checked with `is_plat("*64")` instead of architecture with `is_arch(.*64)`.


WSSDude:
And this apparently breaks ARM again... 😞

updated
```diff
@@ -0,0 +1 @@
+{ }
```
waruqi, line 1: just close and open pr to trigger ci tests.

waruqi:
> And this apparently breaks ARM again... 😞

windows arm is optional, we can merge it first. then improve it for arm later.

PR #2395, opened 2023-08-04
[rsm-binary-io] Update package

Qudix:
- Add `rsm-binary-io` version `2.0.6`
- Author has specified that it does not support building as a shared library

### DIFF
--- a/packages/r/rsm-binary-io/xmake.lua
+++ b/packages/r/rsm-binary-io/xmake.lua
@@ -25,7 +25,7 @@ package("rsm-binary-io")
 
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
-            int main() {
+            void test() {
                 binary_io::span_istream s;
                 assert(s.tell() == 0);
             }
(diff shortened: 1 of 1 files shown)


waruqi:
> Author has specified that it does not support building as a shared library

But we can make it support shared, if it can be compiled well. Why not support it?

Qudix:
I'm just following the authors advice. It's my goal to have packages work the way they intend it to, and trying to get around that can lead to cornering yourself in the future, such as when they officially add their own exports. I'm making these changes now because nothing relies on them at the moment, none the least anything needing shared.

waruqi:
Many of the original packages are missing some shared king, or other platforms support, but since we just need to simply add some patch, we can make it work well for more platforms and types. 

Why not do it? I'd suggest keeping the current configuration, and if the author enables shared support in a new version, it's perfectly fine for us to go back and remove the explicit symbolic export.

waruqi:
Add shared, which does not interfere with the use of static libraries by existing users. It can also be a problem for users who want to use shared libraries, but find that they are still using static libraries. Since we can make it support shared, we should try to support it as much as possible.

Qudix:
Forget about it

PR #2396, opened 2023-08-04
[rsm-mmio] Edit package

Qudix:
- Author has specified that it does not support building as a shared library

### DIFF
--- a/packages/r/rsm-mmio/xmake.lua
+++ b/packages/r/rsm-mmio/xmake.lua
@@ -24,7 +24,7 @@ package("rsm-mmio")
 
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
-            void test() {
+            int main() {
                 mmio::mapped_file_source f;
                 assert(!f.is_open());
             }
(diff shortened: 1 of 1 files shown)


packages/r/rsm-mmio/xmake.lua
```diff
@@ -7,27 +7,24 @@ package("rsm-mmio")
...
-        local configs = {}
-        configs.kind = package:config("shared") and "shared" or "static"
-        import("package.tools.xmake").install(package, configs)
+        import("package.tools.xmake").install(package)
     end)
 
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
-            void test() {
+            int main() {
```
waruqi, line 27: please do not change it

packages/r/rsm-mmio/xmake.lua
```diff
@@ -7,27 +7,24 @@ package("rsm-mmio")
...
             set_languages("c++17")
             target("rsm-mmio")
-                set_kind("$(kind)")
+                set_kind("static")
                 add_files("src/**.cpp")
                 add_includedirs("include/")
                 add_headerfiles("include/(**.hpp)")
-                if is_plat("windows") and is_kind("shared") then
-                    add_rules("utils.symbols.export_all", {export_classes = true})
-                end
```
waruqi, line 21: But we can make it support shared, if it can be compiled well. Why not support it?

PR #2411, opened 2023-08-07
Add BlueZ

AidanSun05:


star-hengxing:
Please add on_test.

AidanSun05:
I have added on_test for the library.

AidanSun05:
@waruqi I saw your recent changes to the package, but now the compiler can't find the header files in the CI:

```
error: @programdir/modules/private/async/runjobs.lua:256: @programdir/modules/private/action/build/object.lua:91: @programdir/modules/core/tools/gcc.lua:797: lib/sdp.h:22:10: fatal error: bluetooth/bluetooth.h: No such file or directory
   22 | #include <bluetooth/bluetooth.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
  > in lib/sdp.c
```

What's the best way to handle multiple include paths that contain the same files? I'd like to know if there's a cleaner or more reliable solution than what I had earlier.

PR #2418, opened 2023-08-10
Update Vulkan SDK packages

SirLynix:
#2407 updated vulkan-headers and volk but did not update other vulkan packages causing issues, I added a warning to prevent it to happen in the future


SirLynix:
FreeBSD keeps failing on a authentication error issue, but I think this can be merged

PR #2441, opened 2023-08-14
Support search python in virtualenv

Yiklek:


packages/p/python/fetch.lua
```diff
@@ -54,14 +54,16 @@ function _find_library(package, opt)
     local includepath = nil
     if package:is_plat("windows") then
         link = format("python" .. table.concat(table.slice(version:split("%."), 1, 2), ""))
-        libpath = find_library(link, {exepath}, {suffixes = {"libs"}})
+        local out, _ = os.iorunv(program, { "-c", "import sys; print(sys.base_prefix, end='')" })
```
waruqi, line 57: it will break for python2, then fetch will abort

```lua
$ python2 -c "import sys; print(sys.base_prefix, end='')"
  File "<string>", line 1
    import sys; print(sys.base_prefix, end='')
                                          ^
SyntaxError: invalid syntax
```

you should add try-catch

```lua
local out = try { function () return os.iorunv(program, { "-c", "import sys; print(sys.base_prefix, end='')" }) end }
```

PR #2452, opened 2023-08-17
rsm-bsa: add package

star-hengxing:


Qudix:
You're missing the separate xmem binary target itself, at the moment if you try to use xmem support it will fail because the bsa library directly looks for `xmem.exe` and starts it.

star-hengxing:
https://github.com/Ryan-rsm-McKenzie/bsa/blob/938c6b5c960f5ba9c4f457f34a41665fd38023e0/extras/xmem/src/main.cpp#L38

`xmem.exe` needs x86 arch to compile, but xmake has been fixed yet.
https://github.com/xmake-io/xmake/issues/4043

Qudix:
Are you going to continue this?

star-hengxing:
Yes, but I've been busy lately and don't have another system (macosx/linux) to debug the library.

star-hengxing:
I think the failed ci does not support full C++20.

waruqi:
> I think the failed ci does not support full C++20.

so, you can disable some platforms, e.g. mac, android, ios.  They are unlikely to support it, at least for the next few years.

Qudix:
I convinced ryan to publish a new github release that includes better support for `directxtex` under linux

Qudix:
`add_resources` seems to always assume the file you are downloading is an archive. This situation seems like the perfect candidate to make a xmake-mirror repo to host the file.

This is the most up to date one (2022.08.01): https://github.com/dotnet/runtime/blob/main/src/coreclr/pal/inc/rt/sal.h

waruqi:
add_resources does not yet support direct downloads of non-archived files.

waruqi:
> `add_resources` seems to always assume the file you are downloading is an archive and runs extract on it. This situation seems like the perfect candidate to make a xmake-mirror repo to host the file.
> 
> This is the most up to date one (2022.08.01): https://github.com/dotnet/runtime/blob/main/src/coreclr/pal/inc/rt/sal.h

I have supported it, you can try it again.

star-hengxing:
directxmath, directx-headers and directxtex now support Linux.

star-hengxing:
@waruqi
Can merge? Just clang and vs2019 build failed?

PR #2474, opened 2023-08-20
[rsm-binary-io] Add `2.0.6`

Qudix:
- Add version `2.0.6`
- Remove unnecessary `kind` config, this is already passed by xmake
- Remove args from test function


Qudix:
The build failures are all the same: `undefined reference to 'fread_s'`, but only for mingw in the shared configuration.

waruqi:
> The build failures are all the same: `undefined reference to 'fread_s'`, but only for mingw in the shared configuration.

you can add patches to fix it, e.g. use fread instead of it.

or use readonly to disable shared for mingw.

Qudix:
Looking at a closed issue on the github repo, mingw isn't supported. Despite static building here, the repo tests don't pass, so I'll just disable mingw entirely.

PR #2510, opened 2023-08-27
seqan3: add package

star-hengxing:
- sdsl-lite
- seqan3


star-hengxing:
ci compiler does not support full c++20

waruqi:
> ci compiler does not support full c++20

you can disable some platform, e.g. ios, android, macos

Issue #2257, opened 2023-06-30
SDL Wayland support


Persson-dev:
### Is your feature request related to a problem? Please describe.

SDL currently supports X11 with the `with_x` config.
It would be nice to have support for Wayland as well (since I'm using it).

### Describe the solution you'd like

Adding a `with_wayland` config to compile with Wayland support.

waruqi:
you can open a pr to improve sdl package.

Persson-dev:
Fixed with #2563

PR #2553, opened 2023-09-03
Add Wayland support for SDL package

Persson-dev:

------

### DIFF
--- a/packages/l/libsdl/xmake.lua
+++ b/packages/l/libsdl/xmake.lua
@@ -61,6 +61,7 @@ package("libsdl")
     add_configs("use_sdlmain", {description = "Use SDL_main entry point", default = true, type = "boolean"})
     if is_plat("linux") then
         add_configs("with_x", {description = "Enables X support (requires it on the system)", default = true, type = "boolean"})
+        add_configs("with_wayland", {description = "Enables Wayland support (requires it on the system)", default = true, type = "boolean"})
     end
 
     if is_plat("wasm") then
@@ -74,6 +75,9 @@ package("libsdl")
         package:add("components", "lib")
         if package:is_plat("linux") and package:config("with_x") then
             package:add("deps", "libxext", {private = true})
+        end
+        if package:is_plat("linux") and package:config("with_wayland") then
+            package:add("deps", "wayland", {private = true})
         end
     end)
 
@@ -192,6 +196,9 @@ package("libsdl")
                 end
                 table.insert(configs, "-DCMAKE_INCLUDE_PATH=" .. table.concat(includedirs, ";"))
             end
+        if package:config("with_wayland") then
+            opt.envs = {PKG_CONFIG_PATH = package:dep("wayland"):installdir("lib", "pkgconfig")}
+        end
         elseif package:is_plat("wasm") then
             -- emscripten enables USE_SDL by default which will conflict with the sdl headers
             opt = opt or {}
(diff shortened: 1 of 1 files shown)


Persson-dev:
Should close #2257

packages/l/libsdl/xmake.lua
```diff
@@ -192,6 +196,9 @@ package("libsdl")
                 end
                 table.insert(configs, "-DCMAKE_INCLUDE_PATH=" .. table.concat(includedirs, ";"))
             end
+        if package:config("with_wayland") then
+            opt.envs = {PKG_CONFIG_PATH = package:dep("wayland"):installdir("lib", "pkgconfig")}
```
waruqi, line 200: you need only add `add_deps("wayland")`, it will add PKG_CONFIG_PATH automatically.

Persson-dev: In that case, without setting manually PKG_CONFIG_PATH, it does not detect wayland.
The thing is, sdl compiles fine without it, so it's hard to know whether or not it compiled with wayland.
I just tried to install libsdl again (with the official repo) so it downloaded wayland as expected but I got during the cmake process (of sdl) :
```
-- Checking for modules 'wayland-client>=1.18;wayland-egl;wayland-cursor;egl;xkbcommon>=0.5.0'
--   No package 'wayland-client' found
--   No package 'wayland-egl' found
--   No package 'wayland-cursor' found
``` 
and a little after
```
--   SDL_WAYLAND                 (Wanted: ON): OFF
--   SDL_WAYLAND_LIBDECOR        (Wanted: ON): OFF
--   SDL_WAYLAND_LIBDECOR_SHARED (Wanted: ON): OFF
--   SDL_WAYLAND_QT_TOUCH        (Wanted: ON): OFF
--   SDL_WAYLAND_SHARED          (Wanted: ON): OFF
```
even though sdl was built just fine
I have no idea on how, but I should try to add a test if sdl compiled with wayland

waruqi: try this patch https://github.com/xmake-io/xmake/pull/4165

```
xmake update -s github:xmake-io/xmake#cmake
```

Persson-dev: I just did
```
xmake update -s github:xmake-io/xmake#cmake
xmake require -v -j 6 --extra={system=false} libsdl
```
and it still does not work

waruqi: It should work, you can print buildenvs and check PKG_CONFIG_PATH

```
package("libsdl")
    on_install()
        print(import("package.tools.cmake").buildenvs(package))
    end
```

waruqi: try it again. 

```
xmake update -s github:xmake-io/xmake#cmake
```

packages/l/libsdl/xmake.lua
```diff
@@ -61,6 +61,7 @@ package("libsdl")
     add_configs("use_sdlmain", {description = "Use SDL_main entry point", default = true, type = "boolean"})
     if is_plat("linux") then
         add_configs("with_x", {description = "Enables X support (requires it on the system)", default = true, type = "boolean"})
+        add_configs("with_wayland", {description = "Enables Wayland support (requires it on the system)", default = true, type = "boolean"})
```
waruqi, line 64: We shouldn't enable x11 and wayland at the same time, it will lead to installing a lot of dependencies. Users don't usually use both at the same time.

If you need wayland, you can enable it yourself, `{configs = {with_wayland = true}}`.

I will disable it by default.

waruqi: I have improved this package. https://github.com/xmake-io/xmake-repo/pull/2563

PR #2572, opened 2023-09-06
update vulkan-memory-allocator-hpp

Arthapz:


Latias94:
Should we make the vma version that vma-hpp depends on consistent with vma itself?

waruqi:
> Should we make the vma version that vma-hpp depends on consistent with vma itself?

You can use `add_requireconfs("vulkan-memory-allocator-hpp.vulkan-memory-allocator", {version = "1.0", override = true})` to rewrite it's dep version in your xmake.lua

PR #2635, opened 2023-09-22
Update MKL to 2023.2.0

samayala22:
I'm not xmake expert so I simply added MKL version 2023.2.0 checksums and urls for all platforms.  Checksums were calculated on a windows machine with powershell command:
```
(Get-FileHash <file>).Hash.ToLower();
```

I verified that I could reproduce the checksums of older versions. 

Tests passed when I did `xmake l scripts/test.lua -v mkl` but I couldn't test directly in my project because I don't know how. 

Maybe @xq114 knows if I need to add anything else. I don't see anything in the MKL release notes that point to changes in the Link Line Advisor.


xq114:
It seems fine, and better if change the `on_install("windows", ...` to `on_install("windows|x86", "windows|x64", ...` to disable arm. Actually I'd recommend install MKL by Intel SDK and let xmake just fetch it from system, 'cause the binaries from anaconda could be lacking some configurations (e.g. shared dlls)

Btw the hash algorithm is built-in in xmake, so you can try `xmake lua hash.sha256 <file>`

PR #2643, opened 2023-09-23
libevdev: add package

soiamsoNG:
add package libevdev


packages/l/libevdev/xmake.lua
```diff
@@ -0,0 +1,32 @@
+package("libevdev")
+
+    set_kind("library")
```
waruqi, line 3: it is default kind, we need not to add it.

PR #2657, opened 2023-09-27
Auto-update libsndfile to 1.2.2

SirLynix:
New version of libsndfile detected (package version: 1.0.31, last github version: 1.2.2)


jingkaimori:
maybe mpg123 is introduced as requirement on macos

SirLynix:
mpg123 should only be used if detected on the system. What I don't understand is that we don't have any cmake logs to see if that's what happened

jingkaimori:
maybe custom detection script of mpg123 in tarball should be removed, thus script in xmake repository will be used

PR #2679, opened 2023-10-06
Auto-update libsdl to 2.28.4

SirLynix:
New version of libsdl detected (package version: 2.28.3, last github version: 2.28.4)


packages/l/libsdl/xmake.lua
```diff
@@ -86,6 +88,9 @@ package("libsdl")
         if package:is_plat("linux") and package:config("wayland") then
             package:add("deps", "wayland", {private = true})
         end
+        if package:is_plat("wasm") and (package:gitref() or package:version():ge("2.28.4")) then
+            package:add("ldflags", "sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=stringToUTF8,UTF8ToString")
```
waruqi, line 92: `sDEFAULT`? `-sDEFAULT`?

SirLynix:
Okay so this was just a emscripten version issue. It looks like this change is the culprit:
> Symbols that were previously exported from native code, but only for internal reasons, are no longer exported on the Module object by default. For example, previously if you were using Module._malloc but not explicitly including _malloc in EXPORTED_FUNCTIONS, it might have been exported anyway due to internal use of malloc within the JS library code. (#18564)

from [emscripten changelog (3.1.31)](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3131---012623)

PR #2703, opened 2023-10-11
libnet: add version 1.2

RunThem:
[libnet](https://github.com/libnet/libnet)

### DIFF
--- a/packages/l/libnet/xmake.lua
+++ b/packages/l/libnet/xmake.lua
@@ -2,9 +2,12 @@ package("libnet")
     set_homepage("https://codedocs.xyz/libnet/libnet/")
     set_description("A portable framework for low-level network packet construction")
 
-    add_urls("https://github.com/libnet/libnet/archive/refs/tags/$(version).tar.gz",
-             "https://github.com/libnet/libnet.git")
-    add_versions("v1.2", "b7a371a337d242c017f3471d70bea2963596bec5bd3bd0e33e8517550e2311ef")
+    add_urls("https://github.com/libnet/libnet/releases/download/$(version).tar.gz", {version = function (version)
+        return version .. "/libnet-" .. (version:gsub("v", ""))
+    end})
+    add_urls("https://github.com/libnet/libnet.git")
+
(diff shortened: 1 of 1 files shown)


packages/l/libnet/xmake.lua
```diff
@@ -0,0 +1,19 @@
+package("libnet")
+    set_homepage("https://codedocs.xyz/libnet/libnet/")
+    set_description("A portable framework for low-level network packet construction")
+
+    add_urls("https://github.com/libnet/libnet/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/libnet/libnet.git")
+    add_versions("v1.2", "b7a371a337d242c017f3471d70bea2963596bec5bd3bd0e33e8517550e2311ef")
+
+    add_deps("autoconf", "automake", "libtool")
```
waruqi, line 9: tar.gz has configure script, we need not add them

```lua
    on_load(function (package)
        if package:gitref() then
            package:add("deps", "automake", "autoconf", "libtool")
        end
    end)
```

packages/l/libnet/xmake.lua
```diff
@@ -0,0 +1,19 @@
...
+
+    on_install(function (package)
```
waruqi, line 11: please disable some unsupported platforms. you can refer other package.

RunThem: This library supports FreeBSD, so what platform name should be filled in.

waruqi: > This library supports FreeBSD, so what platform name should be filled in.

bsd

packages/l/libnet/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+             "https://github.com/libnet/libnet.git")
+    add_versions("v1.2", "b7a371a337d242c017f3471d70bea2963596bec5bd3bd0e33e8517550e2311ef")
+
+    on_load(function (package)
+        if package:gitref() then
+            package:add("deps", "automake", "autoconf", "libtool")
+        end
+    end)
+
+    on_install("linux", "macosx", "windows", function (package)
```
waruqi, line 15: autoconf does not support windows. you can try `mingw`

PR #2760, opened 2023-10-27
Add dynareadout 23.10

PucklaJ:


PucklaJ:
@waruqi FreeBSD failed unfortunately, but all supported platforms work fine

PR #2765, opened 2023-10-28
add Pahomqttcpp package

Altyrost:


packages/p/pahomqttcpp/xmake.lua
```diff
@@ -0,0 +1,38 @@
...
+    add_deps("pahomqttc")
+
+    on_install("windows", "linux", "macos", function (package)
+        local pahomqttc = package:dep("pahomqttc")
+        
+        local configs = {"-DPAHO_BUILD_SAMPLES=FALSE",
+                         "-DPAHO_WITH_SSL=FALSE",
+                         "-DPAHO_BUILD_DOCUMENTATION=FALSE"}
+        table.insert(configs, "-DCMAKE_PREFIX_PATH=" .. pahomqttc:installdir())
+        if is_plat("windows") then
```
waruqi, line 21: package:is_plat

packages/p/pahomqttcpp/xmake.lua
```diff
@@ -0,0 +1,38 @@
...
+    add_deps("cmake")
+    add_deps("pahomqttc")
+
+    on_install("windows", "linux", "macos", function (package)
+        local pahomqttc = package:dep("pahomqttc")
+        
+        local configs = {"-DPAHO_BUILD_SAMPLES=FALSE",
+                         "-DPAHO_WITH_SSL=FALSE",
+                         "-DPAHO_BUILD_DOCUMENTATION=FALSE"}
+        table.insert(configs, "-DCMAKE_PREFIX_PATH=" .. pahomqttc:installdir())
```
waruqi, line 20: please remove it, xmake will add it

Altyrost: The build fails on linux for gcc and clang if I remove "-DPAHO_WITH_SSL=FALSE". For some reason, cmake keeps trying to find OpenSSL in that configuration without it.

packages/p/pahomqttcpp/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+    
+    add_deps("cmake")
+    add_deps("pahomqttc")
+
+    on_install("windows", "linux", "macos", function (package)
+        local pahomqttc = package:dep("pahomqttc")
+        
+        local configs = {"-DPAHO_WITH_SSL=FALSE"}
+        if package:is_plat("windows") then
+            table.insert(configs, "-DPAHO_MQTT_C_LIBRARIES=" .. pahomqttc:installdir("lib"))
```
waruqi, line 19: we should use pahomqttc:fetch(), because xmake maybe find it from system lib path.  installdir may be an empty directory.

https://github.com/xmake-io/xmake-repo/blob/ebe61bb7b8d7ecc11b3258747c80863f02084f74/packages/s/sol2/xmake.lua#L40

packages/p/pahomqttcpp/xmake.lua
waruqi: you can disable it first. if you want to improve it to support ssl.

you can patch cmakelists.tx to disable find openssl, and pass `{packagedeps = "openssl"}` to cmake.install to force set their include/lib flags.

you can see other packages.

packages/p/pahomqttcpp/xmake.lua
```diff
@@ -0,0 +1,45 @@
...
+            local pahomqttc = package:dep("pahomqttc"):fetch()
+
+            local includedirs = pahomqttc.includedirs or pahomqttc.sysincludedirs
+            if includedirs and #includedirs > 0 then
+                table.insert(configs, "-DPAHO_MQTT_C_INCLUDE_DIRS=" .. table.concat(includedirs, " "))
+            end
+
+            local linkdirs = pahomqttc.linkdirs
+            if linkdirs then
+                table.insert(configs, "-DPAHO_MQTT_C_LIBRARIES=" .. table.concat(linkdirs, " "))
```
waruqi, line 30: why not libfiles?

packages/p/pahomqttcpp/xmake.lua
```diff
@@ -0,0 +1,45 @@
...
+
+    add_urls("https://github.com/eclipse/paho.mqtt.cpp/archive/refs/tags/$(version).zip",
+             "https://github.com/eclipse/paho.mqtt.cpp.git")
+
+    add_versions("v1.2.0", "90c4d8ae4f56bb706120fddcc5937cd0a0360b6f39d5cd5574a5846c0f923473")
+    
+    add_deps("cmake")
+    add_deps("pahomqttc")
+
+    on_install("windows", "linux", "macos", function (package)
```
waruqi, line 14: macos -> macosx

packages/p/pahomqttcpp/xmake.lua
Altyrost: a misunderstanding on my end. I've corrected it.

PR #2766, opened 2023-10-28
wayland-protocols: Improve rule wayland.protocols

SuperSodaSea:
It seems that The `wayland.protocols` rule in package `wayland-protocols` didn't seem to work at all before. This PR fixes the rule and adds more options to the rule to provide more flexibility.

```lua
add_rules("@wayland-protocols/wayland.protocols", {
    outputdir = "...",  -- Path for generated files, default is `path.join(target:autogendir(), "rules", "wayland.protocols")`
    client = "...",     -- Path format for client protocol header files, default is `"%s.h"`, set to `false` to disable its generation
    server = "...",     -- Path format for server protocol header files, default is `nil`
    code = "...",       -- Path format for code files, default is `"%s.c"`
    public = false,     -- Visibility for headers and symbols in generated code, can be `false` or `true`, default is `false`
})
```

### DIFF
--- a/packages/w/wayland-protocols/rules/wayland_protocols.lua
+++ b/packages/w/wayland-protocols/rules/wayland_protocols.lua
@@ -61,15 +61,11 @@ rule("wayland.protocols")
                 -- For C++ module dependency discovery
                 if client then
                     local clientfile = path.join(outputdir, client:format(basename))
-                    if not os.exists(clientfile) then
-                        os.touch(clientfile)
-                    end
+                    os.touch(clientfile)
                 end
                 if server then
                     local serverfile = path.join(outputdir, server:format(basename))
-                    if not os.exists(serverfile) then
-                        os.touch(serverfile)
-                    end
+                    os.touch(serverfile)
                 end
 
                 -- Add code file to target
--- a/packages/w/wayland-protocols/rules/wayland_protocols.lua
+++ b/packages/w/wayland-protocols/rules/wayland_protocols.lua
@@ -95,9 +95,7 @@ rule("wayland.protocols")
         end
         if client then
             local clientfile = path.join(outputdir, client:format(basename))
-
             local client_args = {"client-header", sourcefile, clientfile}
-
             batchcmds:show_progress(opt.progress, "${color.build.object}generating.wayland.protocol.client %s", basename)
             batchcmds:vexecv(wayland_scanner.program, client_args)
         end
@@ -106,22 +104,18 @@ rule("wayland.protocols")
         local server = target:extraconf("rules", rule_name, "server")
         if server then
             local serverfile = path.join(outputdir, server:format(basename))
-
             local server_args = {"server-header", sourcefile, serverfile}
-
             batchcmds:show_progress(opt.progress, "${color.build.object}generating.wayland.protocol.server %s", basename)
             batchcmds:vexecv(wayland_scanner.program, server_args)
         end
 
         -- Generate code
-        local code = target:extraconf("rules", rule_name, "code") or "%s.c"
-        local codefile = path.join(outputdir, code:format(basename))
-
         local public = target:extraconf("rules", rule_name, "public")
         local visibility = public and "public" or "private"
 
+        local code = target:extraconf("rules", rule_name, "code") or "%s.c"
+        local codefile = path.join(outputdir, code:format(basename))
         local code_args = {visibility .. "-code", sourcefile, codefile}
-
         batchcmds:show_progress(opt.progress, "${color.build.object}generating.wayland.protocol.%s %s", visibility, basename)
         batchcmds:vexecv(wayland_scanner.program, code_args)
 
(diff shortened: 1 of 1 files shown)


Arthapz (approved):

Arthapz:
Lgtm

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
-
-                -- for c++ module dependency discovery
-                if not os.exists(clientfile) then
-                    os.touch(clientfile)
+        local sourcebatches = target:sourcebatches()[rule_name]
+        if not dryrun and sourcebatches and sourcebatches.sourcefiles then
+            local client = target:extraconf("rules", rule_name, "client")
+            if client == nil then
+                client = "%s.h"
+            end
```
waruqi, line 56:
```
local client = target:extraconf("rules", rule_name, "client") or "%s.h"
```

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+            local server = target:extraconf("rules", rule_name, "server")
+            for _, protocol in ipairs(sourcebatches.sourcefiles) do
+                local basename = path.basename(protocol)
+
+                -- For C++ module dependency discovery
+                if client then
+                    local clientfile = path.join(outputdir, client:format(basename))
+                    if not os.exists(clientfile) then
```
waruqi, line 64: os.isfile

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+            if client == nil then
+                client = "%s.h"
+            end
+            local server = target:extraconf("rules", rule_name, "server")
+            for _, protocol in ipairs(sourcebatches.sourcefiles) do
+                local basename = path.basename(protocol)
+
+                -- For C++ module dependency discovery
+                if client then
+                    local clientfile = path.join(outputdir, client:format(basename))
```
waruqi, line 63: why use `format` without args?

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+                    if not os.exists(clientfile) then
+                        os.touch(clientfile)
+                    end
+                end
+                if server then
+                    local serverfile = path.join(outputdir, server:format(basename))
```
waruqi, line 69: you can call `os.touch` directly

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
         local wayland_scanner = find_tool("wayland-scanner")
         assert(wayland_scanner, "wayland-scanner not found! please install wayland package")
 
-        local clientfile = path.join(outputdir, path.basename(sourcefile) .. ".h")
-        local privatefile = path.join(outputdir, path.basename(sourcefile) .. ".c")
+        local basename = path.basename(sourcefile)
+
+        -- Generate client protocol header
+        local client = target:extraconf("rules", rule_name, "client")
+        if client == nil then
```
waruqi, line 97: and here

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+            client = "%s.h"
+        end
+        if client then
+            local clientfile = path.join(outputdir, client:format(basename))
+
```
waruqi, line 102: remove empty line

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+            local client_args = {"client-header", sourcefile, clientfile}
+
```
waruqi, line 104: remove empty line

packages/w/wayland-protocols/rules/wayland_protocols.lua
```diff
@@ -14,67 +14,122 @@
...
+            batchcmds:vexecv(wayland_scanner.program, client_args)
+        end
+
+        -- Generate server protocol header
+        local server = target:extraconf("rules", rule_name, "server")
+        if server then
+            local serverfile = path.join(outputdir, server:format(basename))
+
+            local server_args = {"server-header", sourcefile, serverfile}
+
```
waruqi, line 115: remove empty line

packages/w/wayland-protocols/rules/wayland_protocols.lua
SuperSodaSea: I would like to use `false` in this option in order to disable the generation of client headers here (e.g. if one only wants to generate server headers), so we cannot simply write `or "%s.h"` here.

packages/w/wayland-protocols/rules/wayland_protocols.lua
SuperSodaSea:
```lua
local client = "%s-client-protocol.h"
local basename = "abc123"
client:format(basename) == "abc123-client-protocol.h"
```

PR #2781, opened 2023-11-01
pango: Fix unexpected -Werror=... errors

SuperSodaSea:
See also: #2780, <https://gitlab.gnome.org/GNOME/pango/-/issues/740>


waruqi:
and macos

SuperSodaSea:
openssl 1.1.1-t seems to be broken for some reason...

waruqi:
maybe this patch broken it. we will fix it. https://github.com/xmake-io/xmake/pull/4358

SuperSodaSea:
Seems that OpenBSD (a8b7d88613e9167653478fff6b330770c32a4a0d) is broken?

Issue #2791, opened 2023-11-04
Add Boost interval library


Theramar:
### Is your feature request related to a problem? Please describe.

It seems that Boost is the sole library offering interval arithmetic.

### Describe the solution you'd like

[Boost Interval](https://www.boost.org/doc/libs/1_83_0/libs/numeric/interval/doc/interval.htm)

PR #2809, opened 2023-11-08
Added Boost icl (interval container library) xmake.lua

Theramar:
#2791


Theramar:
Boost icl needs Boost assert. Another PR is required to add Boost assert

Issue #2819, opened 2023-11-10
drogon redis link error


tashcan:
### Xmake Version

2.8.5

### Operating System Version and Architecture

macOS 14.1

### Describe Bug

At least on Mac, when adding drogon as a package it fails with the following error:
```  => install drogon v1.8.2 .. failed

ld: Undefined symbols:
  _redisAsyncConnect, referenced from:
      drogon::nosql::RedisConnection::startConnectionInLoop() in libdrogon.a[49](RedisConnection.cc.o)
  _redisAsyncDisconnect, referenced from:
      drogon::nosql::RedisConnection::~RedisConnection() in libdrogon.a[46](RedisClientImpl.cc.o)
      std::__1::__function::__func<drogon::nosql::RedisConnection::disconnect()::$_6, std::__1::allocator<drogon::nosql::RedisConnection::disconnect()::$_6>, void ()>::operator()() in libdrogon.a[49](RedisConnection.cc.o)
  _redisAsyncFormattedCommand, referenced from:
      drogon::nosql::RedisConnection::sendCommandInLoop(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::function<void (drogon::nosql::RedisResult const&)>&&, std::__1::function<void (drogon::nosql::RedisException const&)>&&) in libdrogon.a[49](RedisConnection.cc.o)
      drogon::nosql::RedisConnection::sendSubscribeInLoop(std::__1::shared_ptr<drogon::nosql::SubscribeContext> const&) in libdrogon.a[49](RedisConnection.cc.o)
      drogon::nosql::RedisConnection::sendUnsubscribe(std::__1::shared_ptr<drogon::nosql::SubscribeContext> const&) in libdrogon.a[49](RedisConnection.cc.o)
      drogon::nosql::RedisConnection::sendUnsubscribeInLoop(std::__1::shared_ptr<drogon::nosql::SubscribeContext> const&) in libdrogon.a[49](RedisConnection.cc.o)
      drogon::nosql::RedisConnection::sendUnsubscribeInLoop(std::__1::shared_ptr<drogon::nosql::SubscribeContext> const&) in libdrogon.a[49](RedisConnection.cc.o)
[...]

PR #2823, opened 2023-11-11
drogon: fix redis build (#2819)

tashcan:
Add hiredis as an option to drogon. On systems with hiredis installed as a system library drogon will enable redis support but without the hiredis dependency fail to link.

Redis is now optional and if enabled with all the required dependency


tashcan:
See issue #2819

PR #2990, opened 2023-12-24
added llvm 13.0.0, 15.0.7 support for x86 macos

robcholz:
added llvm 13.0.0, 15.0.7 support for x86_64 macOS

### DIFF
--- a/packages/l/llvm/xmake.lua
+++ b/packages/l/llvm/xmake.lua
@@ -21,10 +21,10 @@ package("llvm")
     elseif is_plat("macosx") then
         if is_arch("x86_64") then
                 set_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version).tar.xz", {version = function (version)
-                if version:lt("15.0.7") then
-                    return version .. "/clang+llvm-" .. version .. "-x86_64-apple-darwin21.0"
-                else
+                if version:lt("15.0") then
                     return version .. "/clang+llvm-" .. version .. "-x86_64-apple-darwin"
+                else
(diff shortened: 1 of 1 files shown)


packages/l/llvm/xmake.lua
```diff
@@ -21,8 +21,11 @@ package("llvm")
     elseif is_plat("macosx") then
         if is_arch("x86_64") then
             set_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version)/clang+llvm-$(version)-x86_64-apple-darwin.tar.xz")
+            set_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version)/clang+llvm-$(version)-arm64-apple-darwin21.0.tar.xz", {version = "15.0.7"})
```
waruqi, line 24: ? `version = "15.0.7"` is incorrect syntax

robcholz: the version 15.0.7 for macosx on x86_64 is an exception, it uses a url different from the `set_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version)/clang+llvm-$(version)-x86_64-apple-darwin.tar.xz")`. is there a way to achieve this?

waruqi:
```lua
set_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version).tar.xz", {version = function (version)
    if version:lt("15.0") then
        return version .. "/clang+llvm-" .. version .. "-x86_64-apple-darwin"
    else
       -- TODO
    end
end})
```

PR #3015, opened 2023-12-31
tinyalloc: add package

star-hengxing:


packages/t/tinyalloc/xmake.lua
```diff
@@ -0,0 +1,28 @@
...
+
+    on_install(function (package)
+        io.writefile("xmake.lua", [[
+            add_rules("mode.release", "mode.debug")
+            target("tinyalloc")
+                set_kind("$(kind)")
+                add_files("tinyalloc.c")
+                add_headerfiles("tinyalloc.h")
+                if is_mode("debug") then
+                    add_defines("TA_DEBUG")
```
SirLynix, line 17: Remove TA_DEBUG in debug, it only works when embedding the library and implementing print_s and print_i ourselves

PR #3017, opened 2023-12-31
cmake: Update to 3.28.1 and enable parallel build on FreeBSD

SirLynix:
FreeBSD CI always takes a lot of time to install cmake because it does not parallelize build

Edit: it doesn't seem to help


packages/c/cmake/xmake.lua
```diff
@@ -106,7 +119,7 @@ package("cmake")
     end)
 
     on_install("@bsd", function (package)
-        os.vrunv("sh", {"./bootstrap", "--prefix=" .. package:installdir()})
+        os.vrunv("sh", {"./bootstrap", "--parallel=6", "--prefix=" .. package:installdir()})
```
waruqi, line 122: we can get jobs number  https://github.com/xmake-io/xmake/blob/da3b723be857e3a2fe4e229265caf6246db1f0dd/xmake/modules/package/tools/cmake.lua#L36

PR #3048, opened 2024-01-06
igraph: add package

ZacharyJia:
add igraph package


szhorvat:
You are fetching the wrong archive. 

DO use this: https://github.com/igraph/igraph/releases/download/0.10.8/igraph-0.10.8.tar.gz

Do NOT use this: https://github.com/igraph/igraph/archive/refs/tags/0.10.8.tar.gz

You should not change `IGRAPH_VERSION`. It is already included in the release archive.

PR #3049, opened 2024-01-06
some improvements on ncurses

ookiineko:
hi

i wanted to cross compile my project that uses this library for arm64 linux from an x86_64 linux host,

i used `xmake f -p cross --cross=aarch64-linux-gnu-` (i have the cross gcc and binutils installed on my host, sysroot is at `/usr/aarch64-linux-gnu` and gcc is prefixed by `aarch64-linux-gnu-` and its installed to `/usr/bin`),

but i had got a problem while configuring, a warning complains about ncurses is not supported for `cross/arm64`,
so i had a quick check of the source and found the supported platform list is hardcoded...

but luckily after removing it and tweaking some ac config options, it does seems to build and work with my original project.

also i thought adding these extsources might make it more friendly for some users?

### DIFF
--- a/packages/n/ncurses/xmake.lua
+++ b/packages/n/ncurses/xmake.lua
@@ -14,8 +14,12 @@ package("ncurses")
 
     add_configs("widec", {description = "Compile with wide-char/UTF-8 code.", default = true, type = "boolean"})
 
-    if is_plat("linux") then
-        add_extsources("apt::libncurses-dev")
+    if is_plat("mingw") and is_subhost("msys") then
+        add_extsources("pacman::ncurses")
+    elseif is_plat("linux") then
+        add_extsources("apt::libncurses-dev", "pacman::ncurses")
+    elseif is_plat("macosx") then
+        add_extsources("brew::ncurses")
     end
     on_load(function (package)
         if package:config("widec") then
--- a/packages/n/ncurses/xmake.lua
+++ b/packages/n/ncurses/xmake.lua
@@ -27,9 +27,15 @@ package("ncurses")
         end
     end)
 
-    on_install("linux", "macosx", "bsd", function (package)
-        local configs = {"--without-manpages", "--enable-sigwinch", "--with-gpm=no"}
-        table.insert(configs, "--with-debug=" .. (package:debug() and "yes" or "no"))
+    on_install(function (package)
+        local configs = {"--with-debug=" .. (package:debug() and "yes" or "no"),
+                         "--without-manpages",
+                         "--enable-sigwinch",
+                         "--with-gpm=no",
+                         "--without-ada",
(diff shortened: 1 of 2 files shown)


ookiineko:
another weird behavior is i cannot use linux as platform type

these attempts were failed (because it detects x86_64 which is my host architecture, and it's wrong since i am trying to cross compile for arm64):

```shell
❯ xmake f -m debug -p linux --cross=aarch64-linux-gnu-
checking for architecture ... x86_64
^C
❯ xmake f -m debug -p linux --cross=aarch64-linux-gnu- --bin=/usr/bin --sdk=/usr/aarch64-linux-gnu
checking for architecture ... x86_64
^C
```

ookiineko:
> it detects x86_64 which is my host architecture

i tried to correct it by specifiying `-a arm64` manually, but the build still eventually fail at some where...

ookiineko:
i have also tried vcpkg, but it ran into this issue:

```shell
❯ rm -rf ~/.xmake/ .xmake build /tmp/.xmake1000
❯ vcpkg install ncurses:arm64-linux
Computing installation plan...
The following packages are already installed:
    ncurses:arm64-linux@6.4#1
ncurses:arm64-linux is already installed
Total install time: 24.8 us
The package ncurses is compatible with built-in CMake variables:

    set(CURSES_NEED_NCURSES TRUE)
    find_package(Curses REQUIRED)
    target_include_directories(main ${CURSES_INCLUDE_DIRS})
    target_compile_options(main ${CURSES_CFLAGS})
    target_link_libraries(main PRIVATE ${CURSES_LIBRARIES})

❯ xmake f -m debug -p linux -a arm64 --cross=aarch64-linux-gnu-
updating repositories .. ok
note: install or modify (m) these packages (pass -y to skip confirm)?
in vcpkg:
  -> vcpkg::ncurses latest 
please input: y (y/n/m)
y
  => install vcpkg::ncurses latest .. failed

fetch vcpkg::ncurses-latest failed!
if you want to get more verbose errors, please see:
  -> /home/yurri/nSnake-Xmake/build/.packages/v/vcpkg_ncurses/latest/cache/installdir.failed/logs/install.txt
error: install failed!
❯ cat /home/yurri/nSnake-Xmake/build/.packages/v/vcpkg_ncurses/latest/cache/installdir.failed/logs/install.txt
fetch vcpkg::ncurses-latest failed
```

ookiineko:
humm, looks like windows and macos builds are broken?

waruqi:
> humm, looks like windows and macos builds are broken?

you need remove `--without-progs` and we should disable windows platform.

ookiineko:
i do have a windows and mac pc, so i guess let's see when i will have chance to test build on them and fix...

PR #3106, opened 2024-01-15
Auto-update filament to 1.49.3

waruqi:
New version of filament detected (package version: 1.32.0, last github version: 1.49.3)

### DIFF
--- a/packages/f/filament/xmake.lua
+++ b/packages/f/filament/xmake.lua
@@ -6,7 +6,8 @@ package("filament")
 
     if is_plat("windows") and is_arch("x64") then
         add_urls("https://github.com/google/filament/releases/download/v$(version)/filament-v$(version)-windows.tgz")
-        add_versions("1.9.23", "8f5728dd944c052f991bfb15d18cd6fc8f678d0cbd37bb066d76bf682ab789c8")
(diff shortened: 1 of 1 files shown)


packages/f/filament/xmake.lua
```diff
@@ -6,7 +6,8 @@ package("filament")
 
     if is_plat("windows") and is_arch("x64") then
         add_urls("https://github.com/google/filament/releases/download/v$(version)/filament-v$(version)-windows.tgz")
-        add_versions("1.9.23", "8f5728dd944c052f991bfb15d18cd6fc8f678d0cbd37bb066d76bf682ab789c8")
+        add_versions("1.49.3", "f465e45033b928a92e5b4a5a3b6534022f5428bfdb8016c953204f63842ac55c")
+    add_versions("1.9.23", "8f5728dd944c052f991bfb15d18cd6fc8f678d0cbd37bb066d76bf682ab789c8")
```
SirLynix, line 10: @waruqi I think we should detect when add_versions are behind conditions and disable auto-update, what do you think?

waruqi: I have improved it.

Issue #3128, opened 2024-01-18
Add support for the squat library


EdMUK:
### Is your feature request related to a problem? Please describe.

no

### Describe the solution you'd like

Add support for the squat binding library for the squirrel scripting language

Issue #3127, opened 2024-01-18
Add support for the Squirrel scripting language


EdMUK:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support for the Squirrel scripting language.

apocelipes:
Squirrel is introduced by #4874, so close this issue. Feel free to open a new issue if you run into problems.

PR #3129, opened 2024-01-18
Add support for Squirrel & Squat

EdMUK:
Adds support for Squirrel (#3127 ) a Lua like scripting language, along with Sqrat (#3128), which is a binding library.

### DIFF
--- a/packages/s/sqrat/xmake.lua
+++ b/packages/s/sqrat/xmake.lua
@@ -1,0 +1,17 @@
+    package("sqrat")
+    set_kind("library", {headeronly = true})
+    set_homepage("http://scrat.sourceforge.net/")
+    set_description("Sqrat is a C++ library for Squirrel that facilitates exposing classes and other native functionality to Squirrel scripts.")
+    set_license("zlib")
+    add_urls("git://git.code.sf.net/p/scrat/code")
+    add_deps("squirrel")
+
+    add_includedirs("include", "include/sqrat")
+    on_install(function (package)
+        os.cp("include/*.h", package:installdir("include"))
+        os.cp("include/sqrat/*.h", package:installdir("include/sqrat/"))
+    end)
+
+    on_test(function (package)
+        assert(package:has_cxxtypes("Sqrat::Function", {includedirs = "include", includes = "sqrat.h"}))
+    end)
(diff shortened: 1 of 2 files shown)


packages/s/squirrel/xmake.lua
```diff
@@ -0,0 +1,33 @@
...
+    add_urls("https://github.com/albertodemichelis/squirrel/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/albertodemichelis/squirrel.git")
+
+    add_versions("v3.2", "02805414cfadd5bbb921891d3599b83375a40650abd6404a8ab407dc5e86a996")
+
+    add_deps("cmake")
+
+    on_install(function (package)
+        if package:is_plat("iphoneos") then
+            package:build_addenv("defines", "IOS")
```
waruqi, line 16: ? defines is not environment vars

SirLynix: It's used by cmake tools:
https://github.com/xmake-io/xmake/blob/master/xmake/modules/package/tools/cmake.lua#L176-L185

it seemed like the easiest way to pass a define to CMake (this libs expects `IOS` define on iOS)

waruqi: we can pass cxflags to tools.cmake.

```
cmake.install(package, configs, {cxflags = "-DIOS"})

File: packages/s/sqrat/xmake.lua [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+
+    add_includedirs("include", "include/sqrat")
+
+    on_install(function (package)
+        os.cp("include/*.h", package:installdir("include"))
+        os.cp("include/sqrat/*.h", package:installdir("include/sqrat/"))
+    end)
+
+    on_test(function (package)
+        assert(package:has_cxxtypes("Sqrat::Function", {includedirs = "include", includes = "sqrat.h"}))
```
waruqi, line 19: please use check_cxxsnippets to check links

packages/s/sqrat/xmake.lua
```diff
@@ -0,0 +1,20 @@
+package("sqrat")
+    set_kind("library", {headeronly = true})
+    set_homepage("http://scrat.sourceforge.net/")
+    set_description("Sqrat is a C++ library for Squirrel that facilitates exposing classes and other native functionality to Squirrel scripts.")
+    set_license("zlib")
+
+    add_urls("git://git.code.sf.net/p/scrat/code")
```
waruqi, line 7: please add `add_versions`

Issue #3139, opened 2024-01-21
Unable run sdl program with generated CMakeLists.txt


Latias94:
### Xmake Version

v2.8.6+HEAD.211710b67

### Operating System Version and Architecture

Windows 11

### Describe Bug

I can run program with `xmake && xmake r`, but cannot run in Clion with generated `CMakeLists.txt`(`xmake project -k cmakelists`)

## project
[test_sdl.zip](https://github.com/xmake-io/xmake-repo/files/14001625/test_sdl.zip)

### Expected Behavior

run 

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")

set_exceptions("cxx")

set_runtimes(is_mode("debug") and "MDd" or "MD")

add_requires("libsdl")

target("app")
    set_kind("binary")
    add_files("src/main.cpp")
    add_deps("engine")

target("engine")
    set_kind("static")
    add_files("src/engine.cpp")
    add_packages("libsdl", {public = true})

```

### Additional Information and Error Logs

```shell
"C:\Users\xxx\AppData\Local\Programs\CLion Nova\bin\cmake\win\x64\bin\cmake.exe" --build F:\SourceCodes\Cpp\test_flags\cmake-build-debug --target app -j 30
[1/4] Building CXX object CMakeFiles\app.dir\src\main.cpp.obj
[2/4] Building CXX object CMakeFiles\engine.dir\src\engine.cpp.obj
[3/4] Linking CXX static library F:\SourceCodes\Cpp\test_flags\build\windows\x64\release\engine.lib
[4/4] Linking CXX executable F:\SourceCodes\Cpp\test_flags\build\windows\x64\release\app.exe
FAILED: F:/SourceCodes/Cpp/test_flags/build/windows/x64/release/app.exe 
[...]

Issue #3082, opened 2024-01-11
libsdl shouldn't add -system:windows when using lib component


SirLynix:
### Xmake Version

2.8.5

### Operating System Version and Architecture

Windows 11

### Describe Bug

libsdl package shouldn't add `-subsystem:windows` or `-mwindows` flags when using only lib component (and I wonder if it should do it at all, maybe we should let the user add it).

compilation fails with a missing WinMain error

### Expected Behavior

that libsdl doesn't add -subsystem:windows flags when using lib component. And I think we should remove it completely because:
```lua
add_rules("mode.debug", "mode.release")

add_requires("libsdl", "libsdl_image") -- libsdl_image depends on libsdl

target("test_sdl")
    set_kind("binary")
    add_files("src/*.cpp")
    add_packages("libsdl", { components = "lib" })
    add_packages("libsdl_image") -- use main component of libsdl
```

### Project Configuration

xmake.lua:
```lua
add_rules("mode.debug", "mode.release")

add_requires("libsdl")

target("test_sdl")
    set_kind("binary")
    add_files("src/*.cpp")
    add_packages("libsdl", { components = "lib" })
```

src/main.cpp:
```cpp
#include <SDL2/SDL.h>
#include <iostream>

int main(int argc, char** argv)
{
    std::cout << "hello sdl!" << std::endl;
    return 0;
}
```

### Additional Information and Error Logs

```
[ 25%]: compiling.release src\main.cpp
"C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.38.33130\\bin\\HostX64\\x64\\cl.exe" -c -nologo -O2 -fp:fast /EHsc -external:W0 -external:IC:\Users\lynix\AppData\Local\.xmake\packages\l\libsdl\2.28.5\d8952366ead841df8ae47979da7077b7\include -external:W0 -external:IC:\Users\lynix\AppData\Local\.xmake\packages\l\libsdl\2.28.5\d8952366ead841df8ae47979da7077b7\include\SDL2 -DNDEBUG -Fobuild\.objs\test_sdl\windows\x64\release\src\main.cpp.obj src\main.cpp
[ 50%]: linking.release test_sdl.exe
[...]

PR #3142, opened 2024-01-22
remove subsystem

waruqi:
https://github.com/xmake-io/xmake-repo/issues/3082
https://github.com/xmake-io/xmake-repo/issues/3139


Latias94:
```shell
error: D:\a\xmake-repo\xmake-repo\packages\l\libsdl\xmake.lua:231: ...gramdir\core\sandbox\modules\import\core\tool\linker.lua:75: @programdir\modules\core\tools\link.lua:175: 
LINK : fatal error LNK1561: entry point must be defined
```

SirLynix:
What's the reason xmake keeps adding it even when "main" component isn't active?

waruqi:
> What's the reason xmake keeps adding it even when "main" component isn't active?

I don't know, but I have not time to debug it.

PR #3148, opened 2024-01-22
Add extsources for utf8cpp

ImperatorS79:


ImperatorS79:
I don't know why the mingw macos shared build failed (there is no log).

PR #3146, opened 2024-01-22
cartographer: remove boost & set ceres-solver version

Chi-EEE:
boost is only being used in `common/port.h`  
set ceres-solver version to 2.1.0  
Fix windows from boost:  
```
error: cartographer.lib(submap_3d.cc.obj) : error LNK2019: unresolved external symbol "public: static void __cdecl boost::iostreams::zlib_error::check(int)" (?check@zlib_error@iostreams@boost@@SAXH@Z) referenced in function "public: void __cdecl boost::iostreams::symmetric_filter<class boost::iostreams::detail::zlib_compressor_impl<class std::allocator<char> >,class std::allocator<char> >::close<class boost::iostreams::non_blocking_adapter<class boost::iostreams::detail::linked_streambuf<char,struct std::char_traits<char> > > >(class boost::iostreams::non_blocking_adapter<class boost::iostreams::detail::linked_streambuf<char,struct std::char_traits<char> > > &,int)" (??$close@V?$non_blocking_adapter@V?$linked_streambuf@DU?$char_traits@D@std@@@detail@iostreams@boost@@@iostreams@boost@@@?$symmetric_filter@V?$zlib_compressor_impl@V?$allocator@D@std@@@detail@iostreams@boost@@V?$allocator@D@std@@@iostreams@boost@@QEAAXAEAV?$non_blocking_adapter@V?$linked_streambuf@DU?$char_traits@D@std@@@detail@iostreams@boost@@@12@H@Z)
cartographer.lib(proto_stream.cc.obj) : error LNK2001: unresolved external symbol "public: static void __cdecl boost::iostreams::zlib_error::check(int)" (?check@zlib_error@iostreams@boost@@SAXH@Z)
cartographer.lib(probability_grid.cc.obj) : error LNK2001: unresolved external symbol "public: static void __cdecl boost::iostreams::zlib_error::check(int)" (?check@zlib_error@iostreams@boost@@SAXH@Z)
[...]

PR #3150, opened 2024-01-22
ceres-solver & suitesparse: add support for older version

Chi-EEE:
This is for cartographer: https://github.com/xmake-io/xmake-repo/pull/3146 (ceres-solver 2.1.0)


xq114:
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/v7.5.1/GraphBLAS/Source/Template/GB_atomics.h#L294 I think it's reasonable to just disable graphblas for windows x86 platform since a x64 win32 function is involved, so that suitesparse 7+ can also be used on x86. If the only goal is to use ceres-solver on x86 platform, maybe we do not need to disable ceres 2.2.0. I'll take a look on that

xq114:
https://github.com/xmake-io/xmake-repo/pull/3154

PR #3153, opened 2024-01-22
Add extsources to vulkan-validation-layers

ImperatorS79:


ImperatorS79:
The git repository of this packagessay that only the layer library and the layer manifest are installed by the package (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/BUILD.md#cmake-installed-files). Why does the xmake package check for header files then ?

SirLynix:
> The git repository of this packagessay that only the layer library and the layer manifest are installed by the package (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/BUILD.md#cmake-installed-files). Why does the xmake package check for header files then ?

where does this package check for includes?

ImperatorS79:
As far as I understand, there is a `on_test function` with:
```lua
assert(package:has_cxxfuncs("getLayerOption", {includes = "layers/vk_layer_config.h"}))
```

Maybe I misunderstood that line ? anyway I don't know why the build is failing on so many platform.

ImperatorS79:
There is a problemon Android. The url looks like this:
```
https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/download/$(version1)/android-binaries-$(version2).tar.gz
```

`version1` should be "sdk-versionNumber" or "vulkan-sdk-versionNumber", but `version2` should be only "versionNumber".

How can I do that with `add_url` ?

waruqi:
> There is a problemon Android. The url looks like this:
> 
> ```
> https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/download/$(version1)/android-binaries-$(version2).tar.gz
> ```
> 
> `version1` should be "sdk-versionNumber" or "vulkan-sdk-versionNumber", but `version2` should be only "versionNumber".
> 
> How can I do that with `add_url` ?
> 
> For example:
> 
> ```
> https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/download/vulkan-sdk-1.3.275.0/android-binaries-1.3.275.0.tar.gz
> ```
> 
> But
> 
> ```
> https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/download/sdk-1.3.261.1/android-binaries-sdk-1.3.261.1-android.tar.gz
> ```

`$(version1)/android-binaries-$(version2)` 

-> `$(version)` and return `version1 .. "android-binaries-" .. version2`

ImperatorS79:
It seems to be working now

Issue #3162, opened 2024-01-24
sockpp 安装出错


StarBeat:
### Xmake 版本

 v2.8.2+20230909

### 操作系统版本和架构

win 10 wsl1 4.4.0-19041-Microsoft

### 描述问题

 xrepo install  sockpp  安装 sockpp v0.8.1 出现  unmatched checksum! 错误，无法正常下载包

### 期待的结果

 xrepo install  sockpp 在 linux 下正常安装

### 工程配置

No Need

### 附加信息和错误日志

xmake f -c --require=n -vD
checking for platform ... linux
checking for architecture ... x86_64
checking for gcc ... /usr/bin/gcc
checkinfo: cannot runv(zig version), No such file or directory
checking for zig ... no
checkinfo: cannot runv(zig version), No such file or directory
checking for zig ... no
configure
{
    proxy_pac = pac.lua
    arch = x86_64
    theme = default
    network = public
    plat = linux
    proxy_hosts = github.com,gitlab.*,*.xmake.io
    proxy = http://127.0.0.1:1080
    clean = true
    buildir = build
    ccache = true
    mode = release
    host = linux
    ndk_stdcxx = true
    kind = static
}
xmake require -v -D -j 26 --extra={system=false} sockpp
checking for unzip ... /usr/bin/unzip
checking for git ... /usr/bin/git
checking for gzip ... /bin/gzip
checking for tar ... /bin/tar
/usr/bin/git rev-parse HEAD
checking for cmake ... no
checking for cmake ... /usr/bin/cmake
finding sockpp from xmake ..
checking for xmake::sockpp ... no
note: install or modify (m) these packages (pass -y to skip confirm)?
in myrepo:
  -> sockpp v0.8.1
please input: y (y/n/m)
y
checking for ping ... /bin/ping
pinging the host(github.com) ... 65535 ms
/bin/tar -xf v0.8.1.tar.gz -C source.tmp
patching /mnt/x/wsl/X/xmake-repo/packages/s/sockpp/patches/0.8.1/strerror.patch to sockpp-v0.8.1 ..
error: ...es/private/action/require/impl/actions/patch_sources.lua:96: patch(/mnt/x/wsl/X/xmake-repo/packages/s/sockpp/patches/0.8.1/strerror.patch): unmatched checksum!
stack traceback:
    [C]: in function 'error'
    [@programdir/core/base/os.lua:921]:
[...]

PR #3163, opened 2024-01-24
Fix sockpp patch checksum

SirLynix:
See #3162


SirLynix:
See https://github.com/xmake-io/xmake/pull/4649

PR #3165, opened 2024-01-24
add GLPK package

jube:
This PR adds the GLPK package. "The GLPK (GNU Linear Programming Kit) package is intended for solving large-scale linear programming (LP), mixed integer programming (MIP), and other related problems."

### DIFF
--- a/packages/g/glpk/xmake.lua
+++ b/packages/g/glpk/xmake.lua
@@ -7,6 +7,8 @@ package("glpk")
 
     add_versions("5.0", "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15")
 
+    add_configs("shared", {description = "Build shared binaries", default = false, type = "boolean", readonly = true})
+
     if is_plat("linux") then
         add_extsources("apt::libglpk-dev")
     end
@@ -17,6 +19,22 @@ package("glpk")
 
     on_install("macosx", "linux", function (package)
         import("package.tools.autoconf").install(package)
+    end)
+
+    on_install("windows", function (package)
+        os.cd(is_arch("x64", "x86_64") and "w64" or "w32")
+        os.cp("config_VC", "config.h")
+        import("package.tools.nmake").build(package, {"/f", package:config("shared") and "makefile_VC_DLL" or "makefile_VC"})
+
+        if package:config("shared") then
+            os.cp("glpk_5_0.dll", package:installdir("bin"))
+            os.cp("glpk_5_0.lib", package:installdir("lib"))
+        else
+            os.cp("glpk.lib", package:installdir("lib"))
+        end
+
+        os.cd("..")
+        os.cp("src/glpk.h", package:installdir("include"))
     end)
 
     on_test(function (package)
--- a/packages/g/glpk/xmake.lua
+++ b/packages/g/glpk/xmake.lua
@@ -7,7 +7,7 @@ package("glpk")
 
     add_versions("5.0", "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15")
 
-    add_configs("shared", {description = "Build shared binaries", default = false, type = "boolean", readonly = true})
+    add_configs("shared", {description = "Build shared binaries", default = false, type = "boolean"})
 
     if is_plat("linux") then
         add_extsources("apt::libglpk-dev")
--- a/packages/g/glpk/xmake.lua
+++ b/packages/g/glpk/xmake.lua
@@ -24,6 +24,7 @@ package("glpk")
     on_install("windows", function (package)
         os.cd(is_arch("x64", "x86_64") and "w64" or "w32")
         os.cp("config_VC", "config.h")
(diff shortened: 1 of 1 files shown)


packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+
+    on_install("macosx", "linux", function (package)
+        import("package.tools.autoconf").install(package)
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            #include <glpk.h>
+            int main() {
+              return 0;
```
star-hengxing, line 24: Please add some api call to link.

packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("glpk")
+    set_homepage("https://www.gnu.org/software/glpk/")
+    set_description("The GLPK (GNU Linear Programming Kit) package is intended for solving large-scale linear programming (LP), mixed integer programming (MIP), and other related problems.")
+    set_license("GPL-3.0")
+
+    set_urls("https://ftp.gnu.org/gnu/glpk/glpk-$(version).tar.gz")
+
+    add_versions("5.0", "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15")
+
+    add_extsources("apt::libglpk-dev")
```
star-hengxing, line 10:
```lua
if is_plat("linux") then
    add_extsources("apt::libglpk-dev")
end
```

packages/g/glpk/xmake.lua
jube: Done

packages/g/glpk/xmake.lua
jube: Done

jube:
I've added support for building the library for Windows.
Should I remove the macOS support as it does not build on this platform?

packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+    add_versions("5.0", "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15")
+
+    add_configs("shared", {description = "Build shared binaries", default = false, type = "boolean"})
+
+    if is_plat("linux") then
+        add_extsources("apt::libglpk-dev")
+    end
+
+    if is_plat("macosx|x86_64", "linux") then
+        add_deps("autoconf", "automake", "libtool", "pkg-config")
```
waruqi, line 17: if ./configure file exists, we need not add these deps. they are only for ./autogen.sh to generate configure.

packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,54 @@
+package("glpk")
+    set_homepage("https://www.gnu.org/software/glpk/")
+    set_description("The GLPK (GNU Linear Programming Kit) package is intended for solving large-scale linear programming (LP), mixed integer programming (MIP), and other related problems.")
+    set_license("GPL-3.0")
+
+    set_urls("https://ftp.gnu.org/gnu/glpk/glpk-$(version).tar.gz")
+
+    add_versions("5.0", "4a1013eebb50f728fc601bdd833b0b2870333c3b3e5a816eeba921d95bec6f15")
+
+    add_configs("shared", {description = "Build shared binaries", default = false, type = "boolean"})
```
waruqi, line 10: ? if this library only support static library, we should set `readonly = true`, otherwise please remove it.

packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+    if is_plat("linux") then
+        add_extsources("apt::libglpk-dev")
+    end
+
+    if is_plat("macosx|x86_64", "linux") then
+        add_deps("autoconf", "automake", "libtool", "pkg-config")
+    end
+
+    on_install("macosx|x86_64", "linux", function (package)
+        import("package.tools.autoconf").install(package)
```
waruqi, line 21: please add static/shared configs

packages/g/glpk/xmake.lua
```diff
@@ -0,0 +1,54 @@
...
+    end)
+
+    on_install("windows", function (package)
+        os.cd(is_arch("x64", "x86_64") and "w64" or "w32")
```
waruqi, line 25: package:is_arch

packages/g/glpk/xmake.lua
jube: The configure script provides both `--enable-shared` and `--enable-static` and both are set to yes. Should I disable one of these two according to static/shared configs?

waruqi:
```lua
        local configs = {}
        table.insert(configs, "--enable-shared=" .. (package:config("shared") and "yes" or "no"))
        table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes"))
        import("package.tools.autoconf").install(package, configs)
```

packages/g/glpk/xmake.lua
jube: OK, I also removed this line

PR #3174, opened 2024-01-25
add gspell package

paul-reilly:
This was needed for wxwidgets non-system package to build on Linux Manjaro.

It's not a fully featured package, but it's in the same style as the gtk+3 package:

https://github.com/xmake-io/xmake-repo/blob/67a84e98bd284cc908c561fd4af77bf40be09161/packages/g/gtk%2B3/xmake.lua#L1-L13

### DIFF
--- a/packages/g/gspell/xmake.lua
+++ b/packages/g/gspell/xmake.lua
@@ -1,0 +1,12 @@
+package("gspell")
+
+    set_homepage("https://gitlab.gnome.org/GNOME/gspell")
+    set_description("A spell-checking library for GTK applications")
+    set_license("LGPL-2.0-or-later")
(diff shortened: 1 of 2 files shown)


packages/g/gspell/xmake.lua
```diff
@@ -0,0 +1,14 @@
+package("gspell")
+
+    set_homepage("https://gitlab.gnome.org/GNOME/gspell")
+    set_description("A spell-checking library for GTK applications")
+    set_license("LGPL-2.0-or-later")
+
+    if on_fetch then
```
waruqi, line 7: please remove it, recent versions always support on_fetch

PR #3183, opened 2024-01-26
Upgrade elfutils

zebede22:
I don't know where to get these patches from

### DIFF
--- a/packages/e/elfutils/patches/0.190/configure.patch
+++ b/packages/e/elfutils/patches/0.190/configure.patch
@@ -1,7 +1,7 @@
 diff --git a/configure b/configure
 index acc483f..4a3e0a4 100755
---- a/elfutils-0.189/configure
-+++ b/elfutils-0.190/configure
+--- a/configure
++++ b/configure
 @@ -1,6 +1,6 @@
  #! /bin/sh
  # Guess values for system-dependent variables and create Makefiles.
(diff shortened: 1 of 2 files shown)


waruqi:
Modify the code yourself and generate the patch via git diff > patch.txt

zebede22:
done

zebede22:
anything needs to be done? @waruqi

waruqi:
you need fix all errors on ci

waruqi:
> seems like tests doesn't like my patch, can you tell me what should I change? I tried several times :/ @waruqi
> 
> I still don't understand why we need that patches? (git diff 0.189/configure 0.190/configure > configure.patch)

please use `git diff` to generate patch file instead of `patch` command, because xmake use `git apply` to apply patch.

```console
cd codedir
git init
git add .
git commit -a -m "..."
-- change some codes
git diff > mypatch.txt
xmake l hash.sha256 ./mypatch.txt
```

zebede22:
> ```
> -- change some codes
> ```

That's what I dont understand, "-- change some codes", what code? why?

waruqi:
> > Bot detected the issue body's language is not English, translate it automatically.
> 
> > ```
> > -- change some codes
> > ```
> 
> That's what I don't understand, "-- change some codes", what code? why?

Your patch code

waruqi:
> seems like tests doesn't like my patch, can you tell me what should I change? I tried several times :/ @waruqi
> 
> I still don't understand why we need that patches? (git diff 0.189/configure 0.190/configure > configure.patch)

That's a patch for 0.189, that doesn't mean, 0.190 needs them too, you need to change the code based on the actual errors in the ci that need to be changed to fix all the compilation errors and then make them into patches.

PR #3184, opened 2024-01-26
ormpp: fix config

KkemChen:


xq114 (commented):
`sqlite = "ORMPP_ENABLE_SQLITE3"` sqlite3 here

PR #3222, opened 2024-02-02
Update glfw

nofuncoding:
Updated glfw to 3.3.9


SirLynix:
Do we have the guarantee that `package:runtimes()` returns a single value on Windows?

waruqi:
> Do we have the guarantee that `package:runtimes()` returns a single value on Windows?

sure, but package:runtimes only exists on dev branch now. please use vs_runtime now.

and we should use package:has_runtime later.

PR #3244, opened 2024-02-05
Improve OpenBLAS

samayala22:
Reworked the package for more granular control over the compilation of the library.

- Switch to cmake build.
- Removed fortran option. OpenBLAS has the option to compile LAPACK from C sources and I believe this should be the default.
- Added dynamic arch compilation option using the default target preset provided by OpenBLAS.
- Simplified how openmp is added.

### DIFF
--- a/packages/o/openblas/xmake.lua
+++ b/packages/o/openblas/xmake.lua
@@ -45,8 +45,9 @@ package("openblas")
 
     add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = is_plat("windows")})
     add_configs("lapack", {description = "Build LAPACK", default = not is_plat("android"), type = "boolean", readonly = (is_plat("windows") or is_plat("android"))})
-    add_configs("dynamic-arch", {description = "Enable dynamic arch dispatch", default = (is_plat("linux") or is_plat("windows")), type = "boolean", readonly = not is_plat("linux")})
+    add_configs("dynamic_arch", {description = "Enable dynamic arch dispatch", default = (is_plat("linux") or is_plat("windows")), type = "boolean", readonly = not is_plat("linux")})
     add_configs("openmp", {description = "Compile with OpenMP enabled.", default = (is_plat("windows") or is_plat("linux")), type = "boolean", readonly = not is_plat("linux")})
+    add_configs("fortran", {description = "Compile with Fortran enabled.", default = false, type = "boolean", readonly = (is_plat("windows") or is_plat("android"))})
     
     if not is_plat("windows") then
         add_deps("cmake")
@@ -69,21 +70,23 @@ package("openblas")
     on_install("windows|x64", "windows|x86", function (package)
         os.cp("bin", package:installdir())
         os.cp("include", package:installdir())
-        os.cp("lib", package:installdir())
+        os.cp(path.join("lib", "libopenblas.lib"), path.join(package:installdir("lib"), "openblas.lib"))
         package:addenv("PATH", "bin")
     end)
 
-    on_install("macosx", "linux", "android", function (package)
-        local configs = {"-DCMAKE_BUILD_TYPE=Release", "-DBUILD_TESTING=OFF", "-DNOFORTRAN=ON"}
+    on_install("macosx", "linux", "android", "mingw", function (package)
+        local configs = {"-DCMAKE_BUILD_TYPE=Release", "-DBUILD_TESTING=OFF"}
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
-        table.insert(configs, "-DDYNAMIC_ARCH=" .. (package:config("dynamic-arch") and "ON" or "OFF"))
+        table.insert(configs, "-DDYNAMIC_ARCH=" .. (package:config("dynamic_arch") and "ON" or "OFF"))
         table.insert(configs, "-DUSE_OPENMP=" .. (package:config("openmp") and "ON" or "OFF"))
+        table.insert(configs, "-DNOFORTRAN=" .. (package:config("fortran") and "OFF" or "ON"))
         if package:is_plat("macosx") and package:is_arch("arm64") then
             table.insert(configs, "-DTARGET=VORTEX") 
             table.insert(configs, "-DBINARY=64")
         end
         if package:config("lapack") then
-            table.join2(configs, {"-DC_LAPACK=ON", "-DBUILD_LAPACK_DEPRECATED=OFF"})
+            if not package:config("fortran") then table.insert(configs, "-DC_LAPACK=ON") end
+            table.insert(configs, "-DBUILD_LAPACK_DEPRECATED=OFF")
         else
             table.insert(configs, "-DONLY_CBLAS=ON")
             table.insert(configs, "-DBUILD_WITHOUT_LAPACK=ON")
--- a/packages/o/openblas/xmake.lua
+++ b/packages/o/openblas/xmake.lua
@@ -48,12 +48,15 @@ package("openblas")
     add_configs("lapack", {description = "Build LAPACK", default = true, type = "boolean", readonly = is_plat("windows")})
     add_configs("dynamic-arch", {description = "Enable dynamic arch dispatch", default = true, type = "boolean", readonly = is_plat("windows")})
     add_configs("openmp",  {description = "Compile with OpenMP enabled.", default = not is_plat("macosx"), type = "boolean", readonly = is_plat("windows")})
-
+    
+    if not is_plat("windows") then
+        add_deps("cmake")
+    end
     if is_plat("linux") then
         add_extsources("apt::libopenblas-dev", "pacman::libopenblas") -- remove ?
         add_syslinks("pthread")
     elseif is_plat("macosx") then
-        add_frameworks("Accelerate")
+        add_frameworks("Accelerate") -- remove ?
     end
 
     on_load("macosx", "linux", function (package)
--- a/packages/o/openblas/xmake.lua
+++ b/packages/o/openblas/xmake.lua
@@ -45,7 +45,7 @@ package("openblas")
 
     add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = is_plat("windows")})
     add_configs("lapack", {description = "Build LAPACK", default = not is_plat("android"), type = "boolean", readonly = (is_plat("windows") or is_plat("android"))})
-    add_configs("dynamic_arch", {description = "Enable dynamic arch dispatch", default = (is_plat("linux") or is_plat("windows")), type = "boolean", readonly = not is_plat("linux")})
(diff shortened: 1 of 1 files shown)


samayala22:
I'm investigating build issues.

samayala22:
@xq114 need review

I want your opinion on removing versions below 0.3.21. I'm not going to test every version but I'm pretty sure that cmake support is broken on older versions.

xq114:
It's possible to keep both build systems, see https://github.com/xmake-io/xmake-repo/blob/dev/packages/t/tbb/xmake.lua

samayala22:
> It's possible to keep both build systems, see https://github.com/xmake-io/xmake-repo/blob/dev/packages/t/tbb/xmake.lua

Yeah but it's gonna become a mess. And it's not just about the build system, the dynamic arch thing is extremely messy and probably broken on older versions. Do we REALLY need the old versions.

packages/o/openblas/xmake.lua
```diff
@@ -1,9 +1,9 @@
 package("openblas")
-
     set_homepage("http://www.openblas.net/")
     set_description("OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version.")
     set_license("BSD-3-Clause")
 
+    -- Versions
```
waruqi, line 6: remove comments

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
         add_versions("0.3.23", "5d9491d07168a5d00116cdc068a40022c3455bf9293c7cb86a65b1054d7e5114")
         add_versions("0.3.24", "ceadc5065da97bd92404cac7254da66cc6eb192679cf1002098688978d4d5132")
         add_versions("0.3.26", "4e6e4f5cb14c209262e33e6816d70221a2fe49eb69eaf0a06f065598ac602c68")
-
-        add_configs("fortran", {description = "Compile with fortran enabled.", default = is_plat("linux"), type = "boolean"})
-        add_configs("openmp",  {description = "Compile with OpenMP enabled.", default = not is_plat("macosx"), type = "boolean"})
     end
 
+    -- Configs
```
waruqi, line 47: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
+        readonly = not is_plat("linux")
+    })
+    add_configs("openmp",  {
+        description = "Compile with OpenMP enabled.",
+        default = (is_plat("windows") or is_plat("linux")),
+        type = "boolean",
+        readonly = not is_plat("linux")
+    })
+    
+    -- Dependencies
```
waruqi, line 68: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
+    if not is_plat("windows") then
+        add_deps("cmake")
+    end
     if is_plat("linux") then
-        add_extsources("apt::libopenblas-dev", "pacman::libopenblas")
+        add_extsources("apt::libopenblas-dev", "pacman::libopenblas") -- remove ?
```
waruqi, line 73: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
-    on_install("macosx", "linux", "mingw@windows,msys", function (package)
-        import("lib.detect.find_tool")
-        import("package.tools.make")
-        local configs = {}
-        if package:is_plat("linux") then
-            table.insert(configs, "CC=" .. package:build_getenv("cc"))
+    on_install("macosx", "linux", "android", function (package)
+        local configs = {
+            "-DCMAKE_BUILD_TYPE=Release", -- needed for linux clang https://github.com/OpenMathLib/OpenBLAS/issues/2634
+            "-DBUILD_TESTING=OFF", -- force no lapack test suite
```
waruqi, line 98: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
-            table.insert(configs, "TARGET=VORTEX")
-            table.insert(configs, "BINARY=64")
-            table.insert(configs, "CFLAGS=-arch arm64")
-            table.insert(configs, "LDFLAGS=-arch arm64")
-        end
-        if package:debug() then table.insert(configs, "DEBUG=1") end
-        if package:config("openmp") then table.insert(configs, "USE_OPENMP=1") end
-        if not package:config("shared") then
-            table.insert(configs, "NO_SHARED=1")
+        if (package:config("lapack")) then
```
waruqi, line 108: remove `()`

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
-            if ldflags then
-                table.insert(configs, "LDFLAGS=" .. ldflags)
-            end
-        end
-        make.build(package, configs)
-        make.make(package, table.join("install", "PREFIX=" .. package:installdir():gsub("\\", "/"), configs))
+        import("package.tools.cmake").build(package, configs, {buildir = "build"})
+        import("package.tools.cmake").install(package, configs, {buildir = "build"})
+
+        -- better way to do this ?
```
waruqi, line 117: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
+        os.mv(package:installdir() .. "/include/openblas/*" , package:installdir("include"))
+        os.rm(package:installdir("include/openblas"))
     end)
 
+    -- Tests
```
waruqi, line 122: here

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
                 double B[6] = {1.0,2.0,1.0,-3.0,4.0,-1.0};
-                double C[9] = {.5,.5,.5,.5,.5,.5,.5,.5,.5};
+                double C[9] = {0.8,0.2,0.5,-0.3,0.5,0.2,0.1,0.4,0.1};
+
                 cblas_dgemm(CblasColMajor,CblasNoTrans,CblasTrans,3,3,2,1,A,3,B,3,2,C,3);
             }
-        ]]}, {includes = "cblas.h"}))
+        ]]}, {includes = {"cblas.h"}}))
+
+        if (package:config("lapack")) then
```
waruqi, line 134: remove `()`

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
         add_versions("0.3.24", "ceadc5065da97bd92404cac7254da66cc6eb192679cf1002098688978d4d5132")
         add_versions("0.3.26", "4e6e4f5cb14c209262e33e6816d70221a2fe49eb69eaf0a06f065598ac602c68")
-
-        add_configs("fortran", {description = "Compile with fortran enabled.", default = is_plat("linux"), type = "boolean"})
-        add_configs("openmp",  {description = "Compile with OpenMP enabled.", default = not is_plat("macosx"), type = "boolean"})
     end
 
+    -- Configs
+    add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = is_plat("windows")})
+    add_configs("lapack", {
```
waruqi, line 49: please use single line for each add_configs

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +42,104 @@ package("openblas")
...
         package:addenv("PATH", "bin")
     end)
 
-    on_install("macosx", "linux", "mingw@windows,msys", function (package)
-        import("lib.detect.find_tool")
-        import("package.tools.make")
-        local configs = {}
-        if package:is_plat("linux") then
-            table.insert(configs, "CC=" .. package:build_getenv("cc"))
+    on_install("macosx", "linux", "android", function (package)
```
xq114, line 95: mingw should be supported too

packages/o/openblas/xmake.lua
```diff
@@ -29,8 +29,6 @@ package("openblas")
             add_versions("0.3.24", "92f8e0c73e1eec3c428b210fbd69b91e966f8cf1f998f3b60a52f024b2bf9d27")
             add_versions("0.3.26", "9c3d48c3c21cd2341d642a63ee8a655205587befdab46462df7e0104d6771f67")
         end
-
-        add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = true})
```
xq114, line 33: this config should be kept on windows to remind xmake that it contains only shared binaries

xq114:
C-API LAPACKE is not identical to LAPACK in symbols. Packages may still depend to LAPACK functions in openblas. I think the fortran option should be kept or two build systems should come together to make it work. If it's hard for cmake, at lease we should add an option to switch between make and cmake.

xq114 (changes requested):

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +41,80 @@ package("openblas")
...
         add_versions("0.3.24", "ceadc5065da97bd92404cac7254da66cc6eb192679cf1002098688978d4d5132")
         add_versions("0.3.26", "4e6e4f5cb14c209262e33e6816d70221a2fe49eb69eaf0a06f065598ac602c68")
-
-        add_configs("fortran", {description = "Compile with fortran enabled.", default = is_plat("linux"), type = "boolean"})
-        add_configs("openmp",  {description = "Compile with OpenMP enabled.", default = not is_plat("macosx"), type = "boolean"})
     end
 
+    add_configs("shared", {description = "Build shared library.", default = true, type = "boolean", readonly = is_plat("windows")})
+    add_configs("lapack", {description = "Build LAPACK", default = not is_plat("android"), type = "boolean", readonly = (is_plat("windows") or is_plat("android"))})
+    add_configs("dynamic-arch", {description = "Enable dynamic arch dispatch", default = (is_plat("linux") or is_plat("windows")), type = "boolean", readonly = not is_plat("linux")})
```
waruqi, line 48: please do not use `-` in config name, you can use `_` instead of it.

it will simplify configs.

`add_requires(xxx, configs = {["dynamic-arch"] = "xx"})`

-> 

`add_requires(xxx, configs = {dynamic_arch = "xx"})`

packages/o/openblas/xmake.lua
```diff
@@ -44,116 +41,83 @@ package("openblas")
...
-                io.replace("Makefile.system", "-fopenmp", cflags, {plain = true})
-            end
-            if ldflags then
-                table.insert(configs, "LDFLAGS=" .. ldflags)
-            end
-        end
-        make.build(package, configs)
-        make.make(package, table.join("install", "PREFIX=" .. package:installdir():gsub("\\", "/"), configs))
+        import("package.tools.cmake").build(package, configs, {buildir = "build"})
+        import("package.tools.cmake").install(package, configs, {buildir = "build"})
```
waruqi, line 95: we need only call cmake.install, please remove cmake.build

waruqi: build dir is unused.

PR #3258, opened 2024-02-07
Fix bug on zlib: global 'check_cincludes' is not callable

KnightJun:
编译zlib的时候会出现global 'check_cincludes' is not callable


SirLynix:
That's actually not a bug, `includes("check_cincludes.lua")` was replaced by `includes("@builtin/check")` in [xmake 2.7.7](https://tboox.org/2023/02/23/xmake-update-v2.7.7/) that got out about a year ago.

you should update your xmake version.

waruqi:
please update your xmake.

PR #3269, opened 2024-02-09
libsdl: Fix compilation on mingw-w64

SirLynix:
See https://github.com/libsdl-org/SDL/issues/9031


Biswa96:
The fix is required iff you are using latest mingw-w64 from git repository.

SirLynix:
I had it fail compilation on my repository and I don't compile mingw-w64 myself

PR #3273, opened 2024-02-10
update pciutils, pixman and x11 libs

xq114:


xq114:
macosx上libx11需要运行片段检测所以不支持交叉编译，见 https://github.com/xmake-io/xmake/issues/4657

PR #3310, opened 2024-02-20
Add msys2

waruqi:

------

### DIFF
--- a/packages/m/msys2/xmake.lua
+++ b/packages/m/msys2/xmake.lua
@@ -1,0 +1,15 @@
+package("msys2")
+    set_kind("toolchain")
+    set_homepage("https://www.msys2.org/")
+    set_description("Software Distribution and Building Platform for Windows")
+
+    add_urls("https://github.com/msys2/msys2-installer/releases/download/$(version).tar.xz", {version = function (version)
+            return version:gsub("%.", "-")  .. "/msys2-base-x86_64-" .. version:gsub("%.", "")
+        end})
+    add_versions("2024.01.13", "04456a44a956d3c0b5f9b6c754918bf3a8c3d87c858be7a0c94c9171ab13c58c")
+
+    on_install("@windows", function (package)
+    end)
+
+    on_test(function (package)
+    end)
--- a/packages/m/msys2/xmake.lua
+++ b/packages/m/msys2/xmake.lua
@@ -9,6 +9,7 @@ package("msys2")
     add_versions("2024.01.13", "04456a44a956d3c0b5f9b6c754918bf3a8c3d87c858be7a0c94c9171ab13c58c")
 
     on_install("@windows", function (package)
+        print(os.files("**"))
     end)
 
     on_test(function (package)


SirLynix:
Windows has a bash.exe in C:/Windows/System32/bash.exe
```
bash --version
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
```

waruqi:
msys2 bin has been first path in $PATH

> C:\Users\runneradmin\AppData\Local\.xmake\packages\m\msys2\2024.01.13\a77fa58bdf6d49bbbfa0c92a5cf82862\bin;C:\Users\runneradmin\AppData\Local\.xmake\packages\m\msys2\2024.01.13\a77fa58bdf6d49bbbfa0c92a5cf82862\usr\bin;D:\a\xmake-repo\xmake-repo\xmakesrc\xmake;C:\Program Files\PowerShell\7;C:\hostedtoolcache\windows\xmake\master\x64;C:\Program Files\MongoDB\Server\5.0\bin;C:\aliyun-cli;C:\vcpkg;C:\cf-cli;C:\Program Files (x86)\NSIS\;C:\tools\zstd;C:\Program Files\Mercurial\;C:\hostedtoolcache\windows\stack\2.15.1\x64;C:\cabal\bin;C:\\ghcup\bin;C:\mingw64\bin;C:\Program Files\dotnet;C:\Program Files\MySQL\MySQL Server 5.7\bin;C:\Program Files\R\R-4.3.2\bin\x64;C:\SeleniumWebDrivers\GeckoDriver;C:\SeleniumWebDrivers\EdgeDriver\;C:\SeleniumWebDrivers\ChromeDriver;C:\Program Files (x86)\sbt\bin;C:\Program Files (x86)\GitHub CLI;C:\Program Files\Git\bin;C:\Program Files (x86)\pipx_bin;C:\npm\prefix;C:\hostedtoolcache\windows\go\1.20.14\x64\bin;C:\hostedtoolcache\windows\Python\3.7.9\x64\Scripts;C:\hostedtoolcache\windows\Python\3.7.9\x64;C:\hostedtoolcache\windows\Ruby\2.5.9\x64\bin;C:\Program Files\OpenSSL\bin;C:\tools\kotlinc\bin;C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.402-6\x64\bin;C:\Program Files\ImageMagick-7.1.1-Q16-HDRI;C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin;C:\ProgramData\kind;C:\Program Files\Eclipse Foundation\jdk-8.0.302.8-

SirLynix:
> msys2 bin has been first path in $PATH

Then I don't know why bash.exe in System32 is executed instead 😕

xq114:
Should msystem config be separated into another package? This config does not change the msys2 installation. i.e. 
```
package("msys2")
-- plain install

package("msys2-env")
add_deps("msys2")
-- add_configs("msystem", ...)
-- add environment variables
```

msys2 itself can be updated via pacman, so I think it's much better keeping only one installation on the system (better fix it's shared/runtime settings to ensure this!)

waruqi:
> > msys2 bin has been first path in $PATH
> 
> Then I don't know why bash.exe in System32 is executed instead 😕

Because system catalogue searches have a higher priority than %PATH%

> If the file name does not contain a directory path, the system searches for the executable file in the following sequence:
> 
> The directory from which the application loaded.
> The current directory for the parent process.
> The 32-bit Windows system directory. Use the [GetSystemDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) function to get the path of this directory.
> The 16-bit Windows system directory. There is no function that obtains the path of this directory, but it is searched. The name of this directory is System.
> The Windows directory. Use the [GetWindowsDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) function to get the path of this directory.
> The directories that are listed in the PATH environment variable. Note that this function does not search the per-application path specified by the App Paths registry key. To include this per-application path in the search sequence, use the [ShellExecute](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/axe/shellexecute) function.

waruqi:
> Should msystem config be separated into another package? This config does not change the msys2 installation. i.e.
> 
> ```
> package("msys2")
> -- plain install
> 
> package("msys2-env")
> add_deps("msys2")
> -- add_configs("msystem", ...)
> -- add environment variables
> ```
> 

No, there's no need to make it too complicated.

> msys2 itself can be updated via pacman, so I think it's much better keeping only one installation on the system (better fix it's shared/runtime settings to ensure this!)

The msys2 package should use xrepo/xmake to update its version, not pacman to update itself. If you want to update the msys2 version, use add_versions to add it.

and toolchain/binary kind doesn't care about runtime, it always has a same buildhash.

xq114:
> No, there's no need to make it too complicated.
> 
> > msys2 itself can be updated via pacman, so I think it's much better keeping only one installation on the system (better fix it's shared/runtime settings to ensure this!)
> 
> The msys2 package should use xrepo/xmake to update its version, not pacman to update itself. If you want to update the msys2 version, use add_versions to add it.
> 
> and toolchain/binary kind doesn't care about runtime, it always has a same buildhash.

But configs still change buildhash, then I have to install msys2 multiple times if I want to use different msystem settings.

And only msys2 release can use xrepo/xmake to update its version, packages inside msys2 can only get updated with pacman.

waruqi:
> > No, there's no need to make it too complicated.
> > > msys2 itself can be updated via pacman, so I think it's much better keeping only one installation on the system (better fix it's shared/runtime settings to ensure this!)
> > 
> > 
> > The msys2 package should use xrepo/xmake to update its version, not pacman to update itself. If you want to update the msys2 version, use add_versions to add it.
> > and toolchain/binary kind doesn't care about runtime, it always has a same buildhash.
> 
> But configs still change buildhash, then I have to install msys2 multiple times if I want to use different msystem settings.
> 

As you said, we can continue to add some separate packages. e.g. `msys-mingw64`, `msys-clang64`, ...

> And only msys2 release can use xrepo/xmake to update its version, packages inside msys2 can only get updated with pacman.

add `add_configs("update", ...)` to update packages if user always want to update them. then we can call pacman to update them in on_install.

and we can also enable it by default.

like this https://github.com/msys2/setup-msys2/blob/ae10b6876d23bb08bd632be2951125be2e32793b/main.js#L399

xq114:
> add `add_configs("update", ...)` to update packages if user always want to update them. then we can call pacman to update them in on_install.
> 
> and we can also enable it by default.
> 
> like this [msys2/setup-msys2@`ae10b68`/main.js#L399](https://github.com/msys2/setup-msys2/blob/ae10b6876d23bb08bd632be2951125be2e32793b/main.js#L399)

It has nothing to do with config, I mean, the update of msys2 packages are not syncronized with msys2 itself. For example, I have installed msys2 20240113 in xrepo, and on 2024/03/13 some packages (e.g. gcc) got updated, I can activate the msys2 environment and run pacman to install the latest gcc. It's just like xmake and xrepo, I do not have to update xmake to use the latest xrepo

waruqi:
> > add  to update packages if user always want to update them. then we can call pacman to update them in on_install.`add_configs("update", ...)`
> > and we can also enable it by default.
> > like this [msys2/setup-msys2@`ae10b68`/main.js#L399](https://github.com/msys2/setup-msys2/blob/ae10b6876d23bb08bd632be2951125be2e32793b/main.js#L399)
> 
> It has nothing to do with config, I mean, the update of msys2 packages are not syncronized with msys2 itself. For example, I have installed msys2 20240113 in xrepo, and on 2024/03/13 some packages (e.g. gcc) got updated, I can activate the msys2 environment and run pacman to install the latest gcc. It's just like xmake and xrepo, I do not have to update xmake to use the latest xrepo

so we need only run `xrepo env shell` to activate msys2 env, and run pacman to update or install gcc. 
There doesn't seem to be anything that needs to be done in the package configuration.

> 
> > As you said, we can continue to add some separate packages. e.g. , , ...`msys2-mingw64``msys2-clang64`
> 
> That's why various msystem settings should share the same msys2 installation. Currently if I change the msystem config, I have to reinstall msys2 and re-run pacman to get my desired versions of packages. I think at least there should be a "empty" value for msystem and let user add environments by theirselves

msystem config default value is nil now.

waruqi:
see https://github.com/xmake-io/xmake-repo/pull/3313

PR #3323, opened 2024-02-23
reproc: add multithreaded option and fix link orders

paul-reilly:
Just the link order and multithreaded fixes for now. I've opened an issue asking for a point release in the reproc repo https://github.com/DaanDeMeyer/reproc/issues/109 to fix building with GCC 13.

### DIFF
--- a/packages/r/reproc/patches/0001-to-head.diff
+++ b/packages/r/reproc/patches/0001-to-head.diff
@@ -1,0 +1,289 @@
+diff --git a/.clang-tidy b/.clang-tidy
+index 35bb58f..de92997 100644
+--- a/.clang-tidy
++++ b/.clang-tidy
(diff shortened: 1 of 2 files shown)


paul-reilly:
Just Fedora and Arch failing now which is the GCC 13 issue. Users can use the head of the main branch like:

```lua
add_requires("reproc main")
```

... which includes the GCC 13 fix from Apr 21st 2023 which hasn't been included in a versioned release yet.

waruqi:
> Just Fedora and Arch failing now which is the GCC 13 issue. Users can use the head of the main branch like:
> 
> ```lua
> add_requires("reproc main")
> ```
> 
> ... which includes the GCC 13 fix from Apr 21st 2023 which hasn't been included in a versioned release yet.

please add patch to fix it.

PR #3335, opened 2024-02-25
feat: add duckdb

c8ef:

------


packages/d/duckdb/xmake.lua
```diff
@@ -0,0 +1,32 @@
+package("duckdb")
+    set_homepage("http://duckdb.org/")
+    set_description("DuckDB is an in-process SQL OLAP Database Management System")
+    set_license("MIT")
+
+    add_urls("https://github.com/duckdb/duckdb/releases/download/$(version)/libduckdb-src.zip",
+             "https://github.com/duckdb/duckdb.git")
+    add_versions("v0.10.0", "385e27aa67712813e4a07389465c4c5c45c431d97cddd35713b8a306d2a86f2d")
+
+    on_install("macosx", "linux", "windows", function (package)
```
SirLynix, line 10: is there a reason to restrict it to theses platforms?

packages/d/duckdb/xmake.lua
```diff
@@ -0,0 +1,32 @@
...
+        io.writefile("xmake.lua", [[
+            add_rules("mode.debug", "mode.release")
+            set_languages("c++17")
```
SirLynix, line 13: The README of this library says it needs c++11, why c++17?

packages/d/duckdb/xmake.lua
```diff
@@ -0,0 +1,32 @@
...
+            target("duckdb")
+                set_kind("shared")
```
SirLynix, line 15: why not let package decide kind?

packages/d/duckdb/xmake.lua
```diff
@@ -0,0 +1,32 @@
...
+    on_test(function (package)
+      assert(package:check_cxxsnippets({test = [[
+          #include "duckdb.hpp"
+          using namespace duckdb;
+
+          void test() {
+            DuckDB db(nullptr);
+            Connection con(db);
+          }
+      ]]}, {configs = {languages = "cxx17"}}))
```
SirLynix, line 31: same, why c++17?

SirLynix (changes requested):

packages/d/duckdb/xmake.lua
c8ef: Because the libduckdb package only contains three file: duckdb.hpp, duckdb.h, duckdb.cpp. If we did not set this, the xmake will complain cannot find main.

packages/d/duckdb/xmake.lua
c8ef: Because the upstream provide binary for these platforms.

packages/d/duckdb/xmake.lua
waruqi: use `set_kind("$(kind)")`, we need to switch static/shared dynamically.

c8ef: done

packages/d/duckdb/xmake.lua
c8ef:
```
 49171 |     const S& format, basic_format_args<basic_printf_context_t<Char>> args) {
      |     ^~~~~
duckdb.cpp:49170:41: error: expected ')' before 'const'
49170 | inline std::basic_string<Char> vsprintf(
      |                                        ~^
      |                                         )
49171 |     const S& format, basic_format_args<basic_printf_context_t<Char>> args) {
      |     ~~~~~                                
duckdb.cpp:49188:1: warning: inline variables are only available with '-std=c++17' or '-std=gnu++17' [-Wc++17-extensions]
49188 | inline std::basic_string<Char> sprintf(const S& format, const Args&... args) {
      | ^~~~~~
duckdb.cpp:49186:23: error: parameter pack 'Args' must be at the end of the template parameter list
49186 | template <typename S, typename... Args,
      |                       ^~~~~~~~
duckdb.cpp:49188:40: error: expected unqualified-id before 'const'
49188 | inline std::basic_string<Char> sprintf(const S& format, const Args&... args) {
```

This is the amalgamation library, not the original source. So it might need higher version of cpp

packages/d/duckdb/xmake.lua
c8ef: same as above

PR #3381, opened 2024-03-01
inih: add more options

Chi-EEE:


packages/i/inih/xmake.lua
```diff
@@ -8,16 +8,36 @@ package("inih")
...
     add_configs("heap", {description = "allocate memory on the heap using malloc instead using a fixed-sized line buffer on the stack", default = false, type = "boolean"})
     add_configs("max_line_length", {description = "maximum line length in bytes", default = "200", type = "string"})
+    add_configs("initial_malloc_size", {description = "initial malloc size in bytes (when using the heap)", default = "200", type = "string"})
     add_configs("allow_realloc", {description = "allow initial malloc size to grow to max line length (when using the heap)", default = false, type = "boolean"})
 
     on_install(function (package)
         os.cp(path.join(os.scriptdir(), "port", "xmake.lua"), "xmake.lua")
         local configs = {}
         configs.ini_parser = package:config("ini_parser")
+        configs["multi-line_entries"] = package:config("multi-line_entries")
```
waruqi, line 29: we can use `_` instead of `-` to simplify config name

packages/i/inih/xmake.lua
```diff
@@ -8,16 +8,36 @@ package("inih")
...
+        configs["utf-8_bom"] = package:config("utf-8_bom")
+        configs.inline_comments = package:config("inline_comments")
+        configs.inline_comment_prefix = package:config("inline_comment_prefix")
+        configs["start-of-line_comment_prefix"] = package:config("start-of-line_comment_prefix")
+        configs.allow_no_value = package:config("allow_no_value")
+        configs.stop_on_first_error = package:config("stop_on_first_error")
+        configs.report_line_numbers = package:config("report_line_numbers")
+        configs.call_handler_on_new_section = package:config("call_handler_on_new_section")
```
waruqi, line 37: and we can use for-loop to pass all configs

https://github.com/xmake-io/xmake-repo/blob/719c8010f4205afdd4777622190aa7ee6eba0515/packages/f/ffmpeg/xmake.lua#L117

PR #3391, opened 2024-03-02
injector: add package

Chi-EEE:


packages/i/injector/xmake.lua
```diff
@@ -0,0 +1,29 @@
+package("injector")
```
waruqi, line 1: `set_kind("library", {headeronly = true})`

Chi-EEE: This package is using the compiled files from make/nmake so it would not be headeronly

PR #3539, opened 2024-03-17
[commonlibsse-ng] Disable warning C4530 (MSVC 14.39.33519)

Meridiano:
C4530 is just a one more warning to suppress.
```
D:\Dev\VisualStudio\VSCommunity2022\VC\Tools\MSVC\14.39.33519\include\chrono(2330): error C2220: the following warning is treated as an error
D:\Dev\VisualStudio\VSCommunity2022\VC\Tools\MSVC\14.39.33519\include\chrono(2330): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
```


qudix:
This is a problem with xmake, it will be fixed in the next version, or just downgrade for now.

star-hengxing:
see https://github.com/xmake-io/xmake/discussions/4838

PR #3540, opened 2024-03-18
Auto-update moltenvk to v1.2.8

waruqi:
New version of moltenvk detected (package version: nil, last github version: v1.2.8)

### DIFF
--- a/packages/m/moltenvk/xmake.lua
+++ b/packages/m/moltenvk/xmake.lua
@@ -5,6 +5,7 @@ package("moltenvk")
 
     add_urls("https://github.com/KhronosGroup/MoltenVK/archive/refs/tags/$(version).tar.gz",
              "https://github.com/KhronosGroup/MoltenVK.git")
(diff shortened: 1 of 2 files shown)


packages/m/moltenvk/xmake.lua
```diff
@@ -12,6 +12,9 @@ package("moltenvk")
     add_versions("v1.1.4", "f9bba6d3bf3648e7685c247cb6d126d62508af614bc549cedd5859a7da64967e")
     add_versions("v1.1.0", "0538fa1c23ddae495c7f82ccd0db90790a90b7017a258ca7575fbae8021f3058")
 
+    add_links("MoltenVKShaderConverter", "MoltenVK")
+    add_frameworks("CoreGraphics")
```
waruqi, line 16: move it to `if is_plat("macosx")` branch

SirLynix: but this package only supports macosx and iphoneos plat

waruqi: Keeping a consistent configuration style for all packages allows users to read the configuration more clearly. Also, not everyone knows that this package only supports macosx, which can lead to users incorrectly assuming that add_frameworks can be used on all platforms.

Issue #3547, opened 2024-03-18
Add srpc in xmake-repo


holmes1412:
### Is your feature request related to a problem? Please describe.

In order to help developers use srpc more conveniently, I wonder if srpc could be added to xmake-repo?

project link : https://github.com/sogou/srpc

recent releases : https://github.com/sogou/srpc/releases/tag/v0.10.2

### Describe the solution you'd like

Since I'm not very familiar with all the platforms, it would be very appreciated if you could help me to add it the first time. I will maintain the update in the future.

waruqi:
https://github.com/xmake-io/xmake-repo/pull/3548

youngday:
workflow have add xmake compile project,
if srpc can add xmake compile project?
i can generate  pb.h files with xmake ,but  no rpc.h files,if srpcgenerate can build by xmake ?
thanks!

holmes1412:
@youngday Thank you for the feedback! I would like to add xmake complie project in the future. Although I think users could use cmake or bazel, not necessarily xmake to build SRPC, even if it is in the xmake-repo. Maybe I’m wrong :) 

If you build your own project, you should call command like : ‘srpc_generator …’ to make the `rpc.h` files. Did you get the binary tool `srpc_generator` now in xmake-repo?

youngday:
i get rpc.h by   binary tool srpc_generator,have not do it in xmake-repo ,and  manual copy libworkflow.so to exe path ,and run , these  have not issued.

waruqi:
> workflow have add xmake compile project, if srpc can add xmake compile project? i can generate pb.h files with xmake ,but no rpc.h files,if srpcgenerate can build by xmake ? thanks!

This has nothing to do with whether or not you use xmake to build srpc, you should write a custom rule to call srpc_generator to generate it.

like pb.h, it's generated in proto rule.

https://github.com/xmake-io/xmake/blob/master/xmake/rules/protobuf/xmake.lua
https://github.com/xmake-io/xmake/blob/e9cf2231246548d342de3583b5b7baa1a0fe8842/xmake/rules/protobuf/proto.lua#L151

We can also add custom rules to the srpc package to distribute it.

like this

https://github.com/xmake-io/xmake-repo/blob/dev/packages/c/commonlibsse-ng/rules/plugin.lua

PR #3548, opened 2024-03-18
add srpc

waruqi:
https://github.com/xmake-io/xmake-repo/issues/3547


Barenboim:
这个android编译不过吗？

youngday:
it can compiled ,but need run with libworkflow.so, how to add link in xmake.lua?
my  example rep at 
```sh https://github.com/youngday/xmake-rpc```
thanks

PR #3577, opened 2024-03-21
imgui: fix sdl config

star-hengxing:
https://github.com/xmake-io/xmake/issues/4859


AidanSun05:
I noticed this PR reverts the C++ revision to C++11, is there a reason for this? I had changed it to C++14 in #3545 to fix compilation issues when using Clang on Windows.

PR #3584, opened 2024-03-22
Auto-update emio to 0.7.0

waruqi:
New version of emio detected (package version: nil, last github version: 0.7.0)


star-hengxing:
Full c++20 not support

PR #3593, opened 2024-03-24
Auto-update microsoft-proxy to 2.2.1

waruqi:
New version of microsoft-proxy detected (package version: nil, last github version: 2.2.1)


c8ef:
The CI fails because certain instances in the CI lack full support for C++20 due to compiler limitations.

PR #3643, opened 2024-03-30
openssl: Fix compilation for Android on Windows

SirLynix:
temp fix
see https://github.com/xmake-io/xmake/pull/4903

PR #3644, opened 2024-03-30
libcurl: Use CMake to compile on Android

SirLynix:
Fixes Android on Windows

I had to pass a lot of options (I only supported option using lib supporting Android) and made a `handledependency` function for that.

It seems to be a common need, wouldn't it be nice to make it part of cmake, maybe in package.tools.cmake?


SirLynix:
It will fail on Windows until https://github.com/xmake-io/xmake-repo/pull/3643 is merged

SirLynix:
It fails on Android because `fseeko` is only available starting with API 24, and CMake wrongly detects it as available. Maybe a missing CMake config, or a CMake bug?
```cmake
check_symbol_exists(fseeko         "${CURL_INCLUDES};stdio.h" HAVE_FSEEKO)
```

```c
#if __ANDROID_API__ >= 24
int fgetpos(FILE* _Nonnull __fp, fpos_t* _Nonnull __pos) __RENAME(fgetpos64) __INTRODUCED_IN(24);
int fsetpos(FILE* _Nonnull __fp, const fpos_t* _Nonnull __pos) __RENAME(fsetpos64) __INTRODUCED_IN(24);
int fseeko(FILE* _Nonnull __fp, off_t __offset, int __whence) __RENAME(fseeko64) __INTRODUCED_IN(24);
off_t ftello(FILE* _Nonnull __fp) __RENAME(ftello64) __INTRODUCED_IN(24);
#endif /* __ANDROID_API__ >= 24 */
```

packages/l/libcurl/xmake.lua
```diff
@@ -117,45 +120,45 @@ package("libcurl")
...
-        table.insert(configs, package:config("openldap") and "--enable-ldap" or "--disable-ldap")
-        if package:is_plat("macosx") then
-            local cares = package:dep("c-ares")
-            if cares and not cares:config("shared") then
-                -- we need fix missing `-lresolv` when checking c-ares
-                io.replace("./configure", "PKGCONFIG --libs-only-l libcares", "PKGCONFIG --libs-only-l --static libcares", {plain = true})
+        local function handledependency(conf, depname, includeconfig, libconfig)
+            if package:config(conf) then
+                local dep = package:dep(depname)
+                if dep and not dep:is_system() then
```
waruqi, line 126: why? fetch supports system libs

SirLynix: I can remove it, but the idea is that cmake should be able to find it if it's a system library

waruqi: ok

PR #3655, opened 2024-04-01
Auto-update libremidi to v4.5.0

waruqi:
New version of libremidi detected (package version: nil, last github version: v4.5.0)


star-hengxing:
mac ci unsupportted c++20 std::biset

PR #3695, opened 2024-04-05
Auto-update simdutf to v5.2.2

waruqi:
New version of simdutf detected (package version: nil, last github version: v5.2.2)


star-hengxing:
```
src\arm64/arm_base64.cpp(49,7): error C2664: '__n128x4 neon_ld4m_q8(const char *)': cannot convert argument 1 from 'const uint8_t [64]' to 'const char *' 
```

I think vld4q_u8(neon_ld4m_q8) may not work with some old windows sdk.

PR #3712, opened 2024-04-06
add zmqpb v0.8 & v0.9, add zeromq brew source

SFGrenade:
also use c++17


SFGrenade:
i suppose i'm missing some `utf8_range` link either here or in the zmqpb xmake.lua

waruqi:
try add `add_deps("utf8_range")` ?

SFGrenade:
> Homebrew provides zeromq support for MacOSX arm64, as seen in their formula located at: https://github.com/Homebrew/homebrew-core/blob/master/Formula/z/zeromq.rb.
> 
> Considering this, I believe there is no justification for us to limit our support solely to the x86 platform.

i could add the homebrew source in https://github.com/xmake-io/xmake-repo/blob/master/packages/z/zeromq/xmake.lua

SFGrenade:
> try add `add_deps("utf8_range")` ?

can do, would need to also create a new v0.5, as the xmake.lua needs to have the require and package.

SFGrenade:
i love it when a shared object file of a dependency is not found

SFGrenade:
> [windows arm is apparently not supported by protoc, which this needs](https://github.com/xmake-io/xmake-repo/pull/3712/commits/d489e9c52c325f6e6ed416f29a8242dbb51da4e2)

interesting that protobuf-cpp / protobuf-c list general windows, without excluding arm

SFGrenade:
i love it when protobuf fails to publicise libasl_* for linkage

SFGrenade:
> > > [windows arm is apparently not supported by protoc, which this needs](https://github.com/xmake-io/xmake-repo/pull/3712/commits/d489e9c52c325f6e6ed416f29a8242dbb51da4e2)
> > 
> > interesting that protobuf-cpp / protobuf-c list general windows, without excluding arm
> 
> The reason for this is that the current Windows CI lacks native ARM environment support. However, this does not imply that the package cannot be built on Windows ARM.

ah, fair

SFGrenade:
so the fails are:
- protobuf not allowing linking of libabsl-* on linux shared
- windows arm

packages/z/zmqpb/xmake.lua
```diff
@@ -20,7 +26,6 @@ package("zmqpb")
     end)
 
     on_install("windows", "macosx", "linux", function (package)
-        local configs = {}
         import("package.tools.xmake").install(package, configs)
```
waruqi, line 29: configs is nil

packages/z/zmqpb/xmake.lua
```diff
@@ -9,9 +9,15 @@ package("zmqpb")
...
     add_versions("0.4", "7c0001db73b19e65b007adf6c9c5092c3589f043ab3e95a16b3ec2b2a87fa244")
+    add_versions("0.5", "d177c3b87e367932973da20b72376d1320c7501e0a2175c929d4c5e1f06f68ad")
 
     add_deps("cppzmq")
     add_deps("protobuf-cpp")
+    -- protobuf needs it and somehow just doesn't publicizes the linkage
+    add_deps( "abseil" )
+    add_deps( "utf8_range" )
+
+    add_configs("shared", {description = "Build shared binaries.", default = false, type = "boolean", readonly = false})
```
waruqi, line 20: so is still generated.

```
    "/home/runner/.xmake/packages/z/zmqpb/0.5/98c46d812b8c40388d90d102c2ec2e9b/lib/libZmqPb.so",
    "/home/runner/.xmake/packages/z/zmqpb/0.5/98c46d812b8c40388d90d102c2ec2e9b/lib/libZmqPb.so.0.5.0",
    "/home/runner/.xmake/packages/z/zmqpb/0.5/98c46d812b8c40388d90d102c2ec2e9b/lib/libZmqPb.so.0", 
  },
```

waruqi: zmqpb need to disable shared lib if run `xmake f -k static`

SFGrenade: i saw that in another package, but i took it out just now

SFGrenade:
i believe there were workflow-updates? otherwise https://github.com/xmake-io/xmake-repo/actions/runs/8910062655/job/24468532430 doesn't make sense

waruqi:
Maybe this patch break it. https://github.com/xmake-io/xmake/pull/5023/files

https://github.com/xmake-io/xmake/blob/f6cd253d33ecaa60087e9dd22a5bd92afb31c40f/xmake/modules/package/manager/xmake/find_package.lua#L137

But I don't know why.

waruqi:
> i believe there were workflow-updates? otherwise https://github.com/xmake-io/xmake-repo/actions/runs/8910062655/job/24468532430 doesn't make sense

I have fixed it.

waruqi:
https://github.com/SFGrenade/ZmqPb-Cpp/blob/ab60ed9209531a84044f7a4ebdef53b9797eb143/xmake.lua#L23

it will use protobuf-cpp/shared if this package is shared, but `add_deps("protobuf-cpp")` is static library.

https://github.com/xmake-io/xmake-repo/blob/58b8f25a1f8c961567d4115cda874d713e52d9c0/packages/z/zmqpb/xmake.lua#L15

SFGrenade:
oh, didn't know add_deps adds as a static library

SFGrenade:
now it should be good

PR #3716, opened 2024-04-07
Auto-update openjpeg to v2.5.2

waruqi:
New version of openjpeg detected (package version: nil, last github version: v2.5.2)


apocelipes:
Close because of #8704.

PR #3366, opened 2024-02-29
v8: Add version 12.4.111

PerikiyoXD:
As the title says.

Part of the v8 fix operation.

PR #3717, opened 2024-04-07
v8: Add version 12.8.168

waruqi:
https://github.com/xmake-io/xmake-repo/pull/3366


PerikiyoXD:
I will have a look at it

waruqi:
It's weird. macos ci doesn't work, it's still in the same state as before.

PR #3720, opened 2024-04-07
vulkan-hpp: update to v1.2.281, fix MSVC bug

vspefs:
- Added versions from v1.3.276 to the latest v1.2.281.
- Fixed a bug when trying to compile the `vulkan_hpp` named module using MSVC, by adding a config to enable a workaround provided by vulkan-hpp officially.


qudix:
I noticed this too late, I don't agree with this change.

waruqi:
> I noticed this too late, I don't agree with this change.

You can open a pr to improve it.

c8ef:
> I noticed this too late, I don't agree with this change.

What is your specific reason for this change? We can revert it if needed.

packages/v/vulkan-hpp/xmake.lua
```diff
@@ -18,8 +18,15 @@ package("vulkan-hpp")
...
     add_versions("v1.3.275", "1a24b015830c116632a0723f3ccfd1f06009ce12")
+    add_versions("v1.3.276", "d4b36b82236e052a5e6e4cea5fe7967d5b565ebc")
+    add_versions("v1.3.277", "c5c1994f79298543af93d7956b654bdefdfbdd26")
+    add_versions("v1.3.278", "29723f90a127ff08d9099855378162f04b4ffddd")
+    add_versions("v1.3.279", "6fb8def27290f8b87d7835457a9c68190aed9a9a")
+    add_versions("v1.3.280", "e35acfe75215116029298aebf681170559a4fe6a")
+    add_versions("v1.3.281", "88d508b32f207ba85b37fe22fe3732322d1c248d")
 
     add_configs("modules", {description = "Build with C++20 modules support.", default = false, type = "boolean"})
+    add_configs("msvc_modules", {description = "If 'modules' enabled, and you wish to use MSVC on the package, enable this to avoid a known bug of MSVC.", default = false, type = "boolean"})
```
SirLynix, line 29: Why an option instead of enabling this fix automatically if toolchain is msvc and modules are enabled?

packages/v/vulkan-hpp/xmake.lua
```diff
@@ -30,8 +37,8 @@ package("vulkan-hpp")
         end
     end)
 
-
-    on_install("windows|x86", "windows|x64", "linux", "macosx", "mingw", "android", "iphoneos", function (package)
+    -- TODO: add android, windows|x86, mingw|i386 target
```
SirLynix, line 40: we should not remove already supported platforms without a good reason

packages/v/vulkan-hpp/xmake.lua
```diff
@@ -56,32 +63,41 @@ package("vulkan-hpp")
...
-                    add_headerfiles("Vulkan-Headers/include/(**.h)")
-                    add_headerfiles("Vulkan/(**.h)")
-                    add_headerfiles("Vulkan-Headers/include/(**.hpp)")
-                    add_headerfiles("Vulkan/(**.hpp)")
-                    add_includedirs("Vulkan")
-                    add_includedirs("Vulkan-Headers/include")
-                    add_files("Vulkan/vulkan.cppm", {public = true})
-            ]])
+            if package:config("msvc_modules") then
+                io.writefile("xmake.lua", [[
```
SirLynix, line 67: why is this necessary?

PR #3729, opened 2024-04-08
Update reactiveplusplus to v2.1.1

paul-reilly:
Also fixed `package:version()` returning `nil` when requiring a specific git commit.


paul-reilly:
Some builds appear to be failing because of C++20 compatibility:

```bash
rpp/utils/constraints.hpp:50:44: error: 'Fn' does not refer to a value
```
```cpp
template<typename Ret, typename Fn, typename... Args>
concept invocable_r_v = std::invocable<Fn, Args...> && std::same_as<Ret, std::invoke_result_t<Fn, Args...>>;
```
[Source location](https://github.com/xmake-io/xmake-repo/actions/runs/8597855854/job/23557412579?pr=3729#step:5:129)

https://godbolt.org/z/bPxz4fGbT

PR #3742, opened 2024-04-09
Update all of plflib packages

Cons-Cat:
I have updated `plf_colony` (which changed its versioning scheme to be compatible with vcpkg's), and added plflib libraries which were missing for some reason. I have also added a `plf` metapackage to simply grab all of plflib at once.


Cons-Cat:
The issue appears to be that Visual Studio 2019 cannot build `plf_hive`, however Visual Studio 2022 can. I'm not sure how to properly fix this. @waruqi Do you have advice?

waruqi:
I think that you should split into separate patches, with one pr submitting one package, to make it easier to debug and merge.

Cons-Cat:
As far as I can tell, all the errors are coming from `plf::hive`. The libraries in general are extremely compatible, even for C++03, _except_ for `plf::hive` which is pretty modern. I've removed that from the PR for now.

packages/p/plflibs/xmake.lua
```diff
@@ -0,0 +1,17 @@
...
+
+    on_install(function (package)
+        package:add("deps", "plf_colony")
+        package:add("deps", "plf_indiesort")
+        package:add("deps", "plf_list")
+        package:add("deps", "plf_queue")
+        package:add("deps", "plf_rand")
+        package:add("deps", "plf_reorderase")
+        package:add("deps", "plf_stack")
+    end)
```
star-hengxing, line 17: Unnecessary code, can we remove it?

waruqi: sure

PR #3751, opened 2024-04-10
Add `plf_nanotimer` package

Cons-Cat:
Split this failing package out of https://github.com/xmake-io/xmake-repo/pull/3742


Cons-Cat:
I'm thinking that `plf::nanotimer` simply doesn't support WASM, probably.

star-hengxing:
> I'm thinking that `plf::nanotimer` simply doesn't support WASM, probably.

We can disable unsupported plat
https://xmake.io/#/manual/package_dependencies?id=platform-filtering

PR #3752, opened 2024-04-10
Add `plf_hive` package

Cons-Cat:
Split this failing package out of https://github.com/xmake-io/xmake-repo/pull/3742

### DIFF
--- a/packages/p/plf_hive/xmake.lua
+++ b/packages/p/plf_hive/xmake.lua
@@ -5,23 +5,12 @@ package("plf_hive")
     set_license("zlib")
 
     add_urls("https://github.com/mattreecebentley/plf_hive.git")
-    add_versions("latest", "bebd22a4fa017c09bfa0100603e5c7ff4af6c01e")
+    add_versions("2024.04.21", "7689475b1fa2a95228cf0f44db9c209d7e430748")
 
     on_install(function (package)
         os.cp("plf_hive.h", package:installdir("include"))
     end)
 
     on_test(function (package)
-        assert(package:check_cxxsnippets({test = [[
-            #include <version>
-// plf_hive requires C++20 ranges compatibility.
-#ifdef __cpp_lib_containers_ranges
-            #include <plf_hive.h>
-            void test() {
-                plf::hive<int> i_hive;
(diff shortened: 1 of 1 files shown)


Cons-Cat:
It appears to pass the tests on recent compilers and fail the ones which do not fully support C++20.

packages/p/plf_hive/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("plf_hive")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://plflib.org/colony.htm")
+    set_description("plf::hive is a fork of plf::colony to match the current C++ standards proposal.")
+    set_license("zlib")
+
+    add_urls("https://github.com/mattreecebentley/plf_hive.git")
+    add_versions("latest", "bebd22a4fa017c09bfa0100603e5c7ff4af6c01e")
```
star-hengxing, line 8: We use date for version, like this:

https://github.com/xmake-io/xmake-repo/blob/3a4a573cdfc1c6d86785c7d689ac02fc2863e4b9/packages/f/fast_io/xmake.lua#L11

packages/p/plf_hive/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+
+    on_install(function (package)
+        os.cp("plf_hive.h", package:installdir("include"))
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            #include <version>
+// plf_hive requires C++20 ranges compatibility.
+#ifdef __cpp_lib_containers_ranges
```
star-hengxing, line 18: This is not recommended, we need to know that the reason ci fails is because there is no c++20 ranges lib and not some other error.

packages/p/plf_hive/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+            #include <plf_hive.h>
+            void test() {
+                plf::hive<int> i_hive;
```
star-hengxing, line 21: We can use `has_cxxtypes` simplfy `on_test`

PR #3753, opened 2024-04-10
zeus_expected: add package

Chi-EEE:


packages/z/zeus_expected/xmake.lua
```diff
@@ -0,0 +1,30 @@
...
+
+    add_urls("https://github.com/zeus-cpp/expected/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/zeus-cpp/expected.git")
+
+    add_versions("v1.0.0", "a0d81798b777f9bfcc1e1e4f3046632067bd8c6071dbfcbec5012a31a5aebc68")
+
+    add_patches("v1.0.0", path.join(os.scriptdir(), "patches", "v1.0.0", "fix_typename.patch"), "710d71f8c765a2937df25a2c52abec24f5f4ef5f43281f6aa01853d0498e2a47")
+
+    if is_plat("windows") then
+        add_cxxflags("/Zc:__cplusplus")
```
SirLynix, line 15: I'm not comfortable with the package adding the flag globally. You can add it in the on_test callback and in your project, but having the package add a flag to the whole project seems too much.

Chi-EEE: I moved it to the on_install callback since there are other packages which do the same & you would have to add the flag anyways since the package does not allow you to use it unless it has the flag

SirLynix: You don't have to add the flag to the whole project, people may be willing to only set it up in their one file including this header.
```lua
add_files("src/xx.cpp", {cxflags={"/Zc:__cplusplus"}})
```

unfortunately there is still code that will break with `/Zc:__cplusplus`

Chi-EEE: Ok, I moved the flag into the on_test callback

PR #3754, opened 2024-04-10
Update WebUI Test Code

AlbertShown:
Just adding `webui.js` which is necessary to ensure that test app is working fine.


SirLynix:
Note that the produced executable is not ran so this won't change anything, but I think we can merge this for a more practical example.

star-hengxing:
on_test only tests if the build was successful, and I don't think many people would refer to the code here

PR #3793, opened 2024-04-15
Add cmark package

charlesseizilles:


packages/c/cmark/xmake.lua
```diff
@@ -0,0 +1,23 @@
+package("cmark")
+    set_kind("library")
```
star-hengxing, line 2: package default kind is library, we need't add it

packages/c/cmark/xmake.lua
```diff
@@ -0,0 +1,23 @@
...
+        local configs = {
+            "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"),
+            "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"),
+            "-DBUILD_TESTING=OFF",
+        }
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_test(function (package)
+        io.read()
```
star-hengxing, line 21: Please remove unused code

packages/c/cmark/xmake.lua
```diff
@@ -0,0 +1,21 @@
+package("cmark")
+    set_homepage("https://github.com/commonmark/cmark")
+    set_description("CommonMark parsing and rendering library and program in C")
+
+    add_urls("https://github.com/commonmark/cmark.git")
+    add_versions("0.31.0", "8fbf029685482827828b5858444157052f1b0a5f")
```
star-hengxing, line 6: Why not add [targz](https://github.com/commonmark/cmark/archive/refs/tags/0.31.0.tar.gz) archive?

charlesseizilles: no particular reason, I can add it if you want

star-hengxing: We can use `xmake l scripts/new.lua github:commonmark/cmark` to automatically generate package, which will save some time

charlesseizilles: oh I didn't know about that!

PR #3801, opened 2024-04-16
opencv: add avif to feature table

Chi-EEE:


waruqi:
I have fixed ci error, https://github.com/xmake-io/xmake-repo/pull/4311

PR #3815, opened 2024-04-18
Auto-update zmqpb to 0.6

waruqi:
New version of zmqpb detected (package version: nil, last github version: 0.6)


c8ef:
done in #3712

PR #3825, opened 2024-04-20
Auto-update iowow to v1.4.18

waruqi:
New version of iowow detected (package version: nil, last github version: v1.4.18)


c8ef:
This lib's bsd sort definition seems to have some problem:

- https://github.com/Softmotions/iowow/blob/v1.4.18/src/utils/sort_r.h#L258-L261
- https://man.freebsd.org/cgi/man.cgi?query=qsort_r&sektion=3&format=html

Use its own implementation of sort now.

PR #3861, opened 2024-04-24
Auto-update brpc to 1.9.0

waruqi:
New version of brpc detected (package version: nil, last github version: 1.9.0)


c8ef:
addressed in #4820

PR #3879, opened 2024-04-27
Auto-update h5cpp to v0.6.0

waruqi:
New version of h5cpp detected (package version: nil, last github version: v0.6.0)


star-hengxing:
hdf5 会构建一个程序并运行探测环境，所以做不了交叉编译，只能本地编译

---

hdf5 will build a program and run to detect the environment, so can not do cross compilation, only native compilation

PR #3935, opened 2024-05-03
Auto-update glfw3webgpu to v1.2.0

waruqi:
New version of glfw3webgpu detected (package version: nil, last github version: v1.2.0)

### DIFF
--- a/packages/g/glfw3webgpu/xmake.lua
+++ b/packages/g/glfw3webgpu/xmake.lua
@@ -6,6 +6,7 @@ package("glfw3webgpu")
     add_urls("https://github.com/eliemichel/glfw3webgpu/archive/refs/tags/$(version).tar.gz",
              "https://github.com/eliemichel/glfw3webgpu.git")
 
+    add_versions("v1.2.0", "28387b960aff573728bde2bf0fa876c33608cdadaca8d23f4a46cd31920ab633")
     add_versions("v1.1.0", "307ba86a724adc84a875e8bd2374baad0fabc77797f20f9a1779eef8d9ffe95a")
     add_versions("v1.0.1", "b98c63f1905f0e4cf99229de8b7e5c2693fdf3b8d684b5a43d60f21f67d33e6b")
 


SirLynix:
```
finding libx11 from xmake ..
error: @programdir/modules/package/manager/xmake/find_package.lua:137: wrong number of arguments to 'insert'
stack traceback:
    [C]: in function 'insert'
    [@programdir/modules/package/manager/xmake/find_package.lua:137]: in function '_find_package_from_repo'
    [@programdir/modules/package/manager/xmake/find_package.lua:350]:
    [@programdir/modules/package/manager/find_package.lua:110]: in function '_find_package'
    [@programdir/modules/package/manager/find_package.lua:194]:
    [@programdir/modules/lib/detect/find_package.lua:109]:
    [@programdir/core/package/package.lua:1794]: in function '_fetch_library'
    [@programdir/core/package/package.lua:1947]: in function 'fetch'
    [.../modules/private/action/require/impl/actions/install.lua:426]:

  => install libx11 1.8.7 .. failed
error: @programdir/core/main.lua:329: @programdir/modules/async/runjobs.lua:322: .../modules/private/action/require/impl/actions/install.lua:506: install failed!
stack traceback:
    [C]: in function 'error'
    [@programdir/core/base/os.lua:973]:
    [.../modules/private/action/require/impl/actions/install.lua:506]: in function 'catch'
    [@programdir/core/sandbox/modules/try.lua:123]: in function 'try'
    [.../modules/private/action/require/impl/actions/install.lua:370]:
    [...modules/private/action/require/impl/install_packages.lua:487]: in function 'jobfunc'
    [@programdir/modules/async/runjobs.lua:238]:

```

🤔

SirLynix:
the issue is that the function doesn't return a value anymore, which is not the same as returning nil.

Example:
```lua
function foo()
  return nil
end

function bar()
end

local t = {}
print("inserting nil")
table.insert(t, foo()) -- ok
print("inserting nothing")
table.insert(t, bar()) -- error: wrong number of arguments
```

```
inserting nil
inserting nothing

lua5.3: Main.lua:12: wrong number of arguments to 'insert'
stack traceback:
	[C]: in function 'table.insert'
	Main.lua:12: in main chunk
	[C]: in ?
```
https://onecompiler.com/lua/42c5eryhc

I will make a PR to fix it

waruqi:
> the issue is that the function doesn't return a value anymore, which is not the same as returning nil.
> 
> Example:
> 
> ```lua
> function foo()
>   return nil
> end
> 
> function bar()
> end
> 
> local t = {}
> print("inserting nil")
> table.insert(t, foo()) -- ok
> print("inserting nothing")
> table.insert(t, bar()) -- error: wrong number of arguments
> ```
> 
> ```
> inserting nil
> inserting nothing
> 
> lua5.3: Main.lua:12: wrong number of arguments to 'insert'
> stack traceback:
> 	[C]: in function 'table.insert'
> 	Main.lua:12: in main chunk
> 	[C]: in ?
> ```
> 
> https://onecompiler.com/lua/42c5eryhc
> 
> I will make a PR to fix it

thanks.

PR #3939, opened 2024-05-04
Auto-update libimagequant to 4.3.1

waruqi:
New version of libimagequant detected (package version: nil, last github version: 4.3.1)


star-hengxing:
> libimagequant v2 used to be a C library. libimagequant v4 is written entirely in Rust, but still exports the same C interface for C programs. You will need to install Rust 1.70+ to build it, and adjust your build commands. If you do not want to upgrade, you can keep using [the C version of the library](https://github.com/imageoptim/libimagequant/tree/2.x) in the 2.x branch of the [repo](https://github.com/ImageOptim/libimagequant).

So we need to package cargo first.

PR #3946, opened 2024-05-04
fix limits for packages

Chi-EEE:
https://github.com/xmake-io/xmake/issues/5048

### DIFF
--- a/packages/t/tobiaslocker_base64/xmake.lua
+++ b/packages/t/tobiaslocker_base64/xmake.lua
@@ -7,7 +7,7 @@ package("tobiaslocker_base64")
     add_urls("https://github.com/tobiaslocker/base64.git")
     add_versions("2024.02.26", "387b32f337b83d358ac1ffe574e596ba99c41d31")
 
-    on_install(function (package)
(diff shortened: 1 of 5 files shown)


packages/t/tobiaslocker_base64/xmake.lua
```diff
@@ -7,7 +7,7 @@ package("tobiaslocker_base64")
     add_urls("https://github.com/tobiaslocker/base64.git")
     add_versions("2024.02.26", "387b32f337b83d358ac1ffe574e596ba99c41d31")
 
-    on_install("windows|!arm64", function (package)
+    on_install("!windows|arm64", function (package)
```
qudix, line 10:
```suggestion
    on_install("windows|!arm64", "linux", function (package)
```

Chi-EEE: I would add this change but it would limit the package to only windows and linux.

qudix: That's fine, it was already limited to just windows before, you're just adding linux for now.

PR #3961, opened 2024-05-06
Revert "vulkan-hpp: update to v1.2.281, fix MSVC bug"

c8ef:
Reverts xmake-io/xmake-repo#3720


SirLynix:
I'd like @Arthapz to review the module part

waruqi (approved):

Arthapz:
we should use moduleonly target kind if xmake version is > to 2.8.9

c8ef:
We can mark this package as deprecated like [vcpkg does](https://github.com/microsoft/vcpkg/blob/master/ports/vulkan-hpp/usage), since the vulkan-headers package already includes the vulkan-hpp package and offers [module support](https://github.com/KhronosGroup/Vulkan-Headers/blob/main/include/vulkan/vulkan.cppm). @waruqi

PR #3977, opened 2024-05-07
add wigxjpf

0382:
The wigxjpf library is used for evaluating Wigner symbols, which is very useful for angular momentum coupling. The GNU-GSL library offers `gsl_sf_coupling_?j` functions, but they are inaccurate for very large angular momentum. This library use big integer techniques to give accurate result.

### DIFF
--- a/packages/w/wigxjpf/xmake.lua
+++ b/packages/w/wigxjpf/xmake.lua
@@ -12,18 +12,13 @@ package("wigxjpf")
         os.cp("inc/*.h", package:installdir("include"))
         os.cp("lib/*.a", package:installdir("lib"))
         os.cp("bin/*", package:installdir("bin"))
+        if package:config("shared") then
+            table.insert(configs, "shared")
+            import("package.tools.make").build(package, configs)
+            os.cp("lib/*.so", package:installdir("lib"))
+        end
     end)
 
     on_test(function (package)
-        assert(package:check_csnippets({test = [[
-            #include "wigxjpf.h"
-            void test() {
-                double val6j;
-                wig_table_init(2 * 100, 9);
-                wig_temp_init(2 * 100);
-                val6j = wig6jj(2 * 2, 2 * 2, 2 * 1, 2 * 2, 2 * 1, 2 * 1);
-                wig_temp_free();
-                wig_table_free();
-            }
-        ]]}))
+        assert(package:has_cfuncs("wig3jj", {includes = "wigxjpf.h"}))
     end)
--- a/packages/w/wigxjpf/xmake.lua
+++ b/packages/w/wigxjpf/xmake.lua
@@ -13,9 +13,11 @@ package("wigxjpf")
         end
         import("package.tools.make").build(package, configs)
         os.cp("inc/*.h", package:installdir("include"))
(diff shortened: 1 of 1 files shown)


0382:
I have changed the script following the suggestions.

packages/w/wigxjpf/xmake.lua
```diff
@@ -0,0 +1,29 @@
...
+    set_homepage("https://fy.chalmers.se/subatom/wigxjpf/")
+    set_description("WIGXJPF evaluates Wigner 3j, 6j and 9j symbols accurately using prime factorisation and multi-word integer arithmetic.")
+    set_license("GPL-3.0", "LGPL-3.0")
+
+    set_urls("https://fy.chalmers.se/subatom/wigxjpf/wigxjpf-$(version).tar.gz")
+    add_versions("1.13", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
+
+    on_install("linux", "macosx", function (package)
+        local configs = {}
+        import("package.tools.make").build(package, configs)
```
waruqi, line 11: We need to distinguish between static/shared libraries based on `package:config("shared")` if it only provides static libraries, please mark it as static readonly. 

```
add_configs("shared",      {description = "Build shared library.", default = false, type = "boolean", readonly = true})
```

packages/w/wigxjpf/xmake.lua
```diff
@@ -0,0 +1,29 @@
...
+        os.cp("inc/*.h", package:installdir("include"))
+        os.cp("lib/*.a", package:installdir("lib"))
+        os.cp("bin/*", package:installdir("bin"))
+    end)
+
+    on_test(function (package)
+        assert(package:check_csnippets({test = [[
```
star-hengxing, line 18: We can simplify test with [has_cfuncs](https://xmake.io/#/manual/package_instance?id=packagehas_cfuncs)

packages/w/wigxjpf/xmake.lua
0382: In the Makefile, the shared library is maked by a extra target, thus by running `make shared`. I add some code for the condition.

packages/w/wigxjpf/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+    set_license("GPL-3.0", "LGPL-3.0")
+
+    set_urls("https://fy.chalmers.se/subatom/wigxjpf/wigxjpf-$(version).tar.gz")
+    add_versions("1.13", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
+
+    on_install("linux", "macosx", function (package)
+        local configs = {}
+        import("package.tools.make").build(package, configs)
+        os.cp("inc/*.h", package:installdir("include"))
+        os.cp("lib/*.a", package:installdir("lib"))
```
waruqi, line 13: if the current package is shared library, we cannot install static lib files.

0382: In the Makefile, the shared library depend on the static library. So I think I can just do not copy the static library when installing shared library.

PR #4012, opened 2024-05-11
sfml: use "sfml-all" in pkgconfig if not Windows

xaizek:
There is already a similar external source for OS X, but not for `pkgconfig` everywhere else.

### DIFF
--- a/packages/s/sfml/xmake.lua
+++ b/packages/s/sfml/xmake.lua
@@ -23,6 +23,8 @@ package("sfml")
 
     if is_plat("macosx") then
         add_extsources("brew::sfml/sfml-all")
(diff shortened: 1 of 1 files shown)


xaizek:
The 2 failed CI jobs error like this:

```
LINK : fatal error LNK1181: cannot open input file 'opengl32.lib' [C:\Users\runneradmin\AppData\Local\.xmake\cache\packages\2405\s\sfml\2.6.0\source\build_39e5f11e\src\SFML\Window\sfml-window.vcxproj]
```
which doesn't look like it was caused by this change.

SirLynix:
Thank you!

PR #4099, opened 2024-05-21
upgrade versions of `proj`

zjyhjqs:


packages/p/proj/xmake.lua
```diff
@@ -5,6 +5,10 @@ package("proj")
     set_license("MIT")
 
     add_urls("https://download.osgeo.org/proj/proj-$(version).tar.gz")
+    add_versions("9.4.0", "3643b19b1622fe6b2e3113bdb623969f5117984b39f173b4e3fb19a8833bd216")
```
waruqi, line 8: Please only update one version per pr, as ci will only test the latest version.

zjyhjqs: ok

PR #4121, opened 2024-05-23
Update tbb 2021.12.0

Altina-oz:
Fixed parallel_for_each algorithm behavior for iterators defining iterator_concept trait instead of iterator_category.
Fixed the redefinition issue for std::min and std::max on Windows* OS (https://github.com/oneapi-src/oneTBB/issues/832).
Fixed the incorrect binary search order in TBBConfig.cmake.

https://github.com/oneapi-src/oneTBB/releases

### DIFF
--- a/packages/t/tbb/xmake.lua
+++ b/packages/t/tbb/xmake.lua
@@ -76,8 +76,9 @@ package("tbb")
                 --  for r26c,r26b: ndkver is 26 and ndk_sdkver:="29" 
                 --  for the above ndk versions ,we have to a workaround
                 -- 
-                local exflags = {"-Wl", "--undefined-version"}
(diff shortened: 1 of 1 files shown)


Altina-oz:
It seems that on android r26, many symbols in the TBB linker version scripts are undefined because link time optimization, and there is a temporarily workaround that to add "-Wl, --undefined-version" to ldflags (c.f. 
 https://cgit.freebsd.org/ports/commit/?id=3677983542cc09a0e5f085e463a895e9e4dce9aa) (c.f. https://github.com/oneapi-src/oneTBB/issues/1274)

I wonder how the ndk version can be checked in xmake.

star-hengxing:
thanks

PR #4138, opened 2024-05-24
add slang

exdal:
Adds slang, an HLSL-like shader language and compiler

https://github.com/shader-slang/slang

### DIFF
--- a/packages/s/slang/xmake.lua
+++ b/packages/s/slang/xmake.lua
@@ -6,6 +6,7 @@ package("slang")
     add_urls("https://github.com/shader-slang/slang.git")
 
     add_versions("v2024.1.17", "62b7219e715bd4c0f984bcd98c9767fb6422c78f")
+    add_versions("v2021.1.18", "efdbb954c57b89362e390f955d45f90e59d66878")
 
     add_configs("shared", { description = "Build shared library", default = true, type = "boolean", readonly = true })
     add_configs("embed_stdlib_source", { description = "Embed stdlib source in the binary", default = true, type = "boolean" })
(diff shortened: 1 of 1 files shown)


exdal:
Please do not merge until this pull request is merged on remote: https://github.com/shader-slang/slang/pull/4222

star-hengxing:
If used as a shader compiler, it only needs to be built as a single executable program.

But as a library, we need to unbundle the dependencies (using xrepo package instead of third-party library source code bundled in the project), which may be difficult and a lot of work.

I would suggest downloading [source code](https://github.com/shader-slang/slang/releases/download/v2024.1.17/slang-2024.1.17-source.zip) and building it with xmake, which is easier than the cmake.

packages/s/slang/xmake.lua
```diff
@@ -0,0 +1,53 @@
+package("slang")
+    set_homepage("https://github.com/shader-slang/slang")
+    set_description("Making it easier to work with shaders")
+    set_license("MIT")
+
+    add_urls("https://github.com/shader-slang/slang.git")
+
+    add_versions("v2024.1.17", "62b7219e715bd4c0f984bcd98c9767fb6422c78f")
+
+    add_configs("shared", { description = "Build shared library", default = true, type = "boolean" })
```
waruqi, line 10: if shared is only supported, we need mark it as readonly

packages/s/slang/xmake.lua
```diff
@@ -0,0 +1,53 @@
...
+    add_configs("embed_stdlib_source", { description = "Embed stdlib source in the binary", default = true, type = "boolean" })
+    add_configs("embed_stdlib", { description = "Build slang with an embedded version of the stdlib", default = false, type = "boolean" })
+    add_configs("full_ir_validation", { description = "Enable full IR validation (SLOW!)", default = false, type = "boolean" })
+    add_configs("asan", { description = "Enable ASAN (address sanitizer)", default = false, type = "boolean" })
```
waruqi, line 14: we need not this config. `build.sanitizer.address` policy will handle it.

PR #4159, opened 2024-05-26
feat: Update hdf5 version

DevinShine:
Update hdf5 version(1.14.2/1.14.3).

**1.14.3 CHANGELOG**

* Many parallel HDF5 bug fixes and improvements
* Improved read-only S3 VFD performance
* Many new Fortran API wrappers, including wrappers for async API calls
* A new H5Pset_actual_selection_io_mode() API call for determining if selection I/O was actually used in an I/O operation
* Improved cross-compiling support
* Better compiler support for Intel, AMD, and NVidia

**1.14.2 CHANGELOG**
* Updated HDF5 API tests CMake code to support VOL connectors
* Added support for selection I/O in the MPIO file driver
* Several bug and CVE fixes


Altina-oz:
The hdf5 library needs shlwapi on windows, >=1.14.2 may be missing this requirement in the generated CMakeConfig files, which the newer versions should now include.

cf. https://forum.hdfgroup.org/t/unresolved-external-symbol-imp-strstria/11627/6

PR #4173, opened 2024-05-27
pcre2: fix build type & update version

star-hengxing:


star-hengxing:
https://github.com/llvm/llvm-project/issues/29003

PR #4190, opened 2024-05-29
try to fix error when installing libtorch

Altina-oz:
error: ...epositories\xmake-repo-dev\packages\l\libtorch\xmake.lua:182: attempt to compare number with nil stack traceback:
    [...epositories\xmake-repo-dev\packages\l\libtorch\xmake.lua:182]: in function 'script'
    [...dir\modules\private\action\require\impl\utils\filter.lua:114]: in function 'call'
    [...\modules\private\action\require\impl\actions\install.lua:404]:

  => install libtorch v2.2.2 .. failed

Hope it can do


star-hengxing:
https://github.com/xmake-io/xmake-repo/pull/4367

PR #4204, opened 2024-05-31
Use on_source instead of on_load to set urls if supported (xmake 2.9.3)

SirLynix:
imgui and onnxruntime are not usable on xmake dev with the following error
```
error: package:set("urls", ...) can only be called in on_source().
```


SirLynix:
I can't use version in on_source but can't set urls in on_load either @waruqi

packages/o/onnxruntime/xmake.lua
```diff
@@ -53,7 +53,10 @@ package("onnxruntime")
         end
     end
 
-    on_load(function (package) 
+    -- on_source was introduced in xmake 2.9.3
+    local on_source = on_source or on_load
+
+    on_source(function (package) 
         if package:config("gpu") then
             package:add("deps", "cuda", {configs = {utils = {"cudart", "nvrtc"}}})
```
A2va, line 61: I don't know if the deps can added in the `on_source` callback

waruqi:
> I can't use version in on_source but can't set urls in on_load either @waruqi

This is because when selecting the current version, we need to call package:urls() to iterate over the urls. So we can't always change the urls based on the version. This is a circular dependency.

Previously the behaviour of this package was undefined.

packages/o/onnxruntime/xmake.lua
waruqi: we can add deps in on_load, so we need not modify it here.

waruqi: on_source is only for urls/versions

packages/i/imgui/xmake.lua
```diff
@@ -130,8 +139,10 @@ package("imgui")
         if package:config("freetype") then
             package:add("deps", "freetype")
         end
-        if package:version_str():find("-docking", 1, true) then
-            package:set("urls", {"https://github.com/ocornut/imgui.git"})
+        if not on_source then
+            if package:version_str():find("-docking", 1, true) then
+                package:set("urls", {"https://github.com/ocornut/imgui.git"})
```
waruqi, line 144: I don't know why this has to be set. If I removed it, it works fine too.

waruqi:
Now that I've relaxed the restrictions. 

We can use set/add to modify urls, .. in on_load() if urls have been inited.
But we cannot init urls, ... in on_load() if it has been not inited

packages/o/onnxruntime/xmake.lua
```diff
@@ -53,7 +53,10 @@ package("onnxruntime")
         end
     end
 
-    on_load(function (package) 
+    -- on_source was introduced in xmake 2.9.3
+    local on_source = on_source or on_load
+
+    on_source(function (package) 
         if package:config("gpu") then
```
waruqi, line 60: We can not use package:config in on_source

PR #4219, opened 2024-06-02
add vcpkg tool

sanqideshi:
add vcpkg tool

### DIFF
--- a/packages/v/vcpkg/xmake.lua
+++ b/packages/v/vcpkg/xmake.lua
@@ -10,7 +10,7 @@ package("vcpkg")
     add_deps("zip", "unzip", "cmake", "ninja", "curl")
 
     on_install("@linux", "@macosx", "@windows", function(package)
-        local scriptpath =  package:is_plat("linux") and "./bootstrap-vcpkg.sh" or "bootstrap-vcpkg.bat"
+        local scriptpath = package:is_plat("windows") and "bootstrap-vcpkg.bat" or "./bootstrap-vcpkg.sh"
         os.run(scriptpath)
         os.cp(".", package:installdir())
         package:setenv("VCPKG_ROOT", ".")
--- a/packages/v/vcpkg/xmake.lua
+++ b/packages/v/vcpkg/xmake.lua
@@ -6,10 +6,7 @@ package("vcpkg")
     add_urls("https://github.com/microsoft/vcpkg/archive/refs/tags/$(version).tar.gz")
     add_versions("2024.05.24", "3034e534d4ed13e6e6edad3c331c0e9e3280f579dd4ba86151aa1e2896b85d31")
 
-    add_deps("unzip", "cmake", "ninja", "libcurl")
-    if linuxos.name() == "archlinux" or linuxos.name() == "manjaro" then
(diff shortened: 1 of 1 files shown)


packages/v/vcpkg/xmake.lua
```diff
@@ -0,0 +1,28 @@
+package("vcpkg")
+    set_homepage("https://github.com/microsoft/vcpkg")
+    set_description(
+    "Vcpkg helps you manage C and C++ libraries on Windows, Linux and MacOS. This tool and ecosystem are constantly evolving, and we always appreciate contributions!")
```
waruqi, line 4: too lang description

packages/v/vcpkg/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+    set_homepage("https://github.com/microsoft/vcpkg")
+    set_description("Vcpkg helps you manage C and C++ libraries on Windows, Linux and MacOS.")
+    set_license("MIT")
+
+    add_urls("https://github.com/microsoft/vcpkg/archive/refs/tags/$(version).tar.gz")
+    add_versions("2024.05.24", "3034e534d4ed13e6e6edad3c331c0e9e3280f579dd4ba86151aa1e2896b85d31")
+
+    add_deps("unzip", "cmake", "ninja", "libcurl")
+    if linuxos.name() == "archlinux" or linuxos.name() == "manjaro" then
+        add_deps("pacman::zip", "pacman::curl")
```
waruqi, line 11: we should not use add_deps for 3rd packages.

sanqideshi: Because there is no zip command in the workflow, and there isn't one in xmake either, it was added. If it is removed, Arch Linux will not be supported.

waruqi: so you should add zip package to xmake-repo first and use `add_deps("zip")`, 

please do not add any pacman package deps.

sanqideshi: but,xrepo hava not a package names `zip`

waruqi: > but,xrepo hava not a package names `zip`

so you need open a pr to add it.

sanqideshi: 这个无能为力啊.十几年前的老包了,干不动

sanqideshi: This is beyond my capability. It's an old package from over a decade ago and I can't handle it.

SirLynix: There's also the unzip package, providing unzip.exe, wouldn't that be enough?

sanqideshi: @SirLynix unzip.exe,not enough,I had already tested.

waruqi: https://github.com/xmake-io/xmake-repo/pull/4231

waruqi: And we need also add curl command for libcurl package. or add a new curl package

PR #4265, opened 2024-06-06
Use glog <0.7.0 for folly

jj683:
Folly do not support glog 0.7.0 yet:
https://github.com/facebook/folly/issues/2171


star-hengxing:
Fixed https://github.com/xmake-io/xmake-repo/pull/4228

Issue #4263, opened 2024-06-06
Unable to link boost library when asan is on


01Pollux:
### Xmake Version

v2.9.2+dev.ab97f33e9

### Operating System Version and Architecture

Windows 11 23H2

### Describe Bug

I can't seem to be able to correctly build a project that uses boost library with asan on.
When investigating it, it seems like the program gets compiled correctly but it doesnt link boost library when asan is on.

### Expected Behavior

The link command for the project seems to be like the following:

* without asan

``"C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.40.33807\\bin\\HostX64\\x64\\link.exe" -nologo -dynamicbase -nxcompat -machine:x64 -libpath:C:\Users\medra\AppData\Local\.xmake\packages\b\boost\1.85.0\49166697c65042c091ab029ed57c9f24\lib libboost_atomic-mt-s.lib libboost_filesystem-mt-s.lib -out:build\windows\x64\release\test_boost.exe build\.objs\test_boost\windows\x64\release\src\main.cpp.obj``

* with asan

``"C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.40.33807\\bin\\HostX64\\x64\\link.exe" -nologo -dynamicbase -nxcompat -machine:x64 -debug -pdb:build\windows\x64\debug\test_boost.pdb -out:build\windows\x64\debug\test_boost.exe build\.objs\test_boost\windows\x64\debug\src\main.cpp.obj``

The expected output should've been the same as 'without asan', it seems like it's missing libpath parameters as well as boost's libraries

### Project Configuration

* xmake.lua

```LUA
add_rules("mode.debug")

set_policy("build.sanitizer.address", true)
add_requires("boost")

target("test_boost")
    set_kind("binary")
    add_files("src/main.cpp")
    add_packages("boost", {public = true, inherit = true})
```

* main.cpp
```CPP
#include <iostream>
#include <boost/filesystem.hpp>
int main() {
    boost::filesystem::path path("/path/to/directory");
    if (boost::filesystem::exists(path)) {
        std::cout << "Directory exists" << std::endl;
    } else {
        std::cout << "Directory does not exist" << std::endl;
    }
    return 0;
}
```

### Additional Information and Error Logs

The following files are the result of running `xmake project -k vsxmake -yvD` command.
[...]

PR #4271, opened 2024-06-06
boost: improve links

star-hengxing:
fix https://github.com/xmake-io/xmake-repo/issues/4263

Windows platform doesn't need to deal with link order

### DIFF
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -284,8 +284,11 @@ package("boost")
         else
             table.insert(argv, "address-model=32")
         end
-        local cxxflags
-        local linkflags
+
+        local cxxflags = {}
+        local linkflags = {}
+        table.join2(cxxflags, table.wrap(package:config("cxflags")))
+        table.join2(cxxflags, table.wrap(package:config("cxxflags")))
         if package:is_plat("windows") then
             local vs_runtime = package:config("vs_runtime")
             if package:config("shared") then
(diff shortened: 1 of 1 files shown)


SirLynix:
I don't think this is the right way to fix it, if the name are different then we should adapt the way we compute them

SirLynix:
> If boost breaks the lib name in one day, this will make the package description more complex

We will need to do it anyway as we need them on other platforms

> And most xrepo packages don't need to deal with link order, What's so special about the boost library?

Nothing, all packages providing multiple libs depending on each other (A.so linked to B.so) need explicit link order.

waruqi:
> If boost breaks the lib name in one day, this will make the package description more complex. And most xrepo packages don't need to deal with link order, What's so special about the boost library?

It doesn't simplify the package configuration of boost, as we still need to deal with the linking order on non-windows.

star-hengxing:
I'm wondering if ld linker is the only one that needs to deal with link order. If so, then we don't need to waste too much time dealing with the some corner cases on links

https://github.com/xmake-io/xmake-repo/blob/129c8ccf8b9ba655042f927bfeda1030c4186b7c/packages/b/boost/xmake.lua#L97-L110

SirLynix:
I think we should restore link order on Windows, as far as I know only MSVC linker doesn't care about link order, but not only MSVC linker can be used. You can compile using Clang/LLD for example.

star-hengxing:
> I think we should restore link order on Windows, as far as I know only MSVC linker doesn't care about link order, but not only MSVC linker can be used. You can compile using Clang/LLD for example.

Can we use `package:has_tool`?
```lua
if not package:has_tool("ld", "linker") then
    -- link order
end
```

SirLynix:
It wont help, we better just get the right name in all case and add links everytime imo

star-hengxing:
Thanks for your reply. So we need to improve boost on_test.

PR #4292, opened 2024-06-08
add nolzma and nozstd config option for boost.iostreams

xiongyang:
add nolzma and nozstd config option
default is true  which will disable using zstd and xz
 if set as false it will add_deps xz/zstd then complie boost with the deps

### DIFF
--- a/packages/b/boost/xmake.lua
+++ b/packages/b/boost/xmake.lua
@@ -78,6 +78,8 @@ package("boost")
     for _, libname in ipairs(libnames) do
         add_configs(libname,    { description = "Enable " .. libname .. " library.", default = (libname == "filesystem"), type = "boolean"})
     end
+    add_configs("nozstd", {description = "disable zstd for iostreams", default = true, type = "boolean"})
+    add_configs("nolzma", {description = "disable lzma for iostreams", default = true, type = "boolean"})
 
     on_load(function (package)
         local function get_linkname(package, libname)
@@ -138,6 +140,14 @@ package("boost")
                 package:add("defines", "BOOST_PYTHON_STATIC_LIB")
             end
             package:add("deps", "python " .. package:config("pyver") .. ".x", {configs = {headeronly = true}})
+        end
+        if  package:is_plat("linux") then
+            if not package:config("nozstd") then
+                package:add("deps", "zstd")
+            end
+            if not package:config("nolzma") then
+                package:add("deps", "xz")
+            end
         end
     end)
 
@@ -242,9 +252,21 @@ package("boost")
             end
         end
 
+        local function config_deppath(file,depName,ruleName)
+                local dep = package:dep(depName)
+                local info = dep:fetch({external = false})
+                local usStr = format("\nusing %s : : <include>\"%s\" <search>\"%s\" ;",ruleName, info.includedirs[1], info.linkdirs[1])              
+                file:write(usStr)
+        end
         local file = io.open("user-config.jam", "w")
         if file then
             file:write(get_compiler(package, build_toolchain))
+            if not package:config("nolzma") then
+                config_deppath(file, "xz", "lzma")
+            end
+            if not package:config("nozstd") then
+                config_deppath(file, "zstd", "zstd")
+            end
             file:close()
         end
         os.vrun("./b2 headers")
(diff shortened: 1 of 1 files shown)


packages/b/boost/xmake.lua
```diff
@@ -78,6 +78,8 @@ package("boost")
     for _, libname in ipairs(libnames) do
         add_configs(libname,    { description = "Enable " .. libname .. " library.", default = (libname == "filesystem"), type = "boolean"})
     end
+    add_configs("nozstd", {description = "disable zstd for iostreams", default = true, type = "boolean"})
```
waruqi, line 81: please rename `nozstd` -> `zstd`, set the default value: false

packages/b/boost/xmake.lua
```diff
@@ -78,6 +78,8 @@ package("boost")
...
+    add_configs("nolzma", {description = "disable lzma for iostreams", default = true, type = "boolean"})
```
waruqi, line 82: and here

xiongyang: ok. all neme changed. also  update all the if statement

packages/b/boost/xmake.lua
```diff
@@ -242,9 +252,21 @@ package("boost")
             end
         end
 
+        local function config_deppath(file,depName,ruleName)
```
waruqi, line 255: add spaces ` , `, and use lower variable name

packages/b/boost/xmake.lua
```diff
@@ -242,9 +252,21 @@ package("boost")
...
+                local dep = package:dep(depName)
+                local info = dep:fetch({external = false})
```
waruqi, line 257: info maybe nil, we should add `if info then`

packages/b/boost/xmake.lua
```diff
@@ -242,9 +252,21 @@ package("boost")
...
+                local usStr = format("\nusing %s : : <include>\"%s\" <search>\"%s\" ;",ruleName, info.includedirs[1], info.linkdirs[1])
```
waruqi, line 258: use lowercase, and includedirs may be string instead of table.

packages/b/boost/xmake.lua
```diff
@@ -406,4 +434,30 @@ package("boost")
...
             ]]}, {configs = {languages = "c++14"}}))
         end
+
+        if package:config("iostreams") then
+            if package:config("zstd") then
+                assert(package:check_cxxsnippets({test = [[
+                #include <boost/iostreams/filter/zstd.hpp>
+                #include <boost/iostreams/filtering_stream.hpp>
+                int main(int argc, char* argv[]) {
+                boost::iostreams::filtering_ostream out;
```
waruqi, line 444: incorrect indent

packages/b/boost/xmake.lua
```diff
@@ -406,4 +434,30 @@ package("boost")
                 }
             ]]}, {configs = {languages = "c++14"}}))
         end
+
+        if package:config("iostreams") then
+            if package:config("zstd") then
+                assert(package:check_cxxsnippets({test = [[
+                #include <boost/iostreams/filter/zstd.hpp>
+                #include <boost/iostreams/filtering_stream.hpp>
+                int main(int argc, char* argv[]) {
```
waruqi, line 443: please use `void test()` instead of `main`

packages/b/boost/xmake.lua
```diff
@@ -270,6 +290,14 @@ package("boost")
             "runtime-debugging=" .. (package:is_debug() and "on" or "off")
         }
 
+        if not package:config("lzma") then
+            table.insert(argv,"-sNO_LZMA=1")
+        end
+        if not package:config("zstd") then
+            table.insert(argv,"-sNO_ZSTD=1")
+        end
+
```
waruqi, line 299: remove an empty line

packages/b/boost/xmake.lua
```diff
@@ -270,6 +290,14 @@ package("boost")
             "runtime-debugging=" .. (package:is_debug() and "on" or "off")
         }
 
+        if not package:config("lzma") then
+            table.insert(argv,"-sNO_LZMA=1")
```
waruqi, line 294: add spaces, ` , `

packages/b/boost/xmake.lua
xiongyang: fix the style and format is done.
includedirs is table with only one element for xz and zstd

PR #4327, opened 2024-06-10
chipmunk2d: support bsd

Martinfx:


waruqi:
In order to solve ci problems more quickly and easily, please update only one package at a time per pr.

Martinfx:
> In order to solve ci problems more quickly and easily, please update only one package at a time per pr.

ok

PR #4331, opened 2024-06-11
openal-soft: support bsd

Martinfx:


SirLynix:
OpenAL builds fine on FreeBSD but linking doesn't work. I'm not sure how to fix that.
https://github.com/llvm/llvm-project/issues/30308

PR #4380, opened 2024-06-16
I don't know anymore

maximegmd:
I mean, at some point, things need to stop randomly breaking and requiring nonsense solutions like checking if a function exists.

### DIFF
--- a/packages/l/libuv/xmake.lua
+++ b/packages/l/libuv/xmake.lua
@@ -33,13 +33,15 @@ package("libuv")
     end
 
     -- https://github.com/libuv/libuv/issues/3411
-    on_check("android", function (package)
-        if package:version():ge("1.47.0") then
-            local ndk = package:toolchain("ndk")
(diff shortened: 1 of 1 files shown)


waruqi:
you can also try to update xmake to 2.9.2

SirLynix:
I agree with @maximegmd, this is an oversight, this function is too recent and should be behind a check, however you can also use lock to freeze xmake-repo update if required.

Maybe we should be able to set a minimal xmake version to make this more explicit

qudix:
Like using `set_xmakever' in a package?

SirLynix:
> Like using `set_xmakever' in a package?

Something like that yes, but it won't prevent us from checking if function exist.

Or we need to set xmakever to the whole xmake repository, as it was before.

Something we can do is add versions and branches on the repository, for example we could use the `v1` branch and make xmake target this branch.

When we decide to update the minimal xmake version, we create another branch (v2) and update xmake to target this branch.

This way older xmake versions still download and use previous compatible xmake-repo scripts.

PR #4416, opened 2024-06-20
slang package improvements

exdal:
Extension of my initial slang pull request, this one directly pulls precompiled binaries from release instead of rebuilding it. Massively helps the compile time


waruqi:
I don't recommend precompiled binaries for linux, there are binary compatibility issues and toolchain switching is not supported.

Also, xmake will automatically build the precompiled binary on windows and makes it available to the user, so it's not necessary here either.

PR #4418, opened 2024-06-20
cdt: add package

Chi-EEE:


star-hengxing:
Use [cmake](https://github.com/artem-ogre/CDT/blob/master/CDT%2FCMakeLists.txt) is better than port to xmake

PR #4443, opened 2024-06-23
try enable more feature of boost

### LINKED ISSUE #4227: boost 1.85 add new lib Charconv
### Is your feature request related to a problem? Please describe.

https://www.boost.org/users/history/version_1_85_0.html

### Describe the solution you'd like

https://www.boost.org/users/history/version_1_85_0.html

### Describe alternatives you've considered

https://www.boost.org/users/history/version_1_85_0.html

### Additional context

https://www.boost.org/users/history/version_1_85_0.html

c8ef:
Fixes #4227


evanzd:
this PR breaks our CI that using 1.81.0 since new libs like `charconv` and `cobalt` were introduced after [1.84.0](https://www.boost.org/users/history/): 
```
error: wrong library name 'charconv' in the --with-<library> option.
````

One possible workaround is to replace `configs = {all = true}` by the required library names in `xmake.lua`. However, every time a new package is added as a dependency, xmake will download and build the entire boost lib. Any advice on this issue?
 `add_requires("boost 1.81.0", {system = false, configs = {all = true, multi = true}})`

**Comment by evanzd** [comment]
Is it possible to ensure that the libnames are compatible across different Boost versions? Inspired by [this line](https://github.com/xmake-io/xmake-repo/blob/00fc015b484a9ee0a3035e7f81f3b0f9b9c5e5ef/packages/b/boost/xmake.lua#L389), i think something like below should work (not tested)
```lua
        if package:version():le("1.83.0") then
            table.insert(libnames, "url")
        if package:version():le("1.84.0") then
            table.insert(libnames, "cobalt")
        if package:version():le("1.85.0") then
            table.insert(libnames, "charconv")
```

PR #4504, opened 2024-06-28
Update srpc to v0.10.3

holmes1412:
Add new version v0.10.3 and add dependencies abseil for checks.


waruqi:
https://github.com/xmake-io/xmake-repo/pull/4531

PR #4512, opened 2024-06-29
Update re2 to 2024.06.01

heheda123123:


packages/r/re2/xmake.lua
```diff
@@ -21,6 +21,13 @@ package("re2")
         add_syslinks("pthread")
     end
 
+    on_load(function (package)
+        local version = package:version()
+        if version:eq("2024.06.01") and is_plat("mingw") then
```
waruqi, line 26: package:is_plat

PR #4533, opened 2024-07-01
fmt: add 11.0.0 version

star-hengxing:


packages/f/fmt/xmake.lua
```diff
@@ -34,6 +35,10 @@ package("fmt")
         add_extsources("brew::fmt")
     end
 
+    if is_plat("windows") then
+        add_cxxflags("/utf-8")
```
SirLynix, line 39: I think this should be a config, controlling `FMT_UNICODE` cmake option

PR #4505, opened 2024-06-28
add node-addon-api package for linux

enzalito:
I wrote this package for [node-addon-api](https://github.com/nodejs/node-addon-api/tree/main) as the project relies on their own build system (node-gyp) that I didn't want to use.  
It only supports Linux for now as the `.gyp` files on their repository seem to indicate that platform specific steps are needed for Windows and Macos.

PR #4554, opened 2024-07-03
add nodejs package for linux

enzalito:
Needed by https://github.com/xmake-io/xmake-repo/pull/4505 (https://github.com/xmake-io/xmake-repo/pull/4505#discussion_r1658956000)

### DIFF
--- a/packages/n/nodejs/xmake.lua
+++ b/packages/n/nodejs/xmake.lua
@@ -1,0 +1,15 @@
+package("nodejs")
+
+    set_homepage("https://nodejs.org/")
+    set_description("Cross-platform JavaScript runtime environment.")
+
+    on_fetch("linux", function (package, opt)
+        if opt.system then
+            import("lib.detect.find_path")
+
+            local headers_path = find_path("node.h", { "/usr/**", "/usr/local/**"})
+            if headers_path then
+                return {includedirs = {headers_path}}
+            end
+        end
+    end)


packages/n/nodejs/xmake.lua
```diff
@@ -0,0 +1,15 @@
+package("nodejs")
+
+    set_homepage("https://nodejs.org/")
+    set_description("Cross-platform JavaScript runtime environment.")
+
+    on_fetch("linux", function (package, opt)
+        if opt.system then
+            import("lib.detect.find_path")
+
+            local headers_path = find_path("node.h", { "/usr/**", "/usr/local/**"})
```
waruqi, line 10: We should not return `/usr/include` path, it will break for gcc.

please use find_package.

https://github.com/xmake-io/xmake-repo/blob/12301519c42acd15ba290de1b6cc00cae61d1e00/packages/l/libiconv/xmake.lua#L28

enzalito: ok no problem, can you explain why ? it seems to work fine for me with gcc-13

enzalito: oh is it because the linux headers have file named `node.h` too ?

enzalito: i was only able to find the package with `package:find_package("apt::nodejs")` (i'm on ubuntu). `system` and `pkgconfig` did not work.  
is it ok if i use this: `return package:find_package("apt::nodejs") or package:find_package("pacman::nodejs")` ?

enzalito: `package:find_tool("nodejs")` seems to work too, not sure which one is preferable

enzalito: i'm using local headers_path = find_path("node_api.h", { "/usr/**", "/usr/local/**"}) in my local package for now and everything seems to be working fine.

waruqi: it will pass `-isystem /usr/include` or `-isystem /usr/local/include` to gcc, and it will break gcc.

This is a known issue with gcc. If `node_api.h` is in /usr/include and /usr/local/include, gcc will find it even if you don't add `-isystem /usr/include`.

https://github.com/xmake-io/xmake/issues/4596#issuecomment-1886711974
https://github.com/msys2/MINGW-packages/issues/10761
https://github.com/pcb2gcode/pcb2gcode/issues/587

The `package:find_package("system::xx")` automatically handles this and avoids adding unnecessary `-system /usr/include`. like this https://github.com/xmake-io/xmake-repo/blob/12301519c42acd15ba290de1b6cc00cae61d1e00/packages/l/libiconv/xmake.lua#L28

enzalito: Ok thank you for the explanation.

The headers are located in `/usr/include/node/` so that's why it is working for me.

I still not sure why `package:find_package("system:nodejs")` doesn't find anything but I dug a little more and according to https://github.com/nodejs/node-addon-api/issues/882, there is no way to find the path to the headers that come packaged with `nodejs` in a platform agnostic way.

However, this issue indirectly links to https://github.com/nodejs/node-api-headers which seems to contain exactly what I need for `node-addon-api`. So I think I should probably create a package from this repo instead.

I'll wait for your feedback before creating yet another PR

waruqi: > So I think I should probably create a package from this repo instead.node-addon-api

you can try.

> The headers are located in so that's why it is working for me./usr/include/node/
> I still not sure why doesn't find anything but I dug a little more and according to https://github.com/nodejs/node-addon-api/issues/882, there is no way to find the path to the headers that come packaged with in a platform agnostic way.package:find_package("system:nodejs")nodejs

try

```lua
package:find_package("system:nodejs", {includes = "node/node_api.h"})
```

enzalito: > try
> 
> ```lua
> package:find_package("system:nodejs", {includes = "node/node_api.h"})
> ```

Still nothing :/

I have made a `node-api-headers` package that works perfectly for me. It as the advantage of being way lighter as it doesn't contain the whole Node runtime + ecosystem but just the headers that I need for my c++ addon. I'll make a PR right now.

PR #4625, opened 2024-07-11
aws-lc: fix mingw build

star-hengxing:


star-hengxing:
windows arm64 wait https://github.com/aws/aws-lc/pull/1697

PR #4634, opened 2024-07-12
add node_api_headers package

enzalito:
Needed by https://github.com/xmake-io/xmake-repo/pull/4505

### DIFF
--- a/packages/n/node-api-headers/xmake.lua
+++ b/packages/n/node-api-headers/xmake.lua
@@ -1,0 +1,17 @@
+package("node-api-headers")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://github.com/nodejs/node-api-headers")
+    set_description("C-based Node-API headers")
+    set_license("MIT")
+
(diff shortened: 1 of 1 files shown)


packages/n/node-api-headers/xmake.lua
```diff
@@ -0,0 +1,17 @@
...
+    set_urls("https://github.com/nodejs/node-api-headers/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/nodejs/node-api-headers.git")
+    add_versions("v1.1.0", "70608bc1e6dddce280285f3462f18a106f687c0720a4b90893e1ecd86e5a8bbf")
+
+    on_install(function(package)
+        os.cp("include/*.h", package:installdir("include"))
+    end)
+
+    on_test(function (package)
+        assert(os.isfile(path.join(package:installdir("include"), "node_api.h")))
```
waruqi, line 16: we should use `package:has_cfuncs` to test symbols and links.

enzalito: done, thanks for the feedback

PR #4636, opened 2024-07-12
add on_fetch to binutils package

enzalito:
Fixes https://github.com/xmake-io/xmake/issues/5267

### DIFF
--- a/packages/b/binutils/xmake.lua
+++ b/packages/b/binutils/xmake.lua
@@ -10,7 +10,7 @@ package("binutils")
     add_versions("2.41", "ae9a5789e23459e59606e6714723f2d3ffc31c03174191ef0d015bdf06007450")
     add_versions("2.38", "e316477a914f567eccc34d5d29785b8b0f5a10208d36bbacedcc39048ecfe024")
     add_versions("2.34", "f00b0e8803dc9bab1e2165bd568528135be734df3fabf8d0161828cd56028952")
-    
+
     if is_plat("mingw") and is_subhost("msys") then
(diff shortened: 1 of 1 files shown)


packages/b/binutils/xmake.lua
```diff
@@ -19,6 +19,28 @@ package("binutils")
         add_extsources("brew::binutils")
     end
 
+    on_fetch("@linux", "@macosx", "@msys", function (package, opt)
+        if opt.system then
+            if package:is_binary() then
+                if package:find_tool("system::ld") then
```
waruqi, line 25:
```
package:find_tool("ld")
```

packages/b/binutils/xmake.lua
```diff
@@ -19,6 +19,27 @@ package("binutils")
         add_extsources("brew::binutils")
     end
 
+    on_fetch("@linux", "@macosx", "@msys", function (package, opt)
+        if opt.system then
+            if package:is_binary() then
+                if package:find_tool("ld") then
```
waruqi, line 25: We should check a few more tools, just checking ld doesn't guarantee that the binutils are already there, and we should exclude the windows platform.

enzalito: > We should check a few more tools, just checking ld doesn't guarantee that the binutils are already there  

Ok I'll do that  

> and we should exclude the windows platform  

Do you mean mysys ? I thought it would work because it's based on Arch, plus their website references a [binutils package](https://packages.msys2.org/package/binutils).

PR #4714, opened 2024-07-22
Auto-update vulkan-utility-libraries to v1.3.291

waruqi:
New version of vulkan-utility-libraries detected (package version: v1.3.283, last github version: v1.3.291)


SirLynix:
Vulkan packages shouldn't be auto-updated

PR #4782, opened 2024-07-29
aqt: Add version 3.1.17

SirLynix:
and drop support for signed versions as they're no longer produced

### DIFF
--- a/packages/a/aqt/xmake.lua
+++ b/packages/a/aqt/xmake.lua
@@ -4,24 +4,31 @@ package("aqt")
     set_description("aqt: Another (unofficial) Qt CLI Installer on multi-platforms")
     set_license("MIT")
 
+    add_configs("shared", {description = "Download shared binaries.", default = true, type = "boolean", readonly = true})
+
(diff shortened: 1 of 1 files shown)


SirLynix:
macOS fails with
> cannot runv(aqt -h), Not access because it is busy
what could cause this error?

waruqi:
> { 
  "aqt 3.1.17",
  "aqt 3.1.6",
  "aqt 3.0.2",
  "aqt 2.0.6", 
}

four aqt test processes?

SirLynix:
that's because I removed the signed prefix on Windows, it shouldn't be a big deal except on macOS where those versions don't exist.

however I have a busy error on the existing version

SirLynix:
After digging a bit it's a EACCES error, caused by a missing executable permission. how can I chmod +x using xmake?

waruqi:
os.exec + chmod https://github.com/xmake-io/xmake-repo/blob/c05eb260a8ff354b6226cdbbbb74c7e6ab313718/packages/b/bazel/xmake.lua#L36

Issue #4806, opened 2024-07-31
fmtlog broken now😢 


urasssopretty:
### Xmake Version

2.9.4

### Operating System Version and Architecture

x64 Windows 10, Windows 11, Arch Linux

### Describe Bug

A project that was successfully builed half a year ago is not being builed now. I also tested fmtlot in an empty project. I receive the message:
```
In file included from .xmake/cache/packages/2407/f/fmtlog/v2.2.1/source/fmtlog.cc:1:
.xmake/cache/packages/2407/f/fmtlog/v2.2.1/source/fmtlog-inl.h:550:5: error: no member named 'throw_format_error' in namespace 'fmt::detail'; did you mean 'fmt::throw_format_error'?
  550 |     fmt::detail::throw_format_error(err.c_str());
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     fmt::throw_format_error
.xmake/packages/f/fmt/11.0.2/24b2d6d5e23d42009b64f8ce95ee1fcc/include/fmt/base.h:737:41: note: 'fmt::throw_format_error' declared here
  737 | FMT_DEPRECATED FMT_NORETURN inline void throw_format_error(
      |     
```

Well, let's say fmtlog has a new version, but xrepo packages use a fixed version, so why it broken😢 (udp: omg fmtlog use fmt 10.2.1 now, it's discouraging)

### Expected Behavior

fmtlog that works out of the box

### Project Configuration

```
local dependencies = { "fmtlog" }
for _, dependency in ipairs(dependencies) do
    add_requires(dependency)
end

target("fucking")
    set_default(true)
    set_kind("binary")
    for _, dependency in ipairs(dependencies) do
        add_packages(dependency)
    end
    add_files("**.cc")
target_end()
```

### Additional Information and Error Logs

$ xmake -y
checking for platform ... linux
checking for architecture ... x86_64
  => install fmtlog v2.2.1 .. failed

[ 25%] Building CXX object CMakeFiles/fmtlog-static.dir/fmtlog.cc.o
[ 50%] Building CXX object CMakeFiles/fmtlog-shared.dir/fmtlog.cc.o
In file included from /home/me/.xmake/cache/packages/2407/f/fmtlog/v2.2.1/source/fmtlog.cc:1:
In file included from /home/me/.xmake/cache/packages/2407/f/fmtlog/v2.2.1/source/fmtlog-inl.h:24:
[...]

PR #4808, opened 2024-07-31
fmtlog: set fmt version to 10

Chi-EEE:
https://github.com/xmake-io/xmake-repo/issues/4806

### DIFF
--- a/packages/f/fmtlog/xmake.lua
+++ b/packages/f/fmtlog/xmake.lua
@@ -8,7 +8,7 @@ package("fmtlog")
     add_versions("v2.2.1", "9bc2f1ea37eece0f4807689962b529d2d4fa07654baef184f051319b4eac9304")
     add_versions("v2.1.2", "d286184e04c3c3286417873dd2feac524c53babc6cd60f10179aa5b10416ead7")
 
-    add_deps("cmake", "fmt >=10")
+    add_deps("cmake", "fmt 10")
     if is_plat("linux") then
         add_syslinks("pthread")
     end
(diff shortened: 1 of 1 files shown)


packages/f/fmtlog/xmake.lua
```diff
@@ -8,7 +8,7 @@ package("fmtlog")
     add_versions("v2.2.1", "9bc2f1ea37eece0f4807689962b529d2d4fa07654baef184f051319b4eac9304")
     add_versions("v2.1.2", "d286184e04c3c3286417873dd2feac524c53babc6cd60f10179aa5b10416ead7")
 
-    add_deps("cmake", "fmt")
+    add_deps("cmake", "fmt 10.2.1")
```
SirLynix, line 11: isn't `fmt 10` enough? this allows for minor and patch version mismatch

Chi-EEE: yes that works, thanks

packages/f/fmtlog/xmake.lua
```diff
@@ -8,7 +8,7 @@ package("fmtlog")
     add_versions("v2.2.1", "9bc2f1ea37eece0f4807689962b529d2d4fa07654baef184f051319b4eac9304")
     add_versions("v2.1.2", "d286184e04c3c3286417873dd2feac524c53babc6cd60f10179aa5b10416ead7")
 
-    add_deps("cmake", "fmt")
+    add_deps("cmake", "fmt 10")
```
waruqi, line 11: why not `>=10`?

SirLynix: this would allow for fmt 11 which won't compile here, I'm making a better fix

star-hengxing: I think `fmt <11` is better.

SirLynix: the issue is that the library uses unstable API from fmt (fmt::detail) and will be updated to support fmt 11 in the future, we need to match the package version with a certain fmt version

Issue #4823, opened 2024-08-02
DPP: De-const have_voice option and add coroutines support.


Nk125:
### Is your feature request related to a problem? Please describe.

In DPP, the voice module is an optional feature, while in the current script is read-only, this could leverage to bloated, unwanted and larger compile times/binary sizes.

Also, there's another optional feature that's not covered by the current script, that is coroutines, the attempt to include this before in a pull request (see #1745) was closed after a while due to inconsistent code.

### Describe the solution you'd like

A non-readonly `have_voice` option, with also a new option to enable coroutines in the library.

### Describe alternatives you've considered

Optionally enabled voice feature: There's no local workaround to disable this optional feature, unless writing the whole library xmake script again with the readonly property set to false.

New Coro option: The first alternative I could see, was to hand-write the required definitions through [`add_defines()`](https://xmake.io/#/manual/project_target?id=targetadd_defines) function, but inside the [xmake port script](https://github.com/xmake-io/xmake-repo/blob/bfd2b847f0b006b17aae9fdac2064e2039739cfb/packages/d/dpp/port/xmake.lua#L7), the C++17 standard was hardcoded, therefore leading to compile errors due to incompatibility of C++17 to coroutines feature (released for C++20 standard).

### Additional context

I'm making a pull request to add the features described here.

Nk125:
Merged related pull request #4824

PR #4824, opened 2024-08-02
DPP: De-const have_voice option and add coroutines support.

Nk125:
* Before adding new features and new modules, please go to issues to submit the relevant feature description first. [✔️]
  * See #4823 
* Write good commit messages and use the same coding conventions as the rest of the project. [✔️]
  *  Across the whole pull request conventional commits are used.
* Please commit code to dev branch and we will merge into master branch in feature [✔️]
* Ensure your edited codes with four spaces instead of TAB. [✔️]
  * The EditorConfig plugin for Visual Studio Code managed these based in the public [EditorConfig file in this repo](https://github.com/xmake-io/xmake-repo/blob/dev/.editorconfig).

### DIFF
--- a/packages/d/dpp/port/xmake.lua
+++ b/packages/d/dpp/port/xmake.lua
@@ -12,7 +12,6 @@ end
 
 target("dpp")
     set_kind("$(kind)")
-    set_languages("c++17")
     add_includedirs("include", "include/dpp")
     add_headerfiles("include/(dpp/**.h)")
     add_files("src/dpp/**.cpp")
@@ -24,9 +23,15 @@ target("dpp")
     end
 
     if has_config("coro") then
-        set_languages("c++20")
         add_defines("DPP_CORO")
     end
+
+    local target_cpp_lang = "c++17"
+    if has_config("coro") then
+        target_cpp_lang = "c++20"
+    end
+
+    set_languages(target_cpp_lang)
 
     add_defines("DPP_BUILD", "DPP_USE_EXTERNAL_JSON")
 
--- a/packages/d/dpp/port/xmake.lua
+++ b/packages/d/dpp/port/xmake.lua
@@ -1,6 +1,12 @@
 add_rules("mode.debug", "mode.release")
 
-add_requires("fmt", "nlohmann_json", "libsodium", "libopus", "openssl", "zlib")
+add_requires("fmt", "nlohmann_json", "openssl", "zlib")
+
+option("have_voice", {default = true})
+
+if has_config("have_voice") then
+    add_requires("libopus", "libsodium")
+end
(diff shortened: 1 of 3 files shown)


Nk125:
Before committing, the changes were tested using the test.lua script in this repo with my machine (Windows 10 21H1 x86_64).

packages/d/dpp/xmake.lua
```diff
@@ -53,7 +53,9 @@ package("dpp")
 
     add_deps("nlohmann_json", "openssl", "zlib")
 
-    add_configs("have_voice", { description = "Enable voice support for the library.", default = true, type = "boolean" , readonly = true})
+    add_configs("have_voice", { description = "Enable voice support for the library.", default = true, type = "boolean" , readonly = false})
```
waruqi, line 56: please rename `have_voice` to `voice`, we need not any have_  with_ enable_ preifx.

packages/d/dpp/port/xmake.lua
```diff
@@ -1,14 +1,32 @@
 add_rules("mode.debug", "mode.release")
 
-add_requires("fmt", "nlohmann_json", "libsodium", "libopus", "openssl", "zlib")
+add_requires("fmt", "nlohmann_json", "openssl", "zlib")
+
+option("coro", {default = false})
+
```
waruqi, line 6: remove empty line

Nk125: Resolved the issue in [this commit](https://github.com/xmake-io/xmake-repo/pull/4824/commits/371cb43eaf7abc11bbca8b2e5552821e776f5908)

packages/d/dpp/xmake.lua
Nk125: [This was changed](https://github.com/xmake-io/xmake-repo/pull/4824/commits/84b45d9f44c4a238f933fab601f7e9764c7d5dcb) but the main issue is that this change __will break__ the current user code bases, the flag was left unchanged since 2022 ([introduced here](https://github.com/xmake-io/xmake-repo/pull/1596)), the most reliable solution I was considering is showing a deprecation warning to anyone is using `have_voice` flag and then in later DPP versions erase it from the next versions.

Another solution I propose is doing a temporal alias from `have_voice` to `voice` internally in the script, and doing the same as before, show a deprecation warning and then delete the flag.

waruqi: you can reserve have_voice and add some deprecated warnings.

and add new voice config , then tell user to use voice instead of have_voice

Nk125: Added in the last commit, please review it for other advices, thanks for the support!

Issue #4866, opened 2024-08-07
Unable to compile boost library with iostream's compression


01Pollux:
### Xmake Version

v2.9.4+HEAD.e85b001f1

### Operating System Version and Architecture

Windows 11 23H2

### Describe Bug

As the title says, i get `attempt to index a nil value (local 'dep')` error when attempting to install boost with iostreams+lzma.

### Expected Behavior

The boost with iostreams+lzma library to be correctly installed.

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")
add_requires("boost", {system = false, configs = {
    iostreams = true,
    lzma = true,
}})
target("test")
    set_kind("binary")
    add_files("src/*.cpp")
    add_packages("boost")
```

### Additional Information and Error Logs

[boost.log](https://github.com/user-attachments/files/16533412/boost.log)

waruqi:
try this patch https://github.com/xmake-io/xmake-repo/pull/4868

01Pollux:
Sorry for late reply, it does compile however i get link time error when using library.
(Example taken from boost-xmake's lzma test)
```C++
#include <boost/iostreams/filter/lzma.hpp>
#include <boost/iostreams/filtering_stream.hpp>
void main() {
    boost::iostreams::filtering_ostream out;
    out.push(boost::iostreams::lzma_compressor());
}
```
Produces the following:
```
1>EXEC : error : main.cpp.obj : error LNK2019: unresolved external symbol "public: static void __cdecl boost::iostreams::lzma_error::check(int)" (?check@lzma_error@iostreams@boost@@SAXH@Z) referenced in function "public: bool __cdecl boost::iostreams::detail::lzma_compressor_impl<class std::allocator<char> >::filter(char const * &,char const *,char * &,char *,bool)" (?filter@?$lzma_compressor_impl@V?$allocator@D@std@@@detail@iostreams@boost@@QEAA_NAEAPEBDPEBDAEAPEADPEAD_N@Z)
[...]

PR #4880, opened 2024-08-09
Add zlib and bzip2 support to boost iostreams

01Pollux:
More details on #4866
In summary, Added zlib and bzip2 configurations similar to how it's being done with lzma and, this is done to enable the use of those libraries within boost iostreams.


star-hengxing:
Thanks

PR #4881, opened 2024-08-09
add ZmqPb v0.10.2

SFGrenade:
- it's also headeronly now

adjust repo link, since i renamed it


SFGrenade:
still doesnt like to be used, i'll put 0.10.2 in a bit after verifying that it works flawlessly

SFGrenade:
was verified with these 3 jobs:
https://github.com/SFGrenade/System_Simulator/actions/runs/10331616260
https://github.com/SFGrenade/System_Simulator/actions/runs/10331616254
https://github.com/SFGrenade/System_Simulator/actions/runs/10331616255

PR #4937, opened 2024-08-15
openfbx: add package

star-hengxing:


star-hengxing:
aarch64-linux-musl work fine in first commit, so I don't limit cross platform

PR #4938, opened 2024-08-15
re2: remove limits & add version 2024.07.02

Chi-EEE:


packages/r/re2/xmake.lua
```diff
@@ -14,26 +14,32 @@ package("re2")
...
     end)
 
-    on_install("macosx", "linux", "windows", "mingw", "cross", function (package)
-        local configs = {"-DRE2_BUILD_TESTING=OFF"}
-        table.insert(configs, "-DCMAKE_CXX_STANDARD=17")
+    on_install(function (package)
+        local configs = {
+            "-DRE2_BUILD_TESTING=OFF",
+            "-DCMAKE_CXX_STANDARD=17",
+            "-Dabsl_DIR=" .. (package:dep("abseil"):installdir("lib", "cmake", "absl")),
```
waruqi, line 36: it's installdir is an empty directory if abseil is system library.

we should use `fetch` to get linkdirs.

```lua
local absldir
local abseil = package:dep("abseil"):fetch()
if abseil then
    for _, linkdir in ipairs(abseil.linkdirs) do
        local dir = path.join(linkdir, "cmake", "absl")
        if os.isdir(dir) then
                absldir = dir
                break
        end
    end
end
```

PR #4970, opened 2024-08-17
Add support for rgb/grb config for FreeImage

01Pollux:
This is done by adding config option `rgb` to `FreeImage` for configuring between rgb and grb color formats


star-hengxing:
> Undefined symbols for architecture arm64:
  "_png_init_filter_functions_neon", referenced from:
      _png_read_filter_row in libfreeimage.a[259](pngrutil.c.o)

https://github.com/imagemin/optipng-bin/issues/97

PR #4972, opened 2024-08-18
libcbor: add package

star-hengxing:


star-hengxing:
Merge first, https://github.com/xmake-io/xmake-repo/pull/4982 requires

PR #5022, opened 2024-08-23
soci: fix deps

### LINKED ISSUE #5021: soci: Can't detect dependencies (sqlite3, mysql, postgresql, and boost)
### Xmake Version

2.9.4

### Operating System Version and Architecture

Windows 11 Home 23H2

### Describe Bug

When using soci with sqlite3 configuration it doesn't detect the library during lib compilation

### Expected Behavior

It must detect the dependency in order to copy the required headers

### Project Configuration

```lua
add_requires("soci", {configs = {sqlite3 = true, boost = true}})

target("xxx")
    set_kind("binary")
    add_packages("soci")
    ...
```

### Additional Information and Error Logs

CMake says SQLite3 and Boost weren't found

bkmeneguello:
Fixes #5021

### DIFF
--- a/packages/s/soci/xmake.lua
+++ b/packages/s/soci/xmake.lua
@@ -7,24 +7,34 @@ package("soci")
              "https://github.com/SOCI/soci.git")
 
     add_versions("v4.0.3", "4b1ff9c8545c5d802fbe06ee6cd2886630e5c03bf740e269bb625b45cf934928")
+    add_patches("v4.0.3", path.join(os.scriptdir(), "patches", "v4.0.3", "cmake_policy_fix.patch"), "6d8746c3ae39edf1b750d47dcfde97dedbe4211c2563481e877d36a9dccc556a")
 
-    add_configs("empty", {description = "Build empty backend", default = false, type = "boolean"})
-    add_configs("sqlite3", {description = "Build sqlite3 backend", default = false, type = "boolean"})
-    add_configs("db2", {description = "Build db2 backend", default = false, type = "boolean"})
-    add_configs("odbc", {description = "Build odbc backend", default = false, type = "boolean"})
-    add_configs("oracle", {description = "Build oracle backend", default = false, type = "boolean"})
-    add_configs("firebird", {description = "Build firebird backend", default = false, type = "boolean"})
-    add_configs("mysql", {description = "Build mysql backend", default = false, type = "boolean"})
-    add_configs("postgresql", {description = "Build postgresql backend", default = false, type = "boolean"})
-    add_configs("boost", {description = "Build boost backend", default = false, type = "boolean"})
+    local backends = {
+        "empty",
+        "sqlite3",
+        "db2",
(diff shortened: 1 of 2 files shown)


packages/s/soci/xmake.lua
```diff
@@ -45,10 +54,12 @@ package("soci")
...
-            }
-        ]]}, {configs = {languages = "c++14"}}))
+        local includes = {}
+        for _, pkg in ipairs(backends) do
+            if package:config(pkg) then
+                table.insert(includes, "soci/" .. pkg .. "/soci-" .. pkg .. ".h")
+            end
+        end
+        assert(package:has_cxxtypes("soci::session", {configs = {languages = "c++14"}, includes = "soci/soci.h"}))
+        assert(package:has_cxxincludes(includes, {configs = {languages = "c++14"}}))
```
waruqi, line 64: please use check_cxxsnippets.

has_cxxtypes/has_cxxincludes does not check links.

bkmeneguello: Fixed

c8ef (approved):
The current CI failure is because we are using NDK r27 to test against Android packages. To resolve this issue and make the CI happy, you could consider patching the CMakeLists to include a line containing `cmake_policy(SET CMP0057 NEW)`, or you can update the minimum required version of CMake.

Alternatively, we can merge this as it is for now. @waruqi

packages/s/soci/xmake.lua
```diff
@@ -8,24 +8,33 @@ package("soci")
...
+        "postgresql",
+    }
+    for _, backend in ipairs(backends) do
+        add_configs(backend, {description = "Build " .. backend .. " backend", default = false, type = "boolean"})
+    end
+    add_configs("boost", {description = "Enable boost integration", default = false, type = "boolean"})
+    add_configs("visibility", {description = "Enable hiding private symbol using ELF visibility if supported by the platform", default = true, type = "boolean"})
 
-    add_deps("cmake")
+    add_deps("cmake <=3.20", {private = true})
```
waruqi, line 27: why set `private = true`?

waruqi:
we can add on_check to make ci happy.

packages/s/soci/xmake.lua
c8ef, line 27: I mean patch the lower bound version at [this link](https://github.com/SOCI/soci/blob/master/CMakeLists.txt#L13), not in the xmake port file.

bkmeneguello: Because the CMake dependency is transitively propagated to my project, but it should be only necessary to build this library. Is there a better option?

bkmeneguello: I made the patch, the checksum works on windows but fails on every other platform.

SirLynix: Probably because the checksum was made on a CRLF file, turn the line endings to LF and check it again

PR #5023, opened 2024-08-24
Wren: add package

burdockcascade:
This adds the Wren scripting language to the repo.

**Links**
- https://github.com/wren-lang/wren
- https://wren.io/

**Popularity**
- Over 6k stars on Github
- over 500 forks on Github

### DIFF
--- a/packages/w/wren/xmake.lua
+++ b/packages/w/wren/xmake.lua
@@ -21,8 +21,6 @@ package("wren")
 
                 if is_mode("debug") then
                     add_defines("DEBUG")
-                elseif is_mode("release") then
-                    add_defines("NDEBUG")
                 end
 
         ]], package:config("shared") and "shared" or "static"))
--- a/packages/w/wren/xmake.lua
+++ b/packages/w/wren/xmake.lua
@@ -1,6 +1,6 @@
 package("wren")
     set_homepage("http://wren.io")
-    set_description("The Wren Programming Language. Wren is a small, fast, class-based concurrent scripting language.")
+    set_description("Wren is a small, fast, class-based concurrent scripting language.")
     set_license("MIT")
 
     add_urls("https://github.com/wren-lang/wren/archive/refs/tags/$(version).tar.gz",
--- a/packages/w/wren/xmake.lua
+++ b/packages/w/wren/xmake.lua
@@ -1,0 +1,79 @@
+package("wren")
+    set_homepage("http://wren.io")
+    set_description("The Wren Programming Language. Wren is a small, fast, class-based concurrent scripting language.")
+    set_license("MIT")
+
+    add_urls("https://github.com/wren-lang/wren/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/wren-lang/wren.git")
+
+    add_versions("0.4.0", "23c0ddeb6c67a4ed9285bded49f7c91714922c2e7bb88f42428386bf1cf7b339")
+
+    on_install(function (package)
+        local configs = {}
+        io.writefile("xmake.lua", format([[
+            local kind = "%s"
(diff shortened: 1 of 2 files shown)


packages/w/wren/xmake.lua
```diff
@@ -0,0 +1,79 @@
...
+                if is_mode("debug") then
+                    add_defines("DEBUG")
+                elseif is_mode("release") then
+                    add_defines("NDEBUG")
+                end
+
+        ]], package:config("shared") and "shared" or "static"))
+        if package:config("shared") then
+            configs.kind = "shared"
+        end
```
qudix, line 31: This isn't necessary, xmake already passes kind

packages/w/wren/xmake.lua
```diff
@@ -0,0 +1,79 @@
...
+                set_kind(kind)
+                add_headerfiles("src/include/*.h", "src/vm/*.h", "src/optional/*.h")
+                add_includedirs("src/include", "src/vm", "src/optional")
+                add_files("src/vm/*.c", "src/optional/*.c")
+
+                if is_mode("debug") then
+                    add_defines("DEBUG")
+                elseif is_mode("release") then
+                    add_defines("NDEBUG")
+                end
```
qudix, line 26: The `mode.release` rule already sets `NDEBUG`, but `mode.debug` does not set `DEBUG` which I'm unsure is common

packages/w/wren/xmake.lua
```diff
@@ -0,0 +1,79 @@
+package("wren")
+    set_homepage("http://wren.io")
+    set_description("The Wren Programming Language. Wren is a small, fast, class-based concurrent scripting language.")
```
qudix, line 3: Shorten description by removing the first sentence so it's just:
`Wren is a small, fast, class-based concurrent scripting language.`

packages/w/wren/xmake.lua
burdockcascade: fixed

packages/w/wren/xmake.lua
burdockcascade: fixed

packages/w/wren/xmake.lua
burdockcascade: fixed

packages/w/wren/xmake.lua
```diff
@@ -0,0 +1,74 @@
...
+                    add_defines("DEBUG")
+                end
+
+        ]], package:config("shared") and "shared" or "static"))
+        import("package.tools.xmake").install(package, configs)
+    end)
+
+    on_test(function (package)
+        assert(package:check_csnippets({test = [[
+            int main() {
```
c8ef, line 32: Consider using `void test` instead of `int main`.

c8ef (changes requested):

packages/w/wren/xmake.lua
```diff
@@ -0,0 +1,74 @@
...
+                    
+                    result = wrenCall(vm, 0); 
+
+                    if (result != WREN_RESULT_SUCCESS) {
+                        fprintf(stderr, "Error calling main function.\n");
+                        wrenFreeVM(vm);
+                        return 1; 
+                    }
+                } else {
+                    fprintf(stderr, "Error interpreting Wren script.\n");
```
waruqi, line 63: too large test snippet, we can simply ot. we need only to test api call and links. we can use has_cfuns to check it

burdockcascade: fixed

packages/w/wren/xmake.lua
burdockcascade: this test has now been simplified

PR #5033, opened 2024-08-24
libfido2: add package

star-hengxing:


star-hengxing:
libudev is system library(systemd) and arch linux ci already passed, so we can ignore failed ci.

PR #5043, opened 2024-08-25
jom: add mirror

ChrisCatCP:


packages/j/jom/xmake.lua
```diff
@@ -4,9 +4,10 @@ package("jom")
     set_description("the parallel make tool for Windows.")
     set_license("GPL")
 
-    add_urls("https://download.qt.io/official_releases/jom/jom_$(version).zip", {version = function (version)
-        return version:gsub("%.", "_")
-    end})
+    add_urls("https://download.qt.io/official_releases/jom/jom_$(version).zip",
+    "https://mirrors.ocf.berkeley.edu/qt/official_releases/jom/jom_$(version).zip",
+    "https://mirrors.ukfast.co.uk/sites/qt.io/official_releases/jom/jom_$(version).zip",
```
c8ef, line 9: I'm not sure if it's a good idea for package managers to handle the 'proxy' job manually for users. Instead, you could make this package available in _xmake-mirror/build-artifacts_. This way, as long as the user can access GitHub, they can access the relevant package.

ChrisCatCP: only add mirror

ChrisCatCP: I have already provided this suggestion to ruki

c8ef: Understood. Not blocking, but that would be my take on this.

PR #5046, opened 2024-08-25
Updated ObjFW - added 1.1.7

Frityet:


waruqi:
we can add on_check to make ci happy.

PR #5060, opened 2024-08-27
Auto-update kokkos-kernels to 4.4.00

waruqi:
New version of kokkos-kernels detected (package version: 4.3.01, last github version: 4.4.00)


star-hengxing:
Seem like clang-20(trunk) version break some template code, it work with my clang-18 and clang-14

PR #5063, opened 2024-08-27
vulkan-validationlayers: fixing the detection of layers at runtime

kraydan:
Fix for vulkan not finding validation layers at runtime.

I'm not sure if I did it right, but it worked for me.
Also, this is my first pull request, so I apologize if something is wrong.

### DIFF
--- a/packages/v/vulkan-validationlayers/xmake.lua
+++ b/packages/v/vulkan-validationlayers/xmake.lua
@@ -69,6 +69,9 @@ package("vulkan-validationlayers")
         if package:version():ge("1.3.275") then
             package:add("deps", "vulkan-utility-libraries " .. sdkver)
         end
+
+        package:addenv("VK_ADD_LAYER_PATH", "lib")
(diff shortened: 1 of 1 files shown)


packages/v/vulkan-validationlayers/xmake.lua
```diff
@@ -107,10 +107,16 @@ package("vulkan-validationlayers")
...
+
+        package:addenv("VK_ADD_LAYER_PATH", "lib")
+        package:mark_as_pathenv("VK_ADD_LAYER_PATH")
     end)
 
     on_install("android", function (package)
         os.cp("*", package:installdir("lib"))
+
+        package:addenv("VK_ADD_LAYER_PATH", "lib")
+        package:mark_as_pathenv("VK_ADD_LAYER_PATH")
```
waruqi, line 119: you can add it in on_load

kraydan:
Probably a strange question, but what do i need to do to get a pull request accepted?

PR #5082, opened 2024-08-29
Update box2d

ShifftC:


packages/b/box2d/xmake.lua
```diff
@@ -27,29 +63,57 @@ package("box2d")
...
+                    shapeDef.density = 1.0f;
+                    shapeDef.friction = 0.3f;
+                    b2CreatePolygonShape(bodyId, &shapeDef, &dynamicBox);
+                    float timeStep = 1.0f / 60.0f;
+                    int subStepCount = 4;
+                    for (int i = 0; i < 90; ++i)
+                    {
+                        b2World_Step(worldId, timeStep, subStepCount);
+                        b2Vec2 position = b2Body_GetPosition(bodyId);
+                        b2Rot rotation = b2Body_GetRotation(bodyId);
```
waruqi, line 87: maybe we should simply test code, we need only to test symbols links.

ShifftC: I simplified the test code a bit, we could also go back to only testing world creation.

PR #5090, opened 2024-08-30
async_simple: remove limit plat

star-hengxing:


star-hengxing:
> Alternatively, we could use the aarch64-linux-gnu toolchain that comes with Debian/Ubuntu, or we could maintain our own cutting-edge musl cross-compiling toolchain.

I agree this, And aarch64-linux-gnu toolchain more popular than musl cross-compiling toolchain.

star-hengxing:
https://github.com/xmake-io/xmake-repo/issues/5141

PR #5091, opened 2024-08-30
liblifthttp: add package

star-hengxing:


star-hengxing:
libuv/android ndk api >= 24 after v1.45.0

PR #5106, opened 2024-08-31
bgfx: fix shader compilation

### LINKED ISSUE #5105: bgfx shaders rules not working
### Xmake Version

2.9.4

### Operating System Version and Architecture

Windows 11 Home 23H2

### Describe Bug

When building with bgfx and using add_rules("@bgfx/shaders"), after adding files with the expected suffixes (.vert and .frag), the compilation ignores these files.

### Expected Behavior

Shaders should be compiled by shaderc

### Project Configuration

```lua
add_requires("bgfx", "glfw", {configs = {shared = true}})

target("MyTarget")
    add_rules("@bgfx/shaders")
    set_kind("binary")
    set_languages("c++17")
    add_files("src/main.cpp")
    add_files("triangle.vert")
    add_files("triangle.frag")

    add_packages("bgfx", "glfw")
```

### Additional Information and Error Logs

```console
error: cannot runv(shadercRelease -f triangle.frag --type fragment --varyingdef .\triangle.varying.def.sc --platform windows -o build\windows\x64\debug\shaders\dx11\triangle.frag.bin --profile s_5_0), No such file or directory
```

bkmeneguello:
Fixes #5105

* Adds ".sc" as a supported extension for bgfx shaders;
* Adds support for vs_, fs, and cs_ prefixes for vertex, fragment, and compute shaders;
* Adds support for bin2c confguration (header generation)

### DIFF
--- a/packages/b/bgfx/rules/shaders.lua
+++ b/packages/b/bgfx/rules/shaders.lua
@@ -4,9 +4,10 @@
 --
 -- add_rules("@bgfx/shaders")
 -- add_files("shader.vert", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.bin", profiles = {glsl = "330"}})
+-- add_files("vs_shader.sc", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.h", header = true, profiles = {glsl = "330"}})
 
 rule("shaders")
-    set_extensions(".vert", ".frag", ".comp")
+    set_extensions(".sc", ".vert", ".frag", ".comp")
     on_buildcmd_file(function (target, batchcmds, shaderfile, opt)
         import("lib.detect.find_program")
         import("core.base.option")
@@ -24,7 +25,13 @@ rule("shaders")
         if fileconfig and fileconfig.output_name then
             output_filename = fileconfig.output_name
         else
-            output_filename = path.filename(shaderfile) .. ".bin"
+            local filename = path.filename(shaderfile)
+            output_filename = filename:match("^(.*)%.sc$") or filename
+            if fileconfig and fileconfig.header then
+                output_filename = output_filename .. ".h"
+            else
+                output_filename = output_filename .. ".bin"
+            end
         end
 
         local output_dir
@@ -38,26 +45,7 @@ rule("shaders")
         if fileconfig and fileconfig.vardef then
             vardef_filename = fileconfig.vardef
         else
-            vardef_filename = path.join(
-                path.directory(shaderfile),
-                path.basename(shaderfile) .. ".varying.def.sc")
-        end
-
-        local shader_type
-        if fileconfig and fileconfig.type then
-            if table.contains(bgfx_types, fileconfig.type) then
-                shader_type = fileconfig.type
-            else
-                raise("unsupported shader type " .. fileconfig.type)
-            end
-        elseif shaderfile:match("%.vert$") then
-            shader_type = "vertex"
-        elseif shaderfile:match("%.frag$") then
-            shader_type = "fragment"
-        elseif shaderfile:match("%.comp$") then
-            shader_type = "compute"
-        else
-            raise("cannot determine shader type from file name " .. path.filename(shaderfile))
+            vardef_filename = path.join(path.directory(shaderfile), "varying.def.sc")
         end
 
         -- determine platform-specific shaderc arguments
@@ -73,8 +61,8 @@ rule("shaders")
         }
         local bgfx_default_profiles = {
             windows = {
-                vertex = {dx9 = "s_3_0", dx11 = "s_5_0", glsl = "120"},
-                fragment = {dx9 = "s_3_0", dx11 = "s_5_0", glsl = "120"},
+                vertex = {dx11 = "s_5_0", glsl = "120"},
+                fragment = {dx11 = "s_5_0", glsl = "120"},
                 compute = {dx11 = "s_5_0", glsl = "430"},
             },
             macosx = {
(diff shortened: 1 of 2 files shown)


packages/b/bgfx/xmake.lua
```diff
@@ -47,7 +47,7 @@ package("bgfx")
         local bimgdir = package:resourcefile("bimg")
         local genie = is_host("windows") and "genie.exe" or "genie"
         local args = {}
-        if is_plat("windows|native", "macosx", "linux") then
+        if is_plat("windows", "macosx", "linux") then
```
waruqi, line 50: use `package:is_plat`

packages/b/bgfx/rules/shaders.lua
```diff
@@ -24,7 +25,13 @@ rule("shaders")
         if fileconfig and fileconfig.output_name then
             output_filename = fileconfig.output_name
         else
-            output_filename = path.filename(shaderfile) .. ".bin"
+            local filename = path.filename(shaderfile)
+            output_filename = filename:match("^(.*)%.sc$") or filename
```
xq114, line 29: `path.basename` can also remove the extension. Btw, is it necessary to remove the .sc extension? when there is shader.vert  and shader.vert.sc, it can cause conflicts.

packages/b/bgfx/rules/shaders.lua
```diff
@@ -73,8 +61,8 @@ rule("shaders")
         }
         local bgfx_default_profiles = {
             windows = {
-                vertex = {dx9 = "s_3_0", dx11 = "s_5_0", glsl = "120"},
-                fragment = {dx9 = "s_3_0", dx11 = "s_5_0", glsl = "120"},
```
xq114, line 77: Why removing dx9 support? In xmake-repo old versions of bgfx are also available

packages/b/bgfx/rules/shaders.lua
```diff
@@ -38,26 +45,7 @@ rule("shaders")
...
-            end
-        elseif shaderfile:match("%.vert$") then
-            shader_type = "vertex"
-        elseif shaderfile:match("%.frag$") then
-            shader_type = "fragment"
-        elseif shaderfile:match("%.comp$") then
-            shader_type = "compute"
-        else
-            raise("cannot determine shader type from file name " .. path.filename(shaderfile))
+            vardef_filename = path.join(path.directory(shaderfile), "varying.def.sc")
```
xq114, line 48: Will it conflict when compiling multiple shaders in parallel?

packages/b/bgfx/rules/shaders.lua
```diff
@@ -89,13 +77,39 @@ rule("shaders")
...
             "-f", shaderfile,
             "--type", shader_type,
             "--varyingdef", vardef_filename,
             "--platform", bgfx_platforms[target:plat()],
         }
+
+        if fileconfig and fileconfig.header then
+            table.insert(args, "--bin2c")
+            if fileconfig.header ~= true then
+                table.insert(args, fileconfig.header)
```
xq114, line 108: I think there should be more info about `.header` argument in the comments at the beginning of this file, e.g. What would happen if both `-o <output>` and `--bin2c <header>` are specified? Which of the `<output>` and the `<header>` will be generated? Is `output_name = "shader.vert.h"` really correct?

packages/b/bgfx/rules/shaders.lua
bkmeneguello: I removed the ".sc" extension because it's the source code extension (extensively used in the project sources), but only if it is set, also that's why I used the match function instead of basename, because this way I can remove the ".sc" extension only if it's used. If we have the situation you mentioned both will generate the same output file, indeed. The transformation I used follows the bgfx example files, and the output_name can be used anyway.

packages/b/bgfx/rules/shaders.lua
bkmeneguello: I didn't remove the dx9 support, I only removed it from the default profiles. When the project doesn't specify the bgfx version; by default, it uses the latest (currently 8752) which doesn't have the RendererType::Direct3D9.

packages/b/bgfx/rules/shaders.lua
bkmeneguello: I followed the [shaderc behavior](https://github.com/bkaradzic/bgfx/blob/eaec8ddf3b4c5a11e60cde42728439293721ec8d/tools/shaderc/shaderc.cpp#L2827) which uses the file varying.def.sc in the shader directory.

packages/b/bgfx/rules/shaders.lua
bkmeneguello: The `--bin2c` option generates a header file containing the compiled shader to be included in the project instead of loaded in runtime. It the output_name is defined, the ".h" suffix is recommended, but it just generates a header file. The "<header>" string is optional and, if provided it's used as the variable containing the shader data, if not defined it uses the input file base name. Since not all rule configuration values follow strictly the shaderc inputs I didn't follow also and used a more explicit name. What is the best way to document this rule, if it's necessary to read the source anyway, is that worth documenting it better?

xq114: Of course it's better documenting the rule on some website like xrepo.xmake.io, but it's not yet implemented, so as for now the only way to document it is to do so in the rule file. Seems that when there is `--bin2c <header>`, the variable name in the generated file will be `<header>`? If so, I'd suggest a different name for this config, e.g. `export_var` or `array_name`(align with https://bkaradzic.github.io/bgfx/tools.html#shader-compiler-shaderc ) or just `array` to be more clear.

xq114: `array_name=true` can become an advanced feature (deduce by extension .h is usually enough), so it's ok to document in the following way:
```
-- add_rules("@bgfx/shaders")
-- generate binary file:
-- add_files("shader.vert", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.bin", profiles = {glsl = "330"}})
-- generate header file:
-- add_files("vs_shader.sc", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.h", profiles = {glsl = "330"}})
-- generate header file exporting variable "vertex_src":
-- add_files("vs_shader.sc", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.h", array_name = "vertex_src", profiles = {glsl = "330"}})
-- force to generate header file with default variable name:
-- add_files("vs_shader.sc", {type = "vertex", output_dir = "shaders", output_name = "shader.vert.inc", array_name = true, profiles = {glsl = "330"}})
```

PR #4908, opened 2024-08-12
Update luau to 0.638

SirLynix:
* Before adding new features and new modules, please go to issues to submit the relevant feature description first.
* Write good commit messages and use the same coding conventions as the rest of the project.
* Please commit code to dev branch and we will merge into master branch in feature
* Ensure your edited codes with four spaces instead of TAB.

------

* 增加新特性和新模块之前，请先到issues提交相关特性说明，经过讨论评估确认后，再进行相应的代码提交，避免做无用工作。
* 编写友好可读的提交信息，并使用与工程代码相同的代码规范，代码请用4个空格字符代替tab缩进。
* 请提交代码到dev分支，如果通过，我们会在特定时间合并到master分支上。
* 为了规范化提交日志的格式，commit消息，不要用中文，请用英文描述。

PR #5112, opened 2024-09-01
libwebsockets: add v4.3.3 version

star-hengxing:
Drop v4.1.6 version (4 years ago)


star-hengxing:
android r27 see https://github.com/xmake-io/xmake-repo/pull/4908#issuecomment-2304552906

PR #5140, opened 2024-09-05
atk: add 2.38.0 version

star-hengxing:


star-hengxing:
https://github.com/xmake-io/xmake-repo/issues/5141

PR #5225, opened 2024-09-14
add sqlite_orm version 1.9

KochabStar:
add new version

### DIFF
--- a/packages/s/sqlite_orm/xmake.lua
+++ b/packages/s/sqlite_orm/xmake.lua
@@ -5,6 +5,7 @@ package("sqlite_orm")
 
     add_urls("https://github.com/fnc12/sqlite_orm/archive/refs/tags/v$(version).zip")
     add_versions("1.8.2", "dd098fe06b46640384b77fd937b694af105dab221ab45f574e4ff9bb38bbeb90")
+    add_versions("1.9", "800f76b6a4b4d216d6d462f12cd0b62fa5057e9d2d8992f4ffd70e2b8123624b")
 
     add_deps("sqlite3")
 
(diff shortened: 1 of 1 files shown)


star-hengxing:
```
/home/runner/.xmake/packages/s/sqlite_orm/v1.9/725a69d7ed4f41639aa5fb2cdbb345da/include/sqlite_orm/sqlite_orm.h:1321:20: error: cannot initialize return object of type 'size_t' (aka 'unsigned long') with an lvalue of type '<dependent type>'
 1321 |             return Idx...[Pos];
      |                    ^~~~~~~~~~~
```

I think this is clang bug.

```c++
#if __cpp_pack_indexing >= 202311L
#define SQLITE_ORM_PACK_INDEXING_SUPPORTED
#endif

#if defined(SQLITE_ORM_PACK_INDEXING_SUPPORTED)
        /**
         *  Get the index value of an `index_sequence` at a specific position.
         */
        template<size_t Pos, size_t... Idx>
        SQLITE_ORM_CONSTEVAL size_t index_sequence_value_at(std::index_sequence<Idx...>) {
            return Idx...[Pos];
        }
#elif defined(SQLITE_ORM_FOLD_EXPRESSIONS_SUPPORTED)
        /**
         *  Get the index value of an `index_sequence` at a specific position.
         */
        template<size_t Pos, size_t... Idx>
        SQLITE_ORM_CONSTEVAL size_t index_sequence_value_at(std::index_sequence<Idx...>) {
            static_assert(Pos < sizeof...(Idx));
#ifdef SQLITE_ORM_CONSTEVAL_SUPPORTED
            size_t result;
#else
            size_t result = 0;
#endif
```

PR #5344, opened 2024-09-26
fix x264 on windows with shared build

SnowinterCat:
When x264 compiles dynamic libraries on Windows platform, it will generate libx264.dll.lib (msvc) or libx264.dll.a (mingw) import libraries.
However, configue has hardcoded -lx264 when generating x264.pc, so that libraries using pkgconfig can't find the symbols when linking x264 libraries.
Change the import library to libx264.lib (msvc) or libx264.a (mingw).

### DIFF
--- a/packages/x/x264/xmake.lua
+++ b/packages/x/x264/xmake.lua
@@ -49,7 +49,6 @@ package("x264")
     end)
 
     on_install("windows", "mingw", "linux", "macosx", "wasm", function (package)
-        io.replace("configure", "x264.dll.lib", "x264.lib", {plain = true})
         local configs = {}
         table.insert(configs, "--enable-" .. (package:config("shared") and "shared" or "static"))
         if package:is_plat("mingw") then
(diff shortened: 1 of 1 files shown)


SnowinterCat:
Although mingw64 recognises xxx.a and xxx.lib.a as import libraries, mingw32 only recognises xxx.lib.a as an import library. Therefore, only modify the generated msvc import library, not the mingw

SnowinterCat:
@star-hengxing 这个应该是没问题的，我试了删掉我的修改，发现以前的能过CI是mingw32的测试并没有执行test，仅仅是安装了mingw32。另外我试着打印x264的config.log发现x264在尝试使用-m64来测试，导致失败。出现找不到 c compiler的报错

star-hengxing:
mingw32 的问题大概是 xmake bug。
triples: host -> windows & mingw -> plat & arch == i386
但 xmake 认为不是交叉编译，没有传 `--host=` 进去导致报错。
@waruqi 

在我的 pr 中，手动传 host 进去就没问题了

PR #5419, opened 2024-10-04
PcapPlusPlus: enhance platform support and fix missing syslink on Windows

Doekin:
1. Resolves the ```undefined reference to `GetNetworkParams"``` error by adding the missing `iphlpapi` syslink, which is required for Windows.
2. Adds support for macOS, Android, and BSD platforms.


packages/p/pcapplusplus/xmake.lua
```diff
@@ -13,18 +13,18 @@ package("pcapplusplus")
 
     add_links("Pcap++", "Packet++", "Common++")
 
-    if is_plat("windows") then
-        add_syslinks("ws2_32")
+    if is_plat("windows") or is_plat("mingw") then
```
SirLynix, line 16: `if is_plat("windows", "mingw") then`

packages/p/pcapplusplus/xmake.lua
```diff
@@ -13,18 +13,18 @@ package("pcapplusplus")
...
+        add_syslinks("ws2_32", "iphlpapi")
     end
 
     add_deps("cmake")
-    if is_plat("windows") then
+    if is_plat("windows") or is_plat("mingw") then
```
SirLynix, line 21: same

Doekin: Got it.

star-hengxing:
> Additionally, PcapPlusPlus provides extensive platform support with precompiled releases available at [PcapPlusPlus Releases](https://github.com/seladb/PcapPlusPlus/releases). It is worth considering their utilization.

It's not recommend to use precompiled binary, unless you have't skill for make package.

Doekin:
Understood. In that case, could we extend support to additional platforms like macOS, FreeBSD, and Android?

star-hengxing:
> Understood. In that case, could we extend support to additional platforms like macOS, FreeBSD, and Android?

You can try, I will review.
If package deps(<= 1) is easy to deal with, You only need to patch some code, like https://github.com/xmake-io/xmake-repo/pull/5420 https://github.com/xmake-io/xmake-repo/pull/5421

Doekin:
Okay, PcapPlusPlus relies on libpcap for BSD and Android, and there have been some issues with building libpcap for these targets. I am currently working on resolving these issues.

Doekin:
Failed to build libpcap for the following targets, and I can not figure out why. Could it be an issue with the toolchain configuration?

- macOS (arm64) [My workflow](https://github.com/Doekin/xmake-repo/actions/runs/11188912860)
- Android (Windows) [My workflow](https://github.com/Doekin/xmake-repo/actions/runs/11188912849)
- Android [My workflow](https://github.com/Doekin/xmake-repo/actions/runs/11188912850)

Doekin:
Finally successfully built libpcap on MacOS, Android, and BSD, extending PcapPlusPlus support to these platforms.

PR #5429, opened 2024-10-06
Auto-update thrift to v0.21.0

waruqi:
New version of thrift detected (package version: v0.20.0, last github version: v0.21.0)


Altina-oz:
On （at least) Windows11 msvc , when compiler==True, thrift cannot be compiled with xmake successfully without boost:locale

PR #5438, opened 2024-10-07
add zmqpb v0.10.3

SFGrenade:
this fixes a runtime error i had on windows


SFGrenade:
windows arm is apparently failing, even though it has the exact same platforms listed as both cppzmq and protobuf-cpp

PR #5499, opened 2024-10-15
Auto-update mosquitto to v2.0.19

waruqi:
New version of mosquitto detected (package version: v2.0.18, last github version: v2.0.19)


apocelipes:
Close because of #9359.

PR #5503, opened 2024-10-15
update stringzilla to v3.10.0

45degree:
* update stringzilla to v3.10.0
* fix not copying stringzillia.hpp to include and build error when enable cpp

### DIFF
--- a/packages/s/stringzilla/xmake.lua
+++ b/packages/s/stringzilla/xmake.lua
@@ -56,11 +56,11 @@ package("stringzilla")
                 os.cp("include/stringzilla/stringzilla.hpp", package:installdir("include/stringzilla"))
             end
         end
-        
+
         if package:version():gt("2.0.4") then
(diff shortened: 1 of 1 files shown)


packages/s/stringzilla/xmake.lua
```diff
@@ -42,7 +45,7 @@ package("stringzilla")
     add_versions("v2.0.3", "6b52a7b4eb8383cbcf83608eaa08e5ba588a378449439b73584713a16d8920e3")
     add_versions("v1.2.2", "2e17c49965841647a1c371247f53b2f576e5fb32fe4b84a080d425b12f17703c")
 
-    on_install("android|!armeabi-v7a or !android", function (package)
+    on_install("android|!armeabi-v7a or !android and !cross", function (package)
```
waruqi, line 48: 怪怪的，后面那个 and 不用加 `()`?

45degree: 改成加()的了

PR #5506, opened 2024-10-15
add package ipaddress

Luffbee:
# ipaddress
https://github.com/VladimirShaleev/ipaddress/tree/main

A library for working and manipulating IPv4/IPv6 addresses and networks in modern C++.

This cross-platfrom header-only library (for C++11 and newer) is inspired by the [ipaddress API in Python](https://docs.python.org/3.12/library/ipaddress.html), from which it derives its name. It aims to be simpler to use due to its familiar interface. However, the C++ implementation takes a different approach: it uses static polymorphism through the strategy pattern instead of dynamic polymorphism to handle differences between IP versions (IPv4 and IPv6). This design choice eliminates the overhead of dynamic calls and virtual tables. For instance, an instance of the ipv4_address class will be represented by 4 bytes.

The library leverages modern C++ features, ensuring that all IP address and network operations support constant expressions.

### DIFF
--- a/packages/i/ipaddress/xmake.lua
+++ b/packages/i/ipaddress/xmake.lua
@@ -1,0 +1,95 @@
+package("ipaddress", function()
+    set_homepage("https://vladimirshaleev.github.io/ipaddress/")
+    set_description("A library for working and manipulating IPv4/IPv6 addresses and networks")
+    set_license("MIT")
+    set_kind("library", {
+        headeronly = true
+    })
+
+    add_urls("https://github.com/VladimirShaleev/ipaddress/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/VladimirShaleev/ipaddress.git")
+
+    add_versions("v1.1.0", "e5084d83ebd712210882eb6dac14ed1b9b71584dede523b35c6181e0a06375f1")
+
+    add_configs("noexcept", {
+        description = "Disable handling cpp exception for",
+        default = false,
+        type = "boolean"
+    })
+    add_configs("no_overload_std", {
+        description = "Do not overload std functions such as to_string, hash etc",
+        default = false,
+        type = "boolean"
+    })
+    add_configs("no_ipv6_scope", {
+        description = "Disable scope id for IPv6 addresses",
+        default = false,
+        type = "boolean"
+    })
+    add_configs("ipv6_scope_max_length", {
+        description = "Maximum scope-id length for IPv6 addresses",
+        default = 16,
+        type = "number"
+    })
+
+    add_deps("cmake")
+
+    on_install(function(package)
+        local configs = {"-DBUILD_TESTING=OFF", "-DIPADDRESS_ENABLE_CLANG_TIDY=OFF", "-DIPADDRESS_BUILD_TESTS=OFF",
+                         "-DIPADDRESS_BUILD_BENCHMARK=OFF", "-DIPADDRESS_BUILD_DOC=OFF",
+                         "-DIPADDRESS_BUILD_PACKAGES=OFF"}
+
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+
+        table.insert(configs, "-DIPADDRESS_NO_EXCEPTIONS=" .. (package:config("noexcept") and "ON" or "OFF"))
+        table.insert(configs, "-DIPADDRESS_NO_OVERLOAD_STD=" .. (package:config("no_overload_std") and "ON" or "OFF"))
+        table.insert(configs, "-DIPADDRESS_NO_IPV6_SCOPE=" .. (package:config("no_ipv6_scope") and "ON" or "OFF"))
+        table.insert(configs, "-DIPADDRESS_IPV6_SCOPE_MAX_LENGTH=" .. package:config("ipv6_scope_max_length"))
+
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_test(function(package)
+        assert(package:check_cxxsnippets({
+            test = [[
+            #include <iostream>
+
+            #include <ipaddress/ipaddress.hpp>
+
+            using namespace ipaddress;
+
+            void parse_ip_sample() {
+                constexpr auto ip = ipv6_address::parse("fec0::1ff:fe23:4567:890a%eth2");
+                constexpr auto is_site_local = ip.is_site_local();
+
+                std::cout << "ip " << ip << " is local: " << std::boolalpha << is_site_local << std::endl;
+                std::cout << "DNS PTR " << ip.reverse_pointer() << std::endl << std::endl;
+            }
+
+            void teredo_sample() {
+                constexpr auto teredo_ip = "2001:0000:4136:e378:8000:63bf:3fff:fdd2"_ipv6;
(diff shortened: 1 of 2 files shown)


SirLynix:
`ipaddress` is a very generic name, the package should probably be called `vladimirshaleev_ipaddress` (similar to what vcpkg does)

packages/i/ipaddress/xmake.lua
```diff
@@ -0,0 +1,95 @@
...
+    add_urls("https://github.com/VladimirShaleev/ipaddress/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/VladimirShaleev/ipaddress.git")
+
+    add_versions("v1.1.0", "e5084d83ebd712210882eb6dac14ed1b9b71584dede523b35c6181e0a06375f1")
+
+    add_configs("noexcept", {
+        description = "Disable handling cpp exception for",
+        default = false,
+        type = "boolean"
+    })
```
waruqi, line 18: use single line for each config

packages/i/ipaddress/xmake.lua
```diff
@@ -0,0 +1,95 @@
...
+        table.insert(configs, "-DIPADDRESS_IPV6_SCOPE_MAX_LENGTH=" .. package:config("ipv6_scope_max_length"))
+
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_test(function(package)
+        assert(package:check_cxxsnippets({
+            test = [[
+            #include <iostream>
+
```
waruqi, line 57: remove empty line

packages/i/ipaddress/xmake.lua
```diff
@@ -0,0 +1,95 @@
...
+            void subnets_sample() {
+                constexpr auto net = ipv4_network::parse("192.0.2.0/24");
+
+                std::cout << "subnets for " << net << ':' << std::endl;
+                for (const auto& subnet : net.subnets(2)) {
+                    std::cout << "  " << subnet << std::endl;
+                }
+
+                constexpr auto last_subnet = net.subnets(2).back();
```
waruqi, line 85: too large test code, please simplify it

packages/i/ipaddress/xmake.lua
```diff
@@ -0,0 +1,95 @@
+package("ipaddress", function()
+    set_homepage("https://vladimirshaleev.github.io/ipaddress/")
+    set_description("A library for working and manipulating IPv4/IPv6 addresses and networks")
+    set_license("MIT")
+    set_kind("library", {
```
waruqi, line 5: ues single line

packages/i/ipaddress/xmake.lua
Luffbee: Is there any format tool for xmake? This file is formatted by the Lua extension os vscode.

waruqi: no tool, you can see other packages

waruqi: please do not any format tool

packages/i/ipaddress/xmake.lua
Luffbee: how about now?

Luffbee:
> `ipaddress` is a very generic name, the package should probably be called `vladimirshaleev_ipaddress` (similar to what vcpkg does)

renamed to the same name as vcpkg

packages/v/vladimirshaleev-ipaddress/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+    set_license("MIT")
+    set_kind("library", {headeronly = true})
+
+    add_urls("https://github.com/VladimirShaleev/ipaddress/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/VladimirShaleev/ipaddress.git")
+
+    add_versions("v1.1.0", "e5084d83ebd712210882eb6dac14ed1b9b71584dede523b35c6181e0a06375f1")
+
+    add_configs("noexcept", {description = "Disable handling cpp exception for", default = false, type = "boolean"})
+    add_configs("no_overload_std", {description = "Do not overload std functions such as to_string, hash etc", default = false, type = "boolean"})
```
waruqi, line 13: remove `no_` prefix, default -> true

packages/v/vladimirshaleev-ipaddress/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+    add_configs("no_ipv6_scope", {description = "Disable scope id for IPv6 addresses", default = false, type = "boolean"})
```
waruqi, line 14: and here

packages/v/vladimirshaleev-ipaddress/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+    set_description("A library for working and manipulating IPv4/IPv6 addresses and networks")
+    set_license("MIT")
+    set_kind("library", {headeronly = true})
+
+    add_urls("https://github.com/VladimirShaleev/ipaddress/archive/refs/tags/$(version).tar.gz",
+        "https://github.com/VladimirShaleev/ipaddress.git")
+
+    add_versions("v1.1.0", "e5084d83ebd712210882eb6dac14ed1b9b71584dede523b35c6181e0a06375f1")
+
+    add_configs("noexcept", {description = "Disable handling cpp exception for", default = false, type = "boolean"})
```
waruqi, line 12: noexcept -> except

packages/v/vladimirshaleev-ipaddress/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+            void teredo_sample() {
+                constexpr auto teredo_ip = "2001:0000:4136:e378:8000:63bf:3fff:fdd2"_ipv6;
+                auto [server, client] = teredo_ip.teredo().value();
+                std::cout << "server: " << server << " and client: " << client << " for " << teredo_ip << std::endl << std::endl;
+            }
+        ]]
+        }, {
+            configs = {
+                languages = "c++17"
+            }
```
waruqi, line 54: single line config

packages/v/vladimirshaleev-ipaddress/xmake.lua
```diff
@@ -0,0 +1,56 @@
+package("vladimirshaleev-ipaddress")
+    set_homepage("https://vladimirshaleev.github.io/ipaddress/")
+    set_description("A library for working and manipulating IPv4/IPv6 addresses and networks")
+    set_license("MIT")
+    set_kind("library", {headeronly = true})
```
waruqi, line 5: move it to line 2

packages/v/vladimirshaleev-ipaddress/xmake.lua
Luffbee: I searched similar configs, spdlog use 'noexcept' and some others use 'exceptions'. So maybe 'exceptions' is better?

SirLynix:
I don't get it, if it's a header-only package, options are meaningless if they don't add any define to the package

Luffbee:
> > I don't get it, if it's a header-only package, options are meaningless if they don't add any define to the package
> 
> you are right, they should be exported by `package:add("defines", "xxx")`
> 
> @Luffbee
> 
> https://github.com/VladimirShaleev/ipaddress/blob/fffbd6cb64df8be25de2e3d992771a0fa57bd97c/CMakeLists.txt#L29

Sorry, I thought xmake would extract these defines from cmake. I will fix it in another PR.

P.S. Is there any document for these contribution FAQs? (e.g. format, option naming, options of headeronly package). I did some searching but couldn't find any.

PR #5573, opened 2024-10-24
gperftools: update and manual link management

Doekin:
## Manual Link Addition
By default, all available libraries are linked, including:
- tcmalloc_minimal
- tcmalloc_minimal_debug
- tcmalloc
- tcmalloc_debug
- profiler
- tcmalloc_and_profiler

While this approach doesn't result in link error during compilation, it leads to the following runtime error:
```
Check failed: !internal_init_start_has_run: Heap-check constructor called twice. Perhaps you both linked in the heap checker, and also used LD_PRELOAD to load it?
```
This issue is similar to [apache/brpc#690](https://github.com/apache/brpc/issues/690). A potential solution is to manually add links to prevent duplicate lib loading.

## Installation Policy Change
The CMakeLists file underwent significant refactoring in 2.16. Notably, it no longer uses the `install()` command.


packages/g/gperftools/xmake.lua
```diff
@@ -4,30 +4,86 @@ package("gperftools")
...
-        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
         table.insert(configs, "-DGPERFTOOLS_BUILD_STATIC=" .. (package:config("shared") and "OFF" or "ON"))
+        table.insert(configs, "-Dgperftools_build_minimal=" .. (package:config("minimal") and "ON" or "OFF"))
         if package:is_plat("linux") then
             table.insert(configs, "-Dgperftools_enable_libunwind=" .. (package:config("unwind") and "ON" or "OFF"))
         end
-        import("package.tools.cmake").install(package, configs)
+
+        io.replace("CMakeLists.txt", "if(MINGW OR MSVC)", "if(MINGW OR MSVC)\nlink_libraries(psapi synchronization shlwapi)", {plain = true})
```
Doekin, line 73: On Windows and MinGW platforms, tcmalloc requires linking to specific system libraries. However, `link_libraries()` in the `CMakeLists.txt` is applied too late in the build process, causing linking issues. See https://github.com/gperftools/gperftools/blob/e1014dead2029b341d06027b4f2b5562d799d5b1/CMakeLists.txt#L451-L467

A quick solution can be implemented by leveraging another `if(MINGW OR MSVC) ... endif()` block at the beginning of the `CMakeLists.txt` file:
https://github.com/gperftools/gperftools/blob/e1014dead2029b341d06027b4f2b5562d799d5b1/CMakeLists.txt#L32C1-L37C8

packages/g/gperftools/xmake.lua
```diff
@@ -4,30 +4,90 @@ package("gperftools")
...
     end
 
     add_deps("cmake")
 
-    on_load("linux", function (package)
-        if package:config("unwind") then
+    if on_check then
+        on_check("windows", "mingw", function (package)
+            assert(package:config("minimal"), "package(gperftools): only tcmalloc_minimal is supported on Windows or MinGW")
+            if is_plat("windows") then
```
waruqi, line 36: package:is_plat

PR #5579, opened 2024-10-25
Support pcapplusplus with zstd

jj683:
https://pcapplusplus.github.io/docs/next/features#readingwriting-pcapng-files-with-compression


jj683:
```
/home/runner/.xmake/cache/packages/2410/p/pcapplusplus/v24.09/source/3rdParty/json/include/json.hpp:24356:35: error: identifier '_json' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator]
 24356 | inline nlohmann::json operator "" _json(const char* s, std::size_t n)
       |                       ~~~~~~~~~~~~^~~~~
       |                       operator""_json
```

The problem is that the 3rd party json library that PcapPlusPlus use has a deprecated literal operator in it, which causes a warning (and error as the result of -Werror) in modern versions of clang.
I have created a PR to update the json library version (which fixes this problem):
https://github.com/seladb/PcapPlusPlus/pull/1623
But it will take months for the new release. 

How should we proceed in such a case? The extra flag with zstd (disabled by default) does not affect the build in any way.

waruqi:
use io.replace or add_patches to patch and remove `-Werror`

star-hengxing:
`-Werror` or `-WX` cannot be unconditionally enabled, package management must remove these flags.

star-hengxing:
However, a better way for this library would be to patch cmake to remove the bundles dependencies.
If you're interested, consider using `find_package` or pkgconfig + import target.

https://github.com/xmake-io/xmake-repo/blob/df3a58c3887ca8113621d748c5fb2a3081b4b724/packages/m/minio-cpp/patches/0.3.0/cmake-pkgconfig-find-deps.patch
https://github.com/xmake-io/xmake-repo/blob/df3a58c3887ca8113621d748c5fb2a3081b4b724/packages/m/minio-cpp/xmake.lua#L14-L19

jj683:
Should works now with new cmake `--compile-no-warning-as-error` flag, added in #5419

PR #5691, opened 2024-11-07
cpu_features: add v0.9.0 version

star-hengxing:
- Add alias package.
- Support more platform.


star-hengxing:
Convenient for users to find.

```console
$ xmake require cpu_features
checking for cpu_features ... no
note: the following packages were not found in any repository (check if they are spelled correctly):
  -> cpu_features
```

SirLynix:
We should probably improve xmake tooltips when it doesn't find something (based on levenshtein distance for example) instead of adding fake packages

PR #5733, opened 2024-11-11
boost: add `heap`

qudix:
- add support for the `heap` module

also I'm uncertain if the `sorted_libs` table actually needs to be sorted in some way (if it does there needs to be a comment), but for now I just added `heap` at the end

### DIFF
--- a/packages/b/boost/libs.lua
+++ b/packages/b/boost/libs.lua
@@ -33,7 +33,8 @@ local sorted_libs = {
   "chrono",
   "system",
   "charconv",
-  "atomic"
+  "atomic",
+  "heap"
 }
 
 local libs_dep = {
@@ -149,7 +150,8 @@ local libs_dep = {
     "mpi",
     "random",
     "serialization"
-  }
+  },
+  heap = {}
 }
 
 local header_only_buildable = {
@@ -158,6 +160,7 @@ local header_only_buildable = {
   "exception",
   "regex",
   "math",
+  "heap"
 }
 
 function get_libs()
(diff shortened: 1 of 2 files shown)


star-hengxing:
`sorted_libs` are non-header-only libs in boost that we use to fix link order.

If you find that some header-only libs are missing from the boost cmake version, please open an issue.


```cmake
# Generated by `boostdep --cmake heap`
# Copyright 2020 Peter Dimov
# Distributed under the Boost Software License, Version 1.0.
# https://www.boost.org/LICENSE_1_0.txt

cmake_minimum_required(VERSION 3.5...3.16)

project(boost_heap VERSION "${BOOST_SUPERPROJECT_VERSION}" LANGUAGES CXX)

add_library(boost_heap INTERFACE)
add_library(Boost::heap ALIAS boost_heap)

target_include_directories(boost_heap INTERFACE include)

target_link_libraries(boost_heap
  INTERFACE
    Boost::array
    Boost::assert
    Boost::bind
    Boost::concept_check
    Boost::config
    Boost::core
    Boost::intrusive
    Boost::iterator
    Boost::parameter
    Boost::static_assert
    Boost::throw_exception
    Boost::type_traits
)

if(BUILD_TESTING AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/CMakeLists.txt")

  add_subdirectory(test)

endif()
```

qudix:
The boost package needs more active maintenance in terms of missing libs/sub dependencies (serialization doesn't build). There's no documentation for the custom fetching and whatnot and my attempt at adding a lib was half-baked because of this.

PR #5892, opened 2024-12-02
Update libcurl

Redbeanw44602:


Redbeanw44602:
Reasons for the error on apple platform: https://github.com/curl/curl/issues/14269

SirLynix:
we can add https://github.com/curl/curl/commit/5f6b9244ceecbb1e1b9c08bda5c9292f23f4ea57 as a patch for versions where it fails

Redbeanw44602:
> we can add [curl/curl@5f6b924](https://github.com/curl/curl/commit/5f6b9244ceecbb1e1b9c08bda5c9292f23f4ea57) as a patch for versions where it fails

Thanks, I'm doing this.

Redbeanw44602:
I think it's ready to be merged now.

packages/l/libcurl/xmake.lua
```diff
@@ -30,11 +31,14 @@ package("libcurl")
...
+    if is_plat("android") and is_host("windows") then
+        add_deps("ninja")
+        set_policy("package.cmake_generator.ninja", true)
+    end
+
     -- we init all configurations in on_load, because package("curl") need it.
     on_load(function (package)
-        if package:is_plat("linux", "android", "cross") then
-            package:config_set("openssl", true)
-        end
```
SirLynix, line 37: Removing those lines broke existing xmake projects doing `add_requires("libcurl")`

Redbeanw44602: It works fine on my device (linux, x86_64), can you provide more information?

SirLynix: `add_requires("libcurl")` on Linux no longer links with libcurl and no longer has HTTPS support, resulting in `unsupported protocol` errors.

before your change, openssl support was enabled by default

Redbeanw44602: This does break the default HTTPS support, but this is expected because
1. HTTPS support should not be forced, curl builds do not enforce ssl
2. There are other options now (openssl3)

Downstreams should choose whether to enable encryption, and which one to enable.

Redbeanw44602: If one must be turned on by default, it is difficult to choose which one to turn on. I think it is best to let the user choose.

SirLynix: I understand it's expected, but it's problematic to update a package and break the existing behavior, especially here where all xmake projects using libcurl will get a failure on the broadly used https protocol.

openssl should still be the default when no option is set, I think we can detect the case where openssl and openssl3 are not set (nil) and enable openssl as before for retrocompatibility.

@waruqi is this possible?

Redbeanw44602: I think the ability to build libcurl without SSL support needs to be preserved, is there a way for xmake to tell if a certain config is set "by default"? Otherwise I think it's a good idea to automatically enable openssl.

SirLynix: I think so but I need confirmation from xmake author, afaik configs can have three values : true, false or nil (not set).
what I'm suggesting is basically:
```lua
if package:config("openssl") == nil and package:config("openssl3") == nil then
    package:config_set("openssl", true)
end
```

this will support `add_requires("libcurl", { openssl = false })` as well

waruqi: If openssl is not explicitly set and no default value is configured in add_configs, it defaults to nil

PR #5897, opened 2024-12-03
Update poco to 1.14.0

heheda123123:


heheda123123:
好像是编译器支持的问题？最低要求c++17
https://github.com/pocoproject/poco/blob/main/Foundation/include/Poco/Types.h
https://pocoproject.org/blog/?p=1249

PR #5912, opened 2024-12-06
libxml2: use if statement instead of assert

Chi-EEE:
Using `xrepo info itstool` throws "`error: package(libxml2) python interface require shared lib`"

itstool

### DIFF
--- a/packages/l/libxml2/xmake.lua
+++ b/packages/l/libxml2/xmake.lua
@@ -79,7 +79,7 @@ package("libxml2")
             package:add("deps", "readline")
         end
         if package:config("python") then
-            if package:config("shared") then
+            if not package:config("shared") then
                 raise("package(libxml2) python interface require shared lib")
             end
             if package:is_cross() then
--- a/packages/l/libxml2/xmake.lua
+++ b/packages/l/libxml2/xmake.lua
@@ -79,7 +79,9 @@ package("libxml2")
             package:add("deps", "readline")
         end
         if package:config("python") then
-            assert(package:config("shared"), "package(libxml2) python interface require shared lib")
+            if package:config("shared") then
+                raise("package(libxml2) python interface require shared lib")
+            end
             if package:is_cross() then
                 raise("libxml2 python interface does not support cross-compilation")
             end


packages/l/libxml2/xmake.lua
```diff
@@ -79,7 +79,9 @@ package("libxml2")
             package:add("deps", "readline")
         end
         if package:config("python") then
-            assert(package:config("shared"), "package(libxml2) python interface require shared lib")
+            if package:config("shared") then
```
waruqi, line 82: ? why not `if not package:config("shared") then`? and why not use assert?

Chi-EEE: done, using `xrepo info itstool` throws an error with the following and does not output any info about `itstool`:
```
error: package(libxml2) python interface require shared lib
error: execv(C:\Users\admin\xmake\xmake.exe require --info --extra={system=false} itstool) failed(-1)
```

waruqi: So what is the difference between using assert and the current patch? `xrepo info libxml2` checks the static library libxml2 by default. No matter which one is used, it will trigger raise

you should switch to shared config to show info

```console
ruki-2:tmp ruki$ xrepo info libxml2[python,iconv]
The package info of project:
error: package(libxml2) python interface require shared lib
error: execv(/Users/ruki/projects/personal/xmake/build/xmake require --info --extra={system=fal
se} libxml2[python,iconv]) failed(255)

ruki-2:tmp ruki$ xrepo info -k shared libxml2[python,iconv]
The package info of project:
    require(cmake):
      -> description: A cross-platform family of tool designed to build, test and package softw
are
      -> version: 3.30.2
```

Chi-EEE: Using `assert` throws and does not allow `xrepo info` to display information about the package while the `raise` allows the information to be shown without having to specify `xrepo info` to use shared config

waruqi: No, they are equivalent, you just wrote the `if` condition wrong earlier.

```console
$ xrepo info libxml2[python]
The package info of project:
error: package(libxml2) python interface require shared lib <-------- apply this patch and raise error
error: execv(/Users/ruki/.local/bin/xmake require --info --extra={system=false} libxml2[python]) failed(255)
```

Chi-EEE: Oh, I'll find another way to fix this later

waruqi:
see https://github.com/xmake-io/xmake-repo/pull/5930

PR #5921, opened 2024-12-06
itstool: Support bsd

Chi-EEE:
The package does not work by default but setting the libxml2 dep to shared makes it work

project
build_with_itstool

### DIFF
--- a/packages/i/itstool/xmake.lua
+++ b/packages/i/itstool/xmake.lua
@@ -1,16 +1,17 @@
 package("itstool")
-
     set_kind("binary")
     set_homepage("http://itstool.org/")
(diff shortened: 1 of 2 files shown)


star-hengxing:
Same error on windows, I will try to fix it.
```console
python -c "import libxml2"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\star\AppData\Local\.xmake\packages\l\libxml2\v2.13.4\9478b424b65342c6a463e8d95d9d3e3d\python\libxml2.py", line 1, in <module>
    import libxml2mod
ImportError: DLL load failed while importing libxml2mod: 找不到指定的模块。
```

Issue #5913, opened 2024-12-06
Add `xrepo info <package>` test


Chi-EEE:
### Is your feature request related to a problem? Please describe.

I'm trying to get the list of configs for the packages but errors show when I run `xrepo info` on some of the packages

### Describe the solution you'd like

Add `xrepo info <package>` test to verify that the command works

### Additional context

https://github.com/xmake-io/xmake-repo/pull/5910, https://github.com/xmake-io/xmake-repo/pull/5911, https://github.com/xmake-io/xmake-repo/pull/5921

```lua
package("package_with_broken_on_load")
    add_configs("hide_existing_package_error", {default = is_plat("windows"), type = "boolean"})

    if on_check then
        on_check(function (package)
            if package:is_plat("linux") then
                raise("This package does not support Linux for ... reasons.")
            end
        end)
    end

    on_load(function (package)
        if package:config("hide_existing_package_error") then -- CI does not check this
            add_syslinks("advapi32")
            raise("This package is broken on purpose.")
        end
    end)

    on_install(function (package)
    end)

    on_test(function (package)
    end)
```

CI is green but there is an error:
![image](https://github.com/user-attachments/assets/bb03c430-806d-4650-81cb-2300b54dba5a)

Chi-EEE:
I mean add a test which uses `xrepo info` to test if the on_load method of the package works and if it fails, we know that using `xrepo info` does not work on the package and its information is not displayed

Chi-EEE:
![itstool](https://github.com/user-attachments/assets/1dc5fd9d-2765-4174-82d3-91c150ff49c8)
![node-addon-api](https://github.com/user-attachments/assets/0213a451-80a9-4242-8615-6128078e7db1)
![xerces-c](https://github.com/user-attachments/assets/ae4e01a5-d656-4fe5-8f0c-c0503796ad90)

waruqi:
you need set package config

```console
ruki-2:tmp ruki$ xrepo info -k shared libxml2[python,iconv]
The package info of project:
    require(cmake):
      -> description: A cross-platform family of tool designed to build, test and package softw
are
      -> version: 3.30.2
```

Chi-EEE:
This request is unnecessary as the test script already tests on_load

PR #5922, opened 2024-12-06
add `xmake require --info` test

Chi-EEE:
This adds `xmake require --info` to the test script and allows any platform / arch to use `xmake require --info` to test the package's on_load method.  
The regular test script does not check the `on_load` method if the platform / arch is excluded from the package.  
But any platform / arch can still access the `on_load` method via `xrepo info <package>` and this PR aims to test this.

https://github.com/xmake-io/xmake-repo/issues/5913

### DIFF
--- a/scripts/test.lua
+++ b/scripts/test.lua
@@ -53,6 +53,19 @@ function _check_package_is_supported()
     end
 end
 
+function _info_packages(argv, packages)
+    local info_argv = {"require", "-f", "-y", "--info"}
+    if argv.verbose then
+        table.insert(info_argv, "-v")
+    end
+    if argv.diagnosis then
+        table.insert(info_argv, "-D")
+    end
+    for _, package in irpairs(packages) do
+        os.vexecv(os.programfile(), table.join(info_argv, package))
+    end
+end
+
 -- require packages
 function _require_packages(argv, packages)
     local config_argv = {"f", "-c"}
@@ -265,18 +278,6 @@ function main(...)
         table.insert(packages, "tbox dev")
     end
 
-    -- remove unsupported packages
-    for idx, package in irpairs(packages) do
-        assert(package == package:lower(), "package(%s) must be lower case!", package)
-        if not _package_is_supported(argv, package) then
-            table.remove(packages, idx)
-        end
-    end
-    if #packages == 0 then
-        print("no testable packages on %s!", argv.plat or os.subhost())
-        return
-    end
-
     -- prepare test project
     local repodir = os.curdir()
     local workdir = path.join(os.tmpdir(), "xmake-repo")
@@ -304,6 +305,23 @@ function main(...)
     end
     os.execv(os.programfile(), {"repo", "--add", "local-repo", repodir})
     os.execv(os.programfile(), {"repo", "-l"})
+
+    -- info packages
+    _info_packages(argv, packages)
+
+    os.cd(repodir)
+    -- remove unsupported packages
+    for idx, package in irpairs(packages) do
+        assert(package == package:lower(), "package(%s) must be lower case!", package)
+        if not _package_is_supported(argv, package) then
+            table.remove(packages, idx)
+        end
+    end
+    if #packages == 0 then
+        print("no testable packages on %s!", argv.plat or os.subhost())
+        return
+    end
+    os.cd(path.join(workdir, "test"))
 
     -- require packages
     _require_packages(argv, packages)


waruqi:
I still don't understand what the modification of test.lua has to do with `xrepo info`.

waruqi:
we not need this patch, scripts/check.lua will test them, all assert/raise should be moved to on_check() in package.

Chi-EEE:
I don't see `scripts/check.lua` ?

waruqi:
We shouldn't do any checks in on_load. There was no on_check at the beginning. Now these checks should be moved to on_check.

```lua
    on_check(function (package)
        if package:config("python") then
            assert(package:config("shared"), "package(libxml2) python interface require shared lib")
            if package:is_cross() then
                raise("libxml2 python interface does not support cross-compilation")
            end
            if not package:config("iconv") then
                raise("libxml2 python interface requires iconv to be enabled")
            end
        end
    end)

```

Chi-EEE:
Then I think that we should add this patch as it verifies that there are no checks in on_load

Chi-EEE:
For example, the current CI does not catch errors that are hidden by configs: https://github.com/Chi-EEE/xmake-repo/pull/20

```lua
package("package_with_broken_on_load")
    add_configs("hide_existing_package_error", {default = is_plat("windows"), type = "boolean"})

    if on_check then
        on_check(function (package)
                raise("This package does not support Linux for ... reasons.")
            end
        end)
    end

    on_load(function (package)
        if package:config("hide_existing_package_error") then -- CI does not check this
            add_syslinks("advapi32")
            raise("This package is broken on purpose.")
        end
    end)

    on_install(function (package)
    end)

    on_test(function (package)
    end)

```

Chi-EEE:
With this patch, it highlights the errors: https://github.com/Chi-EEE/xmake-repo/pull/21

waruqi:
it works for me on my macos. test.lua will call on_load, then trigger raise.

```lua
    add_configs("hide_existing_package_error_2", {default = is_plat("macosx"), type = "boolean"})

    on_load(function (package)
        if package:config("hide_existing_package_error_2") then
            raise("This package is broken on purpose.")
        end
```

```console
unknown764636769726:xmake-repo ruki$ xmake l scripts/test.lua -vD --shallow libxml2
{
  "libxml2"
}

/private/var/folders/32/w9cz0y_14hs19lkbs6v6_fm80000gn/T/.xmake501/241208/xmake-repo/test
add local repository(local-repo): /Users/ruki/projects/personal/xmake-repo ok!
local repositories:
    local-repo /Users/ruki/projects/personal/xmake-repo

global repositories:
    build-artifacts https://gitee.com/xmake-mirror/build-artifacts.git main
    xmake-repo https://gitee.com/tboox/xmake-repo.git master
    builtin-repo /Users/ruki/projects/personal/xmake/xmake/repository

4 repositories were found!
checking for platform ... macosx
checking for architecture ... x86_64
checking for Xcode directory ... /Applications/Xcode.app
checking for SDK version of Xcode for macosx (x86_64) ... 15.1
checking for Minimal target version of Xcode for macosx (x86_64) ... 15.1
checking for nim ... /usr/local/bin/nim
configure
{
    host = macosx
    kind = static
    proxy_pac = pac.lua
    buildir = build
    theme = default
    network = public
    ndk_stdcxx = true
    mode = release
    arch = x86_64
    ccache = true
    xcode = /Applications/Xcode.app
    clean = true
    plat = macosx
}
checking for git ... /usr/bin/git
/usr/bin/git rev-parse HEAD
error: @programdir/core/main.lua:329: ...rojects/personal/xmake-repo/packages/l/libxml2/xmake.lua:
60: This package is broken on purpose.
stack traceback:
    [C]: in function 'error'
```

waruqi:
ci will not test all configs combinations, so it is meaningless even if you add `xrepo info`. It is uncertain whether `xrepo info` will trigger raise, and it cannot test all configs combinations.

If you want to ensure that all error checks in on_load are triggered, you should improve CI to traverse and test all configs, but this is impossible, and CI does not have so many resources.

Therefore, at present, we only need to improve the package when encounter problems, and move the remaining error checks in on_load to on_check.

Chi-EEE:
We can set up a basic CI test to verify the package's default configuration, which is the most commonly used. This test would detect and highlight any config errors in its deps.  

Additionally, for any future package that uses a different config for the dep, the CI will also check it:
https://github.com/Chi-EEE/xmake-repo/pull/22

Chi-EEE:
> it works for me on my macos. test.lua will call on_load, then trigger raise.

It shows as green on the CI: https://github.com/Chi-EEE/xmake-repo/actions/runs/12221570001/job/34090775584

waruqi:
so it will show

> no testable packages on macosx or you're using lower version xmake!

this package is unspported on the current platform.

Chi-EEE:
image

image

But CI shows as green and user does not know if there is an error or not without manually checking each CI:
image

PR #5923, opened 2024-12-06
Update Raylib-cpp to v5.5.0

burdockcascade:


packages/r/raylib-cpp/xmake.lua
```diff
@@ -9,10 +9,11 @@ package("raylib-cpp")
 
     add_versions("v5.0.2", "d3a718170882bc873c973a19a824d7fa4bfd9d0087b4778057231409a240920d")
     add_versions("v5.0.1", "6d10469019700fd5993db9a18bdd0ed025105b1bf7dd8916e353eef8bfac6355")
+    add_versions("v5.5.0", "bcb4a4e241a95376e8562aa77c29976ed0921235a9f5326822130a7bcf4860a5")
 
-	add_deps("raylib 5.x")
+	add_deps("raylib 5.5")
```
waruqi, line 14: why? if set `add_requires("raylib-cpp 5.0.2")`, it will break.

PR #5946, opened 2024-12-09
Improve to test packages

waruqi:
https://github.com/xmake-io/xmake-repo/pull/5922


Chi-EEE:
There's a problem that if the platform / arch is excluded, the `xrepo info` test will not get called in the CI:
```lua
package("package_with_broken_on_load")
    on_load(function (package) -- Any platform / arch can call this function through `xrepo info package_with_broken_on_load`
            raise("This package is broken on purpose.") -- raise inside of on_load
        end
    end)

    on_install("linux", function (package)
    end)

    on_test(function (package)
    end)

```
image

waruqi:
done

PR #5953, opened 2024-12-11
fix test.lua for info test

Chi-EEE:
Before:
image

After:
image

PR #5972, opened 2024-12-13
fix test.lua cd

Chi-EEE:
https://github.com/xmake-io/xmake-repo/pull/5953#issuecomment-2541642373


scripts/test.lua
```diff
@@ -335,14 +335,16 @@ function main(...)
     -- load packages
     _load_packages(argv, packages_original)
 
+    os.cd(repodir)
```
waruqi, line 338: why? local oldir = os.cd()

PR #5981, opened 2024-12-14
Test new Apache arrow xmake.lua

Altina-oz:
The old script fails to build Arrow >=10.0, because since 9.0 Apache arrow has significantly evolved, including make arrow-python an independent package -- Pyarrow (which should also get included in xmake-repo in the future. Many components also have their names changed, or get removed, or have some new features thus the dependencies has been changed. 

Caveats:

[component orc]
Due to a bug of gcc-14 or Apache ORC library, the Apache orc cannot be compiled with gcc-14 -O3 -Warray-bounds (c.f. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111273 for a similar case) . Therefore the orc component cannot set ON.

[component gcs]
Component gcs depends on  Google cloud storage cpp library (maybe c.f. https://github.com/googleapis/google-cloud-cpp), which is not included in xmake-repo. Therefore Arrow cmake script would use its bundled gcs which depends on a special version of abseil library conflicting the one from xrepo, and a link time error would make some error involving symbols occur, and the installation would fail. Thus it is left not implemented

[arrow-python]
This option is deprecated since 10.0.0. This will be removed in a future release. -> build an independent package "pyarrow" instead.

[s3, gandiva_java]
I have not test these options, maybe not work. Gandiva_java is jni binding of gandiva component, s3 is for amazon-s3 support.  I don't know what are their true dependencies.

[on Windows& MAC]
I just followed the official document (c.f. https://arrow.apache.org/docs/developers/cpp/windows.html) but not really test if this script can work on windows or mac

-----
Let's see ci result first

------

### DIFF
--- a/packages/a/arrow/xmake.lua
+++ b/packages/a/arrow/xmake.lua
@@ -47,7 +47,7 @@ package("arrow")
     
     ----------------------------------------------------
     --Dependencies
-    add_deps("cmake", "boost", "xsimd")
+    add_deps("cmake", "xsimd")
 
     if is_plat("linux") then
         add_syslinks("pthread")
@@ -216,6 +216,7 @@ package("arrow")
         table.insert(configs, "-DARROW_WITH_ZSTD=ON")
     end)
     on_load("windows", "linux", "macosx", "mingw",function (package)
+        package:add("deps", "boost", {configs = {locale = true}})
         --optional components
         --Allocator
         if package:config("jemalloc") then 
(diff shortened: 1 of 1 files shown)


Altina-oz:
Why this can pass on my local machine but cannot pass ci ()

`File or directory
  /github/home/.xmake/packages/t/thrift/v0.21.0/23344c1027974b71ab16298fb370ee93/bin
  referenced by variable THRIFT_BIN_DIR does not exist !`

Altina-oz:
Seems that the ci system cannot find ANYthing

Altina-oz:
Give up(

PR #6019, opened 2024-12-21
Auto-update openal-soft to 1.24.1

waruqi:
New version of openal-soft detected (package version: 1.23.1, last github version: 1.24.1)


SirLynix:
I don't get why it passes with OpenAL linking in shared, I set up the same -pthread flag 😕

PR #6038, opened 2024-12-24
SFML: update to 3.0.0

Doekin:
### SFML 3.0.0 Noteworthy Changes
- Raised C++ requirement to C++17
- [Removed OpenAL shared library requirement](https://github.com/SFML/SFML/blob/master/changelog.md#general)
- The default constructors `sf::Sound::Sound()`, `sf::Text::Text()`, and `sf::Sprite::Sprite()` were [removed.](https://github.com/SFML/SFML/blob/master/migration.md#removal-of-default-constructors)
- Changed default library type from shared to static
- When using X11 as the backend on Linux, libxi-dev is a newly [required dependency.](https://www.sfml-dev.org/tutorials/3.0/getting-started/migrate/#linux-dependencies)

### DIFF
--- a/packages/s/sfml/xmake.lua
+++ b/packages/s/sfml/xmake.lua
@@ -268,14 +268,14 @@ package("sfml")
                 sf::Clock c;
                 c.restart();
             }
-        ]]}, {includes = "SFML/System.hpp"}))
+        ]]}, {configs = configs, includes = "SFML/System.hpp"}))
         if package:config("graphics") then
             assert(package:check_cxxsnippets({test = [[
                 void test(const sf::Texture& texture, const sf::Color& color) {
                     sf::Sprite sprite(texture);
                     sprite.setColor(color);
                 }
-            ]]}, {includes = "SFML/Graphics.hpp", configs = configs}))
+            ]]}, {configs = configs, includes = "SFML/Graphics.hpp"}))
         end
         if package:config("window") or package:config("graphics") then
             assert(package:check_cxxsnippets({test = [[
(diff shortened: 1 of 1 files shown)


packages/s/sfml/xmake.lua
```diff
@@ -73,7 +80,8 @@ package("sfml")
         component:add("extsources", "brew::sfml/sfml-audio")
         component:add("extsources", "pkgconfig::sfml-audio")
         if not package:config("shared") and package:is_plat("windows", "mingw") then
-            component:add("links", "openal32", "flac", "vorbisenc", "vorbisfile", "vorbis", "ogg")
+            component:add("links", "flac", "vorbisenc", "vorbisfile", "vorbis", "ogg")
+            if package:version():lt("3.0.0") then component:add("links", "openal32") end
```
waruqi, line 84:
```lua
if ... then
end
```

packages/s/sfml/xmake.lua
```diff
@@ -194,8 +203,9 @@ package("sfml")
                                 for _, libfile in ipairs(libfiles) do
                                     table.insert(libraries, (libfile:gsub("\\", "/")))
                                 end
+                                local lib_name = package:version():lt("3.0.0") and "Freetype" or "freetype"
                                 local file = io.open("src/SFML/Graphics/CMakeLists.txt", "a")
-                                file:print("target_link_libraries(Freetype INTERFACE " .. table.concat(libraries, " ") .. ")")
+                                file:print(format("target_link_libraries(%s INTERFACE %s)", lib_name, table.concat(libraries, " ")))
```
waruqi, line 208: file:print support format string

packages/s/sfml/xmake.lua
waruqi: and package:version() maybe nil if use git url + branch

PR #6079, opened 2024-12-27
OpenSSL: improve MinGW support on Windows

Doekin:
OpenSSL targeting MinGW can be built on Linux or MSYS2, but not directly on Windows. This PR introduces a workaround to enable building on Windows using Git Bash.


Doekin:
I discovered that the Perl bundled with Git Bash can actually run OpenSSL's build script. I will rework my PR accordingly.

packages/o/openssl/xmake.lua
```diff
@@ -32,15 +32,18 @@ package("openssl")
...
-                package:add("deps", "nasm", { private = true })
-                -- the perl executable found in GitForWindows will fail to build OpenSSL
-                -- see https://github.com/openssl/openssl/blob/master/NOTES-PERL.md#perl-on-windows
-                package:add("deps", "strawberry-perl", { system = false, private = true })
-                -- check xmake tool jom
-                import("package.tools.jom", {try = true})
-                if jom then
-                    package:add("deps", "jom", {private = true})
+            elseif is_subhost("windows") then
+                package:add("deps", "strawberry-perl", { private = true })
```
waruqi, line 36: why remove `{system = false}`, it will break some things. it will use incorrect perl in some systems and ci envs.

packages/o/openssl/xmake.lua
```diff
@@ -32,15 +32,18 @@ package("openssl")
             if package:is_plat("android") and is_subhost("windows") and os.arch() == "x64" then
                 -- when building for android on windows, use msys2 perl instead of strawberry-perl to avoid configure issue
                 package:add("deps", "msys2", {configs = {msystem = "MINGW64", base_devel = true}, private = true})
-            elseif is_subhost("windows") and not package:is_precompiled() then
```
waruqi, line 35: and why not remove `not package:is_precompiled()`? 
If you are using the build-artifacts binary, you do not need to compile openssl and install its build dependencies.

packages/o/openssl/xmake.lua
Doekin: Could you please share more details about the scenarios where the system Perl might fail? There might be a workaround we can implement to address this issue. Using the system Perl helps save disk space.

For your reference, I tested with Perl installed via `scoop`, and it seemed to work as expected.

waruqi: I can't reproduce it because I don't remember what happened before. But since there is a clear comment to fix this problem by passing `system = false`, we shouldn't just remove it.

Doekin: The comment indicated that Perl bundled with GitForWindows fails to build OpenSSL. I encountered this issue too at the beginning. The error message is as follows:
```
Can't locate Pod/Usage.pm in @INC (you may need to install the Pod::Usage module) (@INC entries checked: . /usr/lib/perl5/site_perl /usr/share/perl5/site_perl /usr/lib/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl) at configdata.pm line 11398.
BEGIN failed--compilation aborted at configdata.pm line 11398.
Compilation failed in require.
BEGIN failed--compilation aborted.
```
I looked into the config script and found that `Pod::Usage` is only used to generate help messages, which are shown with the `./Configure help` command. I guess that the help message is not needed when installing the package. So, I removed the use of the `Pod::Usage` module, and it builds successfully using Perl from GitForWindows.

Doekin:
### Supported Combinations of Targets and Windows Environments
- Build for **Windows** in **PowerShell** with **Strawberry Perl**
- Build for **MinGW** in **Git Bash** with bundled Perl
- Build for **Windows** in **PowerShell** with **Git for Windows**' bundled Perl
- Build for **MinGW** in **Git Bash** with **Strawberry Perl**

### Unsupported Combinations of Targets and Environments
- Build for **MinGW** in **PowerShell**
  - This is expected due to the reliance on many Unix-like shell commands in the `makefile`.

packages/o/openssl/xmake.lua
waruqi: and here

packages/o/openssl/xmake.lua
```diff
@@ -28,20 +28,25 @@ package("openssl")
...
-            if package:is_plat("android") and is_subhost("windows") and os.arch() == "x64" then
-                -- when building for android on windows, use msys2 perl instead of strawberry-perl to avoid configure issue
-                package:add("deps", "msys2", {configs = {msystem = "MINGW64", base_devel = true}, private = true})
-            elseif is_subhost("windows") and not package:is_precompiled() then
+        if package:is_plat("android") and is_subhost("windows") and os.arch() == "x64" then
+            -- when building for android on windows, use msys2 perl instead of strawberry-perl to avoid configure issue
+            package:add("deps", "msys2", {configs = {msystem = "MINGW64", base_devel = true}, private = true})
+        elseif is_subhost("windows") then
+            import("lib.detect.find_tool")
+            local perl = find_tool("perl", {paths={"$(env PERL)"}})
```
waruqi, line 36: we should not call `find_tool` in on_load

packages/o/openssl/xmake.lua
```diff
@@ -61,6 +66,20 @@ package("openssl")
         end
     end)
 
+    if on_check then
+        on_check(function (package)
+            import("lib.detect.find_tool")
+            local perl = assert(find_tool("perl", {paths={"$(env PERL)"}}), "package(openssl): perl not found!")
```
waruqi, line 72: use package:find_tool and add spaces ` = `

packages/o/openssl/xmake.lua
```diff
@@ -83,7 +102,13 @@ package("openssl")
             table.insert(configs, "no-makedepend")
             table.insert(configs, "/FS")
         end
-        os.vrunv("perl", configs)
+        import("lib.detect.find_tool")
+        local perl = assert(find_tool("perl", {paths={"$(env PERL)"}}), "package(openssl): perl not found!")
```
waruqi, line 106: we should not call `find_tool`, we should use `add_deps` to bind perl path end and run `perl`

packages/o/openssl/xmake.lua
Doekin: OK, there are no `is_precompiled` now

SirLynix:
I'm sorry but I don't understand what benefit this PR does, it seems complicated to just use the system perl instead of installing strawberry perl or am I missing something?

Doekin:
The explanation is as follows:

1. If we build OpenSSL for Windows, we can use previously installed Strawberry Perl.
2. If we build OpenSSL for MinGW, Strawberry Perl cannot produce the Unix-style paths that OpenSSL expects, leading to build failures. Therefore, I added prompts to guide users to build in the correct environment.

packages/o/openssl/xmake.lua
waruqi: > OK, there is no `is_precompiled` now

but we need is_precompiled.

Doekin: Sorry, I misunderstood.

waruqi: If you are downloading a precompiled binary, then all build dependencies do not need to be installed.

https://github.com/xmake-mirror/build-artifacts/releases?q=openssl&expanded=true

Doekin: I guess the outer `package:is_precompiled()` might work as well.

https://github.com/xmake-io/xmake-repo/blob/a7144ca9afa10b5fc0c9b584ea2f96231ea53dde/packages/o/openssl/xmake.lua#L30-L31

Doekin:
### Changes up to now

- **Allow Strawberry-Perl not installed by XMake to be used when targeting Windows.**
- **Fix failure when running the configure script in Git Bash.**

<details>
<summary>Example:</summary>

`xrepo install -vD -p mingw openssl`

Git Bash Configure Script
</details>

- **Fix the creation of `NUL` file on Windows.**

<details>
<summary>Example:</summary>

NUL File Issue
NUL File Issue
</details>

- **Fix `The command line is too long` error.**

<details>
<summary>Example:</summary>

`xmake install -vD -p mingw --toolchain=zig openssl`

Command Line Too Long
</details>

- **Patch for using `llvm-rc` as the Windows resource compiler when targeting MinGW.**

packages/o/openssl/makefile/patch.lua
```diff
@@ -0,0 +1,35 @@
+function _patch_for_llvm_rc(package, opt)
+    if package:is_plat("mingw") and package:has_tool("mrc", "llvm_rc", "llvm-rc", "rc") then
+        local cc = package:build_getenv("cc")
+        local cflags = opt.buildenvs.CFLAGS
+        local tmpfile = path.unix(os.tmpfile() .. ".c")
```
waruqi, line 5: we need to remove tmpfile after running this script

packages/o/openssl/makefile/patch.lua
```diff
@@ -0,0 +1,35 @@
...
+        io.writefile(tmpfile, "int main(void) { return 0; }\n")
+        local compile_out, compile_err = try {function() return os.iorun(format("%s -v %s %s", cc, cflags, tmpfile)) end}
```
waruqi, line 7: use lua table as argv and call `os.iorunv(cc, argv)`

packages/o/openssl/xmake.lua
```diff
@@ -83,6 +100,7 @@ package("openssl")
             table.insert(configs, "no-makedepend")
             table.insert(configs, "/FS")
         end
+        import("configure.patch")(package)
```
waruqi, line 103: Did you test all openssl versions? does this patch work for all versions?

packages/o/openssl/makefile/patch.lua
Doekin: OK

packages/o/openssl/makefile/patch.lua
Doekin: I encountered an issue where the `CFLAGS` might be set to `-m64 --target=x86_64-w64-windows-gnu`, resulting in a concatenated command like `clang "-v" "-m64 --target=x86_64-w64-windows-gnu" a.c`. This causes the following error:

```
clang-19: error: unknown argument: '-m64 --target=x86_64-w64-windows-gnu'
clang version 19.1.6 (https://github.com/llvm/llvm-project.git e21dc4bd5474d04b8e62d7331362edcc5648d7e5)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: D:/Applications/Scoop/apps/mingw-mstorsjo-llvm-ucrt/19.1.6-20241217/bin
```

waruqi: because buildenvs.CFLAGS has been merged to string. not lua table.

packages/o/openssl/makefile/patch.lua
```diff
@@ -0,0 +1,36 @@
...
+            include_directive = include_directive .. format([[ -I "%s"]], include_dir)
+        end
+        -- $(RC)  $(RCFLAGS)  -o $@ libcrypto.rc =>  $(RC) -I inc_dir -I inc_dir -FO $@ libcrypto.rc
+        io.gsub("Makefile", [[%$%(RC%).-%$@%s+(%S+)]], format("$(RC) %s -FO $@ ", include_directive).."%1")
+        io.gsub("Makefile", "(%S+).res.o", "%1.res")
+    end
+end
+
+function main(package, opt)
+    -- _patch_for_llvm_rc(package, opt)
```
waruqi, line 35: ?

packages/o/openssl/xmake.lua
Doekin: I tested the latest version (`1.1.1-w`) with the following combinations of targets and environments:

- Platform: Windows, Host: Windows
- Platform: MinGW, Host: MSYS2 (Git Bash)
- Platform: MinGW, Host: Linux
- Platform: Linux, Host: Linux

The patch does not work for versions `1.0.0` and `1.0.2-u`, resulting in the error:
```
error: @programdir\core\sandbox\modules\io.lua:188: cannot open file: Configurations/10-main.conf, Unknown Error (3)
```

Version `1.1.0-l` failed with the error:
```
Unsupported options: no-tests
```

The patch works for versions `1.1.1-h` and `1.1.1-p`.

packages/o/openssl/makefile/patch.lua
Doekin: It was an unintentional change.

packages/o/openssl/xmake.lua
waruqi: you can use `package:version():ge("1.1.1")` to limit patch

Doekin: OpenSSL versions prior to `1.1.1` do not support the `no-tests` build flag, and there is a conflict between `/FS` and `VC-WIN64A`. After removing these flags, I still could not build OpenSSL `1.1.0-l` for MinGW. While `1.1.0-l` builds for Windows, `1.0.2-u` does not.

waruqi: you can improve on_check to detect these cases.

Doekin: I discovered that OpenSSL `1.0` requires a different installation method. Detailed explanations can be found here: [How to Build OpenSSL 1.0 on Windows](https://github.com/pocoproject/poco/wiki/How-to-Build-OpenSSL-1.0-on-Windows).

star-hengxing:
This package does't have git urls, so you don't need to use `package:gitref()`

PR #6091, opened 2024-12-29
Limit ImGui <=1.90.4 to Polyscope > 2.2.0

Whalesea-Gustav:
Polyscope Officially Support ImGui Version is 1.90.4
It has Compilation Error When Using Latest ImGui=1.91.6
So, In this Commit, Limit ImGui Version to 1.90.4 For Polyscope > 2.2.0
Snipaste_2024-12-29_14-37-23


Whalesea-Gustav:
ImGui 1.91.6 is not compatible with polyscope > 2.2.0.
It has to Use Older Version of ImGui

PR #6103, opened 2024-12-31
add ta-lib

fasiondog:
add ta-lib


star-hengxing:
If only static kind is supported, Please add:
```lua
```

Switch to xmake build only if some platforms build failed.
Building with cmake is better for maintenance. If you don't want to add cmake support, please add this config.
```lua
add_configs("cmake", {description = "Use cmake buildsystem", default = false, type = "boolean", readonly = true})
```

fasiondog:
Already added configs "shared".

There're some problems with cmake,  so use xmake:
1. different include dir on different platform
2. dll on windows, only for java/python wrap.  But use c/c++, have a problems.

PR #6163, opened 2025-01-15
pcl: fix boost version requirement

longhao-li:
PCL uses some internal boost functions and classes. Therefore, some versions of PCL relies on specific boost version.

Seems like that the PCL developers are working with boost 1.87.0 now, but boost 1.87.0 currently does not work with PCL 1.14.1. Maybe this should be changed in the future.


xq114:
https://github.com/PointCloudLibrary/pcl/issues/6157

PR #6173, opened 2025-01-17
spdlog: only add '/utf-8' when not use std_format and external fmt

wsw0108:


qudix:
I'm not too in the loop but why does the source/execution character set have to be explicitly set for spdlog? star-hengxing's pr which introduced the change doesn't have any details.

SirLynix:
`/utf-8` is a [requirement for fmt package](https://github.com/xmake-io/xmake-repo/blob/dev/packages/f/fmt/xmake.lua#L62) when unicode is enabled (default), but not for std::format.

I think we can remove /utf-8 flag from spdlog as it will automatically be inherited by fmt package as well if it's used

wsw0108:
The flag `/utf-8` may only be used for bundled fmt.

But I think maybe we should remove support of bundled fmt.

luadebug:
> The flag `/utf-8` may only be used for bundled fmt.
> 
> But I think maybe we should remove support of bundled fmt.

Does lief port rely over bundled one, or even better question is how many ports rely over bundle? :/

star-hengxing:
This package is not complicated. To maintain compatibility, it is not necessary to remove the bundle fmt config

star-hengxing (approved):

SirLynix (changes requested):
These changes are not correct, /utf-8 should only be added with bundled fmt

SirLynix (approved):

PR #6205, opened 2025-01-21
fix reflect-cpp compilation when msgpack is enabled

mirageAlchemy:

------

### DIFF
--- a/packages/r/reflect-cpp/patches/0.16.0/cmake.patch
+++ b/packages/r/reflect-cpp/patches/0.16.0/cmake.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 2e16468..38d289f 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -138,7 +138,7 @@ if (REFLECTCPP_MSGPACK)
+     if(REFLECTCPP_USE_VCPKG)
+         find_package(msgpack-c CONFIG REQUIRED)
+     else()
+-        find_package(msgpack CONFIG REQUIRED)
++        find_package(msgpack-c CONFIG REQUIRED)
+     endif()
+     target_link_libraries(reflectcpp PUBLIC msgpack-c)
+ endif ()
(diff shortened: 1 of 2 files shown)


mirageAlchemy:
i would actually like to include some trivial tests for msgpack behaviour, maybe like this:
```lua
        
            #include <rfl/msgpack.hpp>
            #include <rfl.hpp>
            struct Person {
                std::string first_name;
                std::string last_name;
                int age;
            };
            const auto homer = Person{.first_name = "Homer",
                                      .last_name = "Simpson",
                                      .age = 45};
            void test() {
                const std::vector<char> msgpack_data = rfl::msgpack::write(homer);
                auto homer2 = rfl::msgpack::read<Person>(msgpack_data).value();
            }   
        ]]}, {configs = {languages = "c++20", msgpack = true}}))
```
But I am not quite sure it is advocated?

waruqi:
> i would actually like to include some trivial tests for msgpack behaviour, maybe like this:
> 
> ```lua
>         
>         assert(package:check_cxxsnippets({test = [[
>             #include <rfl/msgpack.hpp>
>             #include <rfl.hpp>
>             struct Person {
>                 std::string first_name;
>                 std::string last_name;
>                 int age;
>             };
>             const auto homer = Person{.first_name = "Homer",
>                                       .last_name = "Simpson",
>                                       .age = 45};
>             void test() {
>                 const std::vector<char> msgpack_data = rfl::msgpack::write(homer);
>                 auto homer2 = rfl::msgpack::read<Person>(msgpack_data).value();
>             }   
>         ]]}, {configs = {languages = "c++20", msgpack = true}}))
> ```
> 
> But I am not quite sure it is advocated?

How is it different from the current test? As long as the test code is kept as short as possible, it is allowed to add some other tests as appropriate.

mirageAlchemy:
Current test only covers a subset of all configuration cases, that is only json support is enabled. The test I provided above has `msgpack` configured, so ensures any possible upgrade does not break `msgpack` support. I was hesitant to add it because I am not sure if I should add the test only when the user demands `msgpack` explicitly in config, or should I include it by default so it is checked by CI.

waruqi:
> Current test only covers a subset of all configuration cases, that is only json support is enabled. The test I provided above has `msgpack` configured, so ensures any possible upgrade does not break `msgpack` support. I was hesitant to add it because I am not sure if I should add the test only when the user demands `msgpack` explicitly in config, or should I include it by default so it is checked by CI.

we can test it when msgpack is enabled.

```lua
if package:config("msgpack") then
    -- test it
end
```

mirageAlchemy:
I have added a test only when msgpack is specified by the user

PR #6216, opened 2025-01-22
update nim to 2.2.0

luadebug:
It looks like works only for win :/
For arm64 win it looks like downloading nim x64 instead of building... and it looks working.

os.vrunv("./bootstrap-vcpkg.sh", {shell = true}) there is no {} for some reason.

os.vrunv("chmod", {"+x", "./build_all.sh"}) ?

### DIFF
--- a/packages/n/nim/xmake.lua
+++ b/packages/n/nim/xmake.lua
@@ -21,6 +21,7 @@ package("nim")
         add_urls("https://github.com/nim-lang/Nim/archive/refs/tags/v$(version).tar.gz")
         add_versions("2.0.2", "2ca2f559d05e29f130cb4f319ebb93a98e7c0e2187716b17b2cb4e747f5ff798")
         add_versions("2.2.0", "5fc7059f777a665df09465419d187ac2911548715fbb31ccb78d51704ece1784")
+        add_deps("bdwgc", "make", "sfml")
     end
 
     on_install("@windows", "@msys", function (package)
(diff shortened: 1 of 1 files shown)


packages/n/nim/xmake.lua
```diff
@@ -21,6 +21,7 @@ package("nim")
         add_urls("https://github.com/nim-lang/Nim/archive/refs/tags/v$(version).tar.gz")
         add_versions("2.0.2", "2ca2f559d05e29f130cb4f319ebb93a98e7c0e2187716b17b2cb4e747f5ff798")
         add_versions("2.2.0", "5fc7059f777a665df09465419d187ac2911548715fbb31ccb78d51704ece1784")
+        add_deps("bdwgc", "make", "sfml")
```
waruqi, line 24: 为啥要加 make? ./build_all.bat 这里面用了？

另外， macos/linux 下就没必要了吧

luadebug: > 为啥要加 make?

I have tried to repeat dependencies gathering like this https://github.com/nim-lang/Nim/blob/devel/azure-pipelines.yml#L133 and did not expect CI has *make* over Mac OS. OK.
> ./build_all.bat 这里面用了？

I do not have arm64 win, but I assume it uses binary win64 because 
build (windows-2022, shared, arm64, MT)
```
  => download https://nim-lang.org/download/nim-2.2.0_x64.zip .. ok
```
it should be?
```
  => download https://github.com/nim-lang/Nim/archive/refs/tags/v2.2.0.tar.gz .. ok
```
  nim --version
Nim Compiler Version 2.2.0 [Windows: amd64]
>另外， macos/linux 下就没必要了吧

OK, I will remove "make" from add_deps. But I do not know what to do next. %_%

Issue #6156, opened 2025-01-12
Add prerelease SDL3 to repo


PerikiyoXD:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Implement `sdl3` as package or `libsdl` versioning

qudix:
I'm fairly certain we're waiting for an official non-pre release

vspefs:
just a quick tip: sdl3 finally released.

PR #6225, opened 2025-01-23
Add libsdl3

SirLynix:
#6156 

I took the libsdl package and removed all deprecated stuff, I may have missed some things though.

### DIFF
--- a/packages/l/libsdl3/xmake.lua
+++ b/packages/l/libsdl3/xmake.lua
@@ -18,7 +18,7 @@ package("libsdl3")
     add_versions("archive:3.2.0", "abe7114fa42edcc8097856787fa5d37f256d97e365b71368b60764fe7c10e4f8")
     add_versions("github:3.2.0", "release-3.2.0")
 
-    add_deps("cmake", "egl-headers")
+    add_deps("cmake", "egl-headers", "opengl-headers")
 
     if is_plat("linux", "bsd") then
         add_configs("x11", {description = "Enables X11 support (requires it on the system)", default = true, type = "boolean"})
@@ -69,20 +69,22 @@ package("libsdl3")
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DSDL_TEST_LIBRARY=OFF")
         table.insert(configs, "-DSDL_EXAMPLES=OFF")
-        local opt = {}
+
+        local cflags
+        local packagedeps
         if not package:is_plat("wasm") then
-            opt.packagedeps = {"egl-headers"}
+            packagedeps = {"egl-headers", "opengl-headers"}
         end
 
         if package:is_plat("linux", "bsd", "cross") then
-            table.insert(opt.packagedeps, "libxext")
-            table.insert(opt.packagedeps, "libx11")
-            table.insert(opt.packagedeps, "xorgproto")
+            table.insert(packagedeps, "libxext")
+            table.insert(packagedeps, "libx11")
+            table.insert(packagedeps, "xorgproto")
         elseif package:is_plat("wasm") then
             -- emscripten enables USE_SDL by default which will conflict with the sdl headers
-            opt.cflags = {"-sUSE_SDL=0"}
+            cflags = {"-sUSE_SDL=0"}
         end
-        import("package.tools.cmake").install(package, configs, opt)
+        import("package.tools.cmake").install(package, configs, {cflags = cflags, packagedeps = packagedeps})
     end)
 
     on_test(function (package)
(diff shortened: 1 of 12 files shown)


packages/l/libsdl3/xmake.lua
```diff
@@ -0,0 +1,112 @@
...
+    add_urls("https://www.libsdl.org/release/SDL3-$(version).zip",
+             "https://github.com/libsdl-org/SDL/releases/download/release-$(version)/SDL3-$(version).zip", { alias = "archive" })
+    add_urls("https://github.com/libsdl-org/SDL.git", { alias = "github" })
+
+    add_versions("archive:3.2.0", "abe7114fa42edcc8097856787fa5d37f256d97e365b71368b60764fe7c10e4f8")
+    add_versions("github:3.2.0", "release-3.2.0")
+
+    add_deps("cmake")
+
+    add_includedirs("include", "include/SDL3")
```
qudix, line 23: Maybe remove this, or at least the second include.

luadebug:
https://github.com/actions/runner-images/issues/10980#issuecomment-2513163023
Altering 2022 MSVC CI maybe would help, though I do not understand exactly how this change should look like.
```
 windows:
    runs-on: windows-2025
```

SirLynix:
I don't think it's the best way to fix this since some users will have 2022 CI as well.

Kbz-8:
if the new package name is `libsdl3`, wouln't it be logical to name the "old" one `libsdl2` instead of `libsdl` ?

SirLynix:
Yes indeed, the next step is to rename libsdl to libsdl2 and keep a libsdl package using libsdl2 as a base, like this:
```lua
package("libsdl")
    set_base("libsdl2")

    on_load(function (package)
        wprint("libsdl package has been renamed to libsdl2 following the release of SDL3 which is also available under the name libsdl3.${clear}")
    end)
```

packages/l/libsdl3/xmake.lua
```diff
@@ -100,6 +100,10 @@ package("libsdl3")
         if #includedirs > 0 then
             includedirs = table.unique(includedirs)
             table.insert(configs, "-DCMAKE_INCLUDE_PATH=" .. table.concat(includedirs, ";"))
+            cflags = cflags or {}
+            for _, includedir in impairs(includedirs) do
```
waruqi: ipairs

packages/l/libsdl3/xmake.lua
```diff
@@ -0,0 +1,122 @@
...
+    add_urls("https://github.com/libsdl-org/SDL.git", { alias = "github" })
+
+    add_versions("archive:3.2.0", "abe7114fa42edcc8097856787fa5d37f256d97e365b71368b60764fe7c10e4f8")
+    add_versions("github:3.2.0", "release-3.2.0")
+
+    add_deps("cmake", "egl-headers", "opengl-headers")
+
+    if is_plat("linux", "bsd") then
+        add_configs("x11", {description = "Enables X11 support (requires it on the system)", default = true, type = "boolean"})
+        add_configs("wayland", {description = "Enables Wayland support", default = true, type = "boolean"})
```
qudix, line 25: I don't do linux development, but is there a reason why these are enabled by default on that platform? I thought you could develop a libsdl3 application without these?

SirLynix: One of the main feature of SDL (and reason people use it) is input and window management, which need x11/wayland to work.
It's very rare to use libsdl without these

qudix: I'm always wary about defaults being enabled because you can't really take it back later. Even if it's rare, minimal dependencies by default should still be a consideration.

SirLynix: I understand, but libsdl is a bit special, it only requires headers to build as it doesn't link x11 nor wayland, it loads them manually (dlopen/dlsym) at runtime only if they're available.

A good default for libsdl is to build it with as many deps as possible for this reason

SirLynix:
cross and linux ARM64 fail because of wayland not supporting cross-building at the moment (because of wayland-scanner), I think this should be addressed. I'm disabling wayland by default for now on cross-platforms

packages/l/libsdl3/xmake.lua
```diff
@@ -0,0 +1,132 @@
...
+    add_versions("archive:3.2.0", "abe7114fa42edcc8097856787fa5d37f256d97e365b71368b60764fe7c10e4f8")
+
+    add_versions("github:3.2.2", "release-3.2.2")
+    add_versions("github:3.2.0", "release-3.2.0")
+
+    add_deps("cmake", "egl-headers", "opengl-headers")
+
+    if is_plat("linux", "bsd", "cross") then
+        add_configs("x11", {description = "Enables X11 support", default = true, type = "boolean"})
+        add_configs("wayland", {description = "Enables Wayland support", default = nil, type = "boolean"})
```
waruqi, line 28: if default is nil, we need not set it.

SirLynix: yeah that's just to make it explicit and not let other developer think it's missing

PR #6245, opened 2025-01-29
fix(reflect-cpp): missing headers

std-microblock:


packages/r/reflect-cpp/xmake.lua
```diff
@@ -120,10 +120,9 @@ package("reflect-cpp")
             table.insert(configs, "-DREFLECTCPP_UBJSON=" .. (package:config("ubjson") and "ON" or "OFF"))
             table.insert(configs, "-DREFLECTCPP_YAML=" .. (package:config("yaml") and "ON" or "OFF"))
             import("package.tools.cmake").install(package, configs)
-        else
-            os.rm("include/thirdparty")
-            os.cp("include", package:installdir())
         end
+        os.rm("include/thirdparty")
```
mirageAlchemy, line 124: hi @std-microblock , may I ask why did you add these two lines of code when compiling on the newest version? For me the cmake tool seems to already install the include to the correct path, I have locally the following as result:
```bash
Install the project...
-- Install configuration: "Release"
-- Installing: /home/dev/.xmake/packages/r/reflect-cpp/v0.17.0/802210e8886b421c95a7dc2c7ad507eb/lib/cmake/reflectcpp/reflectcpp-config.cmake
-- Installing: /home/dev/.xmake/packages/r/reflect-cpp/v0.17.0/802210e8886b421c95a7dc2c7ad507eb/lib/libreflectcpp.a
-- Installing: /home/dev/.xmake/packages/r/reflect-cpp/v0.17.0/802210e8886b421c95a7dc2c7ad507eb/include/rfl.hpp
-- Installing: /home/dev/.xmake/packages/r/reflect-cpp/v0.17.0/802210e8886b421c95a7dc2c7ad507eb/include/rfl/AddStructName.hpp
```
Are you trying to install an older version?

std-microblock: No, I'm trying to install the latest and it seems that only a partial of the headers are available. For example, DefaultIfMissing.hpp is not here. So I added these lines and it appeared.

mirageAlchemy: That is strange, I have the `DefaultIfMissing.hpp` locally at the correct path before as well. My local package test also works, below is the test I used:
image
My best guess is the cmake tool might not work as expected on other platform(I am on macos and linux), can you confirm that?

std-microblock: I'm on windows. And it's not only DefaultIfMissing. I have no preprocessor headers locally.. You can check the precompiled binary.

mirageAlchemy: I can confirm I do have a `Processor.hpp` in my installation. I dont have a windows env now, so cant really test if the cmake tooling breaks things. Do you mind if I open a new pull request adding the test snippet above, purely for test purpose? What it concerns mostly to me is maybe there is better way to fix it. By the way, what do you mean by checking precompiled binary?

std-microblock: wtf...I did a full rebuild and the problem disappeared...
you can change the line back anyway..
image
this is what it like when I first used it

mirageAlchemy: > wtf...I did a full rebuild and the problem disappeared... you can change the line back anyway.. image this is what it like when I first used it

sure thing, I need to start a pull request upgrading to v0.17.0 anyway. thanks for confirming!

PR #6263, opened 2025-02-01
Fix mimalloc debug build

WSSDude:
Unsure if this is proper fix as I am unsure if the folder may be different on different platform or not, but this fixes compilation on the latest Windows 11 system using latest Visual Studio 2022 and xmake 2.9.7 for debug builds.

### DIFF
--- a/packages/m/mimalloc/xmake.lua
+++ b/packages/m/mimalloc/xmake.lua
@@ -30,6 +30,8 @@ package("mimalloc")
         add_configs("etw", {description = "Enable Event tracing for Windows", default = false, type = "boolean"})
     end
 
+    set_policy("package.cmake_generator.ninja", true)
+
     add_deps("cmake")
 
     if is_plat("windows") then
@@ -79,11 +81,6 @@ package("mimalloc")
         if package:gitref() or package:version():ge("2.1.2") then
             table.insert(configs, "-DMI_INSTALL_TOPLEVEL=ON")
             import("package.tools.cmake").install(package, configs, {cxflags = cxflags})
-
-            if package:is_plat("windows") and package:is_debug() then
-                local dir = package:installdir(package:config("shared") and "bin" or "lib")
-                os.cp(path.join(package:buildir(), "mimalloc-debug.pdb"), dir)
-            end
         else
             import("package.tools.cmake").build(package, configs, {buildir = "build", cxflags = cxflags})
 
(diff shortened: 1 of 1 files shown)


packages/m/mimalloc/xmake.lua
```diff
@@ -81,8 +81,9 @@ package("mimalloc")
             import("package.tools.cmake").install(package, configs, {cxflags = cxflags})
 
             if package:is_plat("windows") and package:is_debug() then
+                local pdb = package:config("shared") and "mimalloc-debug.pdb" or "mimalloc-static-debug.pdb"
                 local dir = package:installdir(package:config("shared") and "bin" or "lib")
-                os.cp(path.join(package:buildir(), "mimalloc-debug.pdb"), dir)
+                os.cp(path.join(package:buildir(), "pdb", "Debug", pdb), dir)
             end
```
star-hengxing, line 87: This code can be removed, fixed by https://github.com/xmake-io/xmake/issues/5566

star-hengxing:
```lua
add_requires("mimalloc[shared]")
```

pdb generate by msbuild:
```
build_xxx/pdb/<build type>/
```

pdb generate by ninja:
```
build_xxx/pdb/
```
> Enable by policy `package.cmake_generator.ninja`

https://github.com/xmake-io/xmake/blob/78371cb084abfe3e80e928adae4e55a0df98f34b/xmake/modules/package/tools/cmake.lua#L1383-L1389
```lua
if package:is_plat("windows") and os.isdir("pdb") then
    if package:config("shared") or not package:is_library() then
        os.trycp("pdb/**.pdb", package:installdir("bin"))
    else
        os.trycp("pdb/**.pdb", package:installdir("lib"))
    end
end
```

star-hengxing:
> Can confirm the line can be removed but it still does not fix the issue.
> 
> `mimalloc` included by `tiltedcore` package seems to be an issue somehow, but that package is not doing anything special.
> 
> ```
> checking for debug.x64 ...
> checking for Microsoft Visual Studio (x64) version ... 2022
> checking for Microsoft C/C++ Compiler (x64) version ... 19.40.33813
>   => install mimalloc v2.1.7 .. ok
>   => install mimalloc#1 v2.1.7 .. failed
> 
> cannot copy file build_5604f22e\pdb\Debug\mimalloc-static-debug.pdb, file not found!
> if you want to get more verbose errors, please see:
>   -> C:\Users\andre\AppData\Local\.xmake\cache\packages\2502\m\mimalloc_1\v2.1.7\installdir.failed\logs\install.txt
> error: install failed!
> ```
> 
> There is nothing special in the install.txt, only that single error. Looking into `build_5604f22e`, there exists folder `pdb/Debug` but it is empty. For the one that runs OK, there is a normal PDB file present in that folder.
> 
> PDB simply does not get generated for the case where `mimalloc` is included from another package, at least not from `tiltedcore` which literally only has `add_requires("mimalloc")` looking at it.
> 
> Trying to use `add_requires("mimalloc[shared]")` to bypass this (at least I understood it should be some bypass), doesnt work still (above log is from the run with it included)

@waruqi Any idea?

star-hengxing:
Try `add_requireconfs("mimalloc", {configs = {rltgenrandom = true}})` to avoid building multiple version (configs) mimalloc

PR #6283, opened 2025-02-04
wasm-micro-runtime: add latest version (2.2.0)

ethindp:
This PR adds version 2.2.0 of WAMR. I have verified that it builds (without need for patches) with both the default configuration and with a modified configuration enabling most options.


packages/w/wasm-micro-runtime/xmake.lua
```diff
@@ -19,8 +20,8 @@ package("wasm-micro-runtime")
     add_configs("fast_interp", {description = "Enable fast interpreter", default = false, type = "boolean"})
     add_configs("aot", {description = "Enable AOT", default = false, type = "boolean"})
     -- TODO: improve llvm
-    add_configs("jit", {description = "Enable JIT", default = false, type = "boolean", readonly = true})
-    add_configs("fast_jit", {description = "Enable Fast JIT", default = false, type = "boolean", readonly = true})
+    add_configs("jit", {description = "Enable JIT", default = false, type = "boolean", readonly=true})
+    add_configs("fast_jit", {description = "Enable Fast JIT", default = false, type = "boolean", readonly=true})
```
waruqi, line 24: why? add spaces

ethindp:
I'm a bit confused on the failures here, it builds fine on my machine (though I'm using VS2022 so... Maybe that's it?). The error looks like it can't find a kernel function but that doesn't make any sense, the CMakeLists should be handling that part.

packages/w/wasm-micro-runtime/xmake.lua
ethindp: I tend to instinctively pad my operators, so I didn't really think much about it, sorry about that.

ethindp:
Okay, so it looks like, for the failing windows targets, it's trying to redefine `NtQueryTimerResolution`. I'm not entirely sure why it's trying to redefine this function, but I'm also very confused why it's succeeding on many other windows targets but not on these ones.

star-hengxing:
You can patch it.
https://github.com/bytecodealliance/wasm-micro-runtime/commit/8c2759c9d62a4e3203b4c7d210640cf2d031c388.patch

ethindp:
@star-hengxing I didn't know about that one, thanks for that, I believe I applied it correctly.

ethindp:
cc @waruqi

ethindp:
I just tried to fix the patch thing for windows by only having it apply on windows... Hopefully that fixes it.

ethindp:
@star-hengxing Looks like either I broke the patch somehow or something else is wrong, it's being rejected when it gets applied

star-hengxing:
It seems this patch has too many discrepancies with the version 2.2.0 codebase.
https://github.com/bytecodealliance/wasm-micro-runtime/commits/main/core/shared/platform/windows/win_clock.c

My recommendation is to temporarily disable Windows x86 builds for this version.
```lua
if on_check then
    on_check("windows", function (package)
        if package:version() and package:version():eq("2.2.0") then
            if package:is_arch("x86") then
                raise("package(wasm-micro-runtime 2.2.0) unsupported arch.")
            end
        end
    end)
end
```

ethindp:
@star-hengxing I assume this would be a temporary thing until a new version comes out? I've just applied it, was hoping this would work everywhere but... Oh well.

PR #6303, opened 2025-02-06
tinygltf: fix deps include

star-hengxing:
```console
error: /home/runner/.xmake/packages/t/tinygltf/v2.9.5/f012e308851442128d62ffb7ad2d897b/include/tiny_gltf.h:1725:10: fatal error: json.hpp: No such file or directory
 1725 | #include "json.hpp"
      |          ^~~~~~~~~~
```
related https://github.com/xmake-io/xmake-repo/pull/6298

### DIFF
--- a/packages/n/nlohmann_json/xmake.lua
+++ b/packages/n/nlohmann_json/xmake.lua
@@ -4,8 +4,9 @@ package("nlohmann_json")
     set_description("JSON for Modern C++")
     set_license("MIT")
 
-    add_urls("https://github.com/nlohmann/json/archive/$(version).tar.gz",
+    add_urls("https://github.com/nlohmann/json/archive/refs/tags/$(version).tar.gz",
              "https://github.com/nlohmann/json.git")
+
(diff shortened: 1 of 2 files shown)


zse4t7i:
Can I ask why `"-DTINYGLTF_INSTALL_VENDOR=OFF"` and then replace `#include "json.hpp"` to `"#include <nlohmann/json.hpp>"`?

SirLynix:
> Can I ask why `"-DTINYGLTF_INSTALL_VENDOR=OFF"` and then replace `#include "json.hpp"` to `"#include <nlohmann/json.hpp>"`?

To use nlohmann_json package instead of its own packaged version

zse4t7i:
Yes, I know that. I just don't understand why not use its own dependencies. So it's a common practice in package management?

SirLynix:
> Yes, I know that. I just don't understand why not use its own dependencies. So it's a common practice in package management?

It can cause multiple versions of nlohmann_json to exist in the user space, prevent updating it, and more. It's better to use packages so xmake is aware of it.

zse4t7i:
Thanks for your answer.

PR #6355, opened 2025-02-13
extra-cmake-modules: add package

Doekin:
This package is required as a dependency for certain CMake-based projects.

### DIFF
--- a/packages/e/extra-cmake-modules/xmake.lua
+++ b/packages/e/extra-cmake-modules/xmake.lua
@@ -1,0 +1,33 @@
+package("extra-cmake-modules")
+    set_homepage("https://invent.kde.org/frameworks/extra-cmake-modules")
+    set_description("Extra CMake Modules (ECM) extends CMake with additional modules and scripts.")
+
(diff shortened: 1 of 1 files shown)


packages/e/extra-cmake-modules/xmake.lua
```diff
@@ -0,0 +1,32 @@
+package("extra-cmake-modules")
+    set_homepage("https://invent.kde.org/frameworks/extra-cmake-modules")
+    set_description("Extra CMake Modules (ECM) extends CMake with additional modules and scripts.")
+
+    add_urls("https://invent.kde.org/frameworks/extra-cmake-modules/-/archive/v$(version)/extra-cmake-modules-v$(version).tar.gz" , {alias = "gitlab"})
+    -- add_urls("https://invent.kde.org/frameworks/extra-cmake-modules.git")
```
Doekin, line 6: I attempted to use the following versioning logic:  
```lua
{version = function (version) return "v" .. tostring(version) end}
```  
However, it did not work as expected. Any suggestions or guidance would be greatly appreciated.

PR #6389, opened 2025-02-18
Add imgui versions v1.91.7 and v1.91.8

AidanSun05:


qudix:
Just add `v1.91.8`, there's no need to add the intermediary versions as I don't think CI will test those, only the latest.

AidanSun05:
I think the CI should be testing everything, in the logs (like from `build (ubuntu-latest, shared, release)` for example) it contains these lines:

```text
[...]
  => install imgui v1.91.8 .. ok
[...]
  => install imgui#1 v1.91.7 .. ok
[...]
  => install imgui#2 v1.91.8-docking .. ok
[...]
  => install imgui#3 v1.91.7-docking .. ok
```

star-hengxing:
> Just add `v1.91.8`, there's no need to add the intermediary versions as I don't think CI will test those, only the latest.

https://github.com/xmake-io/xmake-repo/blob/cf1cdfee6dd8f70f3b34a06ccf458ac7e9f612e8/scripts/test.lua#L267-L286

The test script will find all `add_versions` strings in the git diff.

Issue #6399, opened 2025-02-20
Using `boost_filesystem` shared lib without `BOOST_FILESYSTEM_DYN_LINK=1` defined results in runtime double free


Doekin:
### Xmake Version

xmake v2.9.8+20250204

### Operating System Version and Architecture

Arch Linux

### Describe Bug

This issue occurs when using the system-installed `boost` libraries. If `boost` is built from source via XMake, this issue does not occur because XMake automatically defines `BOOST_ALL_DYN_LINK` when building shared libraries. However, with system-installed shared `boost` libraries, if `BOOST_FILESYSTEM_DYN_LINK=1` (or `BOOST_ALL_DYN_LINK`) is not defined, it results in a runtime error: `free(): double free detected in tcache 2`.

### Expected Behavior

If system-installed shared `boost` libraries are being used, XMake should automatically define `BOOST_FILESYSTEM_DYN_LINK=1` (or `BOOST_ALL_DYN_LINK`).

### Project Configuration

```cpp
#include <boost/filesystem.hpp>
int main() {}
```

### Additional Information and Error Logs

```bash
free(): double free detected in tcache 2
```

waruqi:
This seems to be a boost issue. Alternatively, you can define BOOST_FILESYSTEM_DYN_LINK yourself via add_cxflags

Doekin:
Thank you for the suggestion! However, when using `xrepo` to install packages like `xrepo install boostdep`, this isn't feasible.

Doekin:
Apologies, I misidentified the cause of this issue. The problem is not related to the absence of the `BOOST_FILESYSTEM_DYN_LINK=1` macro.

The issue occurs when we use system-installed `boost` like this:

```lua
add_requires("boost", {configs = {filesystem = true}})
```

However, if we explicitly specify shared `boost`, the issue disappears:

```lua
add_requires("boost", {shared = true, configs = {filesystem = true}})
```

The difference seems to be that additional libraries are linked when using the shared version. For example, here are the compile commands for both cases:

With `add_requires("boost", {configs = {filesystem = true}})`:
```
/usr/bin/gcc -c -m64 -fvisibility=hidden -fvisibility-inlines-hidden -Wall -Wextra -Wpedantic -O3 -DNDEBUG -o build/.objs/boost_fs/linux/x86_64/release/src/boost_fs.cpp.o src/boost_fs.cpp
[...]

Issue #6398, opened 2025-02-20
Fail to use system `boost` with `undefined reference` in `libboost_python`


Doekin:
### Xmake Version

xmake v2.9.8+20250204

### Operating System Version and Architecture

Arch Linux

### Describe Bug

When using the system-installed `boost`, which includes various modules such as `libboost_python`, attempting to link `libboost_python` without explicitly adding `-lpython` results in linker errors (undefined references).

### Expected Behavior

If `libboost_python` is detected in the system, perhaps it could also link against `python` to avoid the issue.

### Project Configuration

```bash
xrepo install boostdep
```

### Additional Information and Error Logs

N/A

waruqi:
you can improve boost package, add python syslinks in on_fetch()

waruqi:
you can add python library dep in boost package when boost/python is enabled.

Doekin:
I tried adding the following to `on_load`:

```lua
if package:config("python") then
    package:add("deps", "python")
end
```

However, it didn’t work. This might be because `config("python")` is disabled by default, and the system `boost` is detected during the `on_fetch` step.

star-hengxing:
Can we add deps in `on_fetch`?
https://github.com/xmake-io/xmake-repo/blob/ec1c0e01d915c53bf56d8ce2be16d0dfbe9e7d88/packages/b/boost/fetch.lua#L5-L21

waruqi:
> Can we add deps in `on_fetch`?

no, we should use `add_deps` to add python, boost still can fetch system libs.

star-hengxing:
I thought that xrepo does not execute `on_load` after executing `on_fetch`, so these code should work.

https://github.com/xmake-io/xmake-repo/blob/14f527a6c13fdc5f9a8e40755c5220255bcd7e10/packages/b/boost/b2/load.lua#L68-L73

https://github.com/xmake-io/xmake-repo/blob/bd8a6e07a9da25cbf27eb25a39d306c66ded366f/packages/b/boost/cmake/load.lua#L65-L67

`add_requires("boost[python]")` work for me.

```console
[ 50%]: cache compiling.release src/main.cpp
/usr/bin/gcc -c -m64 -fPIC -isystem /usr/include/python3.10 -isystem /home/star/.xmake/packages/o/openssl/1.1.1-w/6c51ab6278e2479b883dffafac69fdaf/include -o build/.objs/hello/linux/x86_64/release/src/main.cpp.o src/main.cpp
[...]

PR #6447, opened 2025-02-25
boost: Link Python when boost_python is fetched

Doekin:
This PR attempts to address issues #6398 ~~and #6399~~ by:  
~~1. Preferring shared libraries for Boost when they are available on the system~~
2. linking Python when `boost_python` is present

### DIFF
--- a/packages/b/boost/fetch.lua
+++ b/packages/b/boost/fetch.lua
@@ -60,22 +60,24 @@ function main(package, opt)
         local found
         libs.for_each(function (libname)
             local sub_libs = sub_libs_map[libname]
-            if sub_libs then
-                for _, sub_libname in ipairs(sub_libs) do
-                    local linkinfo = find_library("boost_" .. sub_libname, paths, opt)
-                    if linkinfo then
-                        _add_info(linkinfo, result)
-                        found = true
-                    end
-                end
-            else
-                local linkinfo = find_library("boost_" .. libname, paths, opt)
+            for _, sub_libname in ipairs(sub_libs or {libname}) do
+                local linkinfo = find_library("boost_" .. sub_libname, paths, opt)
                 if linkinfo then
                     _add_info(linkinfo, result)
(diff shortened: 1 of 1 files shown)


packages/b/boost/fetch.lua
```diff
@@ -50,32 +50,39 @@ function main(package, opt)
...
                         found = true
                     end
                 end
-            else
-                local linkinfo = find_library("boost_" .. libname, paths, opt)
-                if linkinfo then
-                    _add_info(linkinfo, result)
-                    found = true
+            end)
+            if found or package:config("shared") then
```
waruqi, line 72: when `package:config("shared") == false`, use static libs, but shared libs are found. result will be shared libs.

waruqi: I know, but if `package:config("shared") == false`, we should only fetch static libraries.

if you want to use shared libraries, you should use `add_requires("boost", {configs = {shared = true}})`

Doekin: Sorry for the confusion, here’s my thought:

Currently, when static libraries are fetched, shared libraries are actually used due to their presence.

To clarify this situation and avoid mixing shared libraries (which are available both as shared and static) with pure static libraries (only available as static), it might make sense to always prefer shared libraries when possible, 
considering that `add_requires("openssl")` or `add_deps("python")` are common ways to fetch system-installed libraries, where `package:config("shared")` is false, yet shared libraries are still being fetched.

Alternatively, perhaps we could use `add_linkgroups("a", "b", {static = true})` to ensure that static libraries are exactly used when fetched?

waruqi: If you want to ensure compatibility and stability, you should disable system libraries. `add_requires("boost", {system = false})`

Now many packages' system libraries cannot control whether to choose static libraries or dynamic libraries, and there are many uncertainties.

PR #6450, opened 2025-02-26
Add botan versions 3.6.0 to 3.7.1

AidanSun05:


AidanSun05:
Some actions in the CI are failing for these reasons:

- Windows: `LINK : fatal error LNK1146: no argument specified with option '/MT'` (This is a regression in the botan library: https://github.com/randombit/botan/issues/4424; fixed in 3.7.0)
- Linux arm64: `ERROR: Unknown or unidentifiable processor "arm64-v8a"` (`arm64-v8a` will need to be aliased to `aarch64` or `arm64`: https://github.com/randombit/botan/blob/master/src/build-data/arch/arm64.txt)
- macOS: `src/lib/utils/cpuid/cpuid_aarch64.cpp:147:10: error: use of undeclared identifier 'OS'` (Will need to enable `os_utils` module since 3.7.0: https://github.com/randombit/botan/pull/4457)

AidanSun05:
The Ubuntu ARM CI is still failing since the configure script doesn't recognize `arm64-v8a`. This is what is set in the CI's yml config:

https://github.com/xmake-io/xmake-repo/blob/ec5b0ef6e7029e6a92e49a4676f65795274868c5/.github/workflows/ubuntu_arm64.yml#L35-L37

However, `arm64-v8a` is an Android ABI and a 64-bit Linux ARM system would report its architecture as something like `arm64` or `aarch64`. Aliasing `arm64-v8a` to e.g., `aarch64` in this package's `xmake.lua` would not be a good long term solution, and a better approach would be to set up this CI with `arm64` or `aarch64` to better reflect the behavior of ARM Linux. @waruqi @star-hengxing What do you think?

waruqi:
why not remove `-a arm64-v8a`? if ci is arm64, xmake will detect arch automatically. @star-hengxing

AidanSun05:
I pushed some changes to fix the remaining CI errors:

- macOS and FreeBSD static: `ar: @build/response_files/static.txt: No such file or directory` (some distributions of `ar` on these systems don't support the `@file` syntax so the build rules have been patched)
- Temporarily set an alias to `arm64-v8a` to confirm that the library builds on ARM Linux in the CI. The CI setup may be affecting other packages as well, so if it is resolved we will probably remove this alias.

waruqi:
https://github.com/xmake-io/xmake-repo/pull/6473

PR #6461, opened 2025-02-27
adding tinycthread package for windows, linux and macos

Kbz-8:


SirLynix:
I don't understand how it passes tests without linking pthread

Kbz-8:
> I don't understand how it passes tests without linking pthread

tenor

Kbz-8:
isn't FreeBSD failing because phread is not linked ?

PR #6466, opened 2025-02-27
test async_simple

luadebug:
It seems after following upstream idea of using -m32/-m64 in case of host cpu x86 did mot resolve arm64 issue and made another issue with arm64 osx.
 Maybe related to https://github.com/xmake-io/xmake-repo/commit/585f9b0e62f78e48f67f636a51ba38e089467029

### DIFF
--- a/.github/workflows/ubuntu_arm64.yml
+++ b/.github/workflows/ubuntu_arm64.yml
@@ -34,4 +34,4 @@ jobs:
 
       - name: Tests
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ubuntu_arm64.yml
```diff
@@ -34,4 +34,4 @@ jobs:
 
       - name: Tests
         run: |
-          xmake l ./scripts/test.lua -D -a arm64-v8a -k ${{ matrix.kind }} -m ${{ matrix.mode }}
```
luadebug, line 37: Create another PR with same content maybe in order to avoid editing workflow. :/

PR #6505, opened 2025-03-03
icu4c: add support for cross-compilation

Doekin:


Doekin:
I’m having trouble figuring out why `icu4c` fails to build for the following targets. Any help or suggestions would be greatly appreciated:

- [ ] Android on Windows  
- [ ] Android on Ubuntu with NDK r27 (fine with NDK r22)  
- [ ] MinGW on macOS

Doekin:
It seems that the Android CI with NDK r27 is missing the `CXX` environment variable, which is causing the C++ compiler to be unavailable. Here’s the relevant error message produced by the configure script:

```
checking if we have a C++ compiler... no
configure: error: C++ compiler c++ does not work or no compiler found
```

When using NDK r22, the environment variables are properly set:

```
{ 
  CXX = "/home/runner/work/xmake-repo/xmake-repo/android-ndk-r22/toolchains/llvm/prebuilt/linux-x86_64/bin/clang++",
  ...
}
```

However, with NDK r27, the `CXX` variable appears to be missing from the environment:

```
{ 
  ASFLAGS = "--target=aarch64-none-linux-android30 ...",
  CC = "/home/runner/work/xmake-repo/xmake-repo/android-ndk-r27/toolchains/llvm/prebuilt/linux-x86_64/bin/clang",
  ...
}
```

Doekin:
Perfect, thanks a lot!

PR #6556, opened 2025-03-08
add_more control macro for spdlog

SnowinterCat:
1. Add spdlog's control macros SPDLOG_WCHAR_FILENAMES and SPDLOG_UTF8_TO_WCHAR_CONSOLE to xrepo's package management
2. Remove the restriction of shared build for spdlog on Windows platform.

### DIFF
--- a/packages/s/spdlog/xmake.lua
+++ b/packages/s/spdlog/xmake.lua
@@ -36,8 +36,8 @@ package("spdlog")
 
     if is_plat("windows") then
         add_configs("wchar",  {description = "Support wchar api.", default = false, type = "boolean"})
-        add_configs("wchar_filename",  {description = "Support wchar api.", default = false, type = "boolean"})
-        add_configs("wchar_console",  {description = "Support wchar api.", default = false, type = "boolean"})
+        add_configs("wchar_filenames",  {description = "Support wchar filenames.", default = false, type = "boolean"})
+        add_configs("wchar_console",  {description = "Support wchar output to console.", default = false, type = "boolean"})
     elseif is_plat("linux", "bsd") then
         add_syslinks("pthread")
     end
@@ -65,7 +65,7 @@ package("spdlog")
         if package:config("wchar") then
             package:add("defines", "SPDLOG_WCHAR_TO_UTF8_SUPPORT")
         end
-        if package:config("wchar_filename") then
+        if package:config("wchar_filenames") then
             package:add("defines", "SPDLOG_WCHAR_FILENAMES")
         end
         if package:config("wchar_console") then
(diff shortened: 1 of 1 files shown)


packages/s/spdlog/xmake.lua
```diff
@@ -27,15 +28,16 @@ package("spdlog")
...
     add_configs("std_format",      {description = "Use std::format instead of fmt library.", default = false, type = "boolean"})
     add_configs("fmt_external",    {description = "Use external fmt library instead of bundled.", default = false, type = "boolean"})
     add_configs("fmt_external_ho", {description = "Use external fmt header-only library instead of bundled.", default = false, type = "boolean"})
     add_configs("noexcept",        {description = "Compile with -fno-exceptions. Call abort() on any spdlog exceptions.", default = false, type = "boolean"})
 
     if is_plat("windows") then
-        add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
         add_configs("wchar",  {description = "Support wchar api.", default = false, type = "boolean"})
+        add_configs("wchar_filename",  {description = "Support wchar api.", default = false, type = "boolean"})
+        add_configs("wchar_console",  {description = "Support wchar api.", default = false, type = "boolean"})
```
qudix, line 40:
```suggestion
        add_configs("wchar_filenames",  {description = "Support wchar filenames.", default = false, type = "boolean"})
        add_configs("wchar_console",  {description = "Support wchar output to console.", default = false, type = "boolean"})
```

packages/s/spdlog/xmake.lua
```diff
@@ -27,15 +28,16 @@ package("spdlog")
 
     add_patches("v1.11.0", path.join(os.scriptdir(), "patches", "v1.11.0", "fmt10.patch"), "61efa804845141ffa86532d9be7103d4dc8185e96de69d5efca42ebd7058e13d")
 
-    add_configs("header_only",     {description = "Use header only version.", default = true, type = "boolean"})
+    add_configs("header_only",     {description = "Use header only version.", default = false, type = "boolean"})
```
qudix, line 31: I don't think the maintainers will allow this change. As much as I also want it to default to false, changing the default behavior can cause build failures for consumers who either knowingly or unknowingly relied on it being header only by default.

SnowinterCat: > I don't think the maintainers will allow this change. As much as I also want it to default to false, changing the default behavior can cause build failures for consumers who either knowingly or unknowingly relied on it being header only by default.

This commit is just a test, otherwise header-only wouldn't do a SHARED build. After the ci runs, I'll revert this try

packages/s/spdlog/xmake.lua
SnowinterCat: thanks. I forget this when ctrl+c/v.

PR #6559, opened 2025-03-08
Imgui backend libsdl3

SnowinterCat:
try to add libsdl3 backend to imgui
#6327

### DIFF
--- a/packages/i/imgui/port/xmake.lua
+++ b/packages/i/imgui/port/xmake.lua
@@ -12,6 +12,8 @@ option("opengl3",          {showmenu = true,  default = false})
 option("glad",             {showmenu = true,  default = false})
 option("sdl2",             {showmenu = true,  default = false})
 option("sdl2_renderer",    {showmenu = true,  default = false})
+option("sdl3",             {showmenu = true,  default = false})
+option("sdl3_renderer",    {showmenu = true,  default = false})
 option("vulkan",           {showmenu = true,  default = false})
 option("win32",            {showmenu = true,  default = false})
 option("wgpu",             {showmenu = true,  default = false})
@@ -31,6 +33,9 @@ if has_config("sdl2_renderer") then
     add_requires("libsdl2 >=2.0.17")
 elseif has_config("sdl2") then
     add_requires("libsdl2")
+end
+if has_config("sdl3") or has_config("sdl3_renderer") then
+    add_requires("libsdl3")
 end
 
 if has_config("vulkan") then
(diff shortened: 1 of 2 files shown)


packages/i/imgui/xmake.lua
```diff
@@ -116,6 +119,9 @@ package("imgui")
         if package:config("sdl2") or package:config("sdlrenderer") then
             package:config_set("sdl2_renderer", true)
         end
+        if package:config("sdl3") or package:config("sdlrenderer") then
+            package:config_set("sdl3_renderer", true)
+        end
```
qudix, line 124:
```suggestion
```
This section is for backwards compat with old configs, you don't need this. `sdlrenderer` was the old name for `sdl2_renderer`

packages/i/imgui/xmake.lua
```diff
@@ -92,6 +92,9 @@ package("imgui")
     add_configs("sdl2",             {description = "Enable the sdl2 backend with sdl2_renderer", default = false, type = "boolean"})
     add_configs("sdl2_no_renderer", {description = "Enable the sdl2 backend without sdl2_renderer", default = false, type = "boolean"})
     add_configs("sdl2_renderer",    {description = "Enable the sdl2 renderer backend", default = false, type = "boolean"})
+    add_configs("sdl3",             {description = "Enable the sdl3 backend with sdl3_renderer", default = false, type = "boolean"})
+    add_configs("sdl3_no_renderer", {description = "Enable the sdl3 backend without sdl3_renderer", default = false, type = "boolean"})
```
qudix, line 96:
```suggestion
```
the `_no_renderer` was to work around the fact that `sdl2` has the renderer enabled by default for compat reasons, sdl3 doesn't need this when it can have sane defaults to start with. you can remove this

packages/i/imgui/xmake.lua
```diff
@@ -170,6 +179,8 @@ package("imgui")
             glad             = package:config("opengl3") and (not package:gitref() and package:version():lt("1.84")),
             sdl2             = package:config("sdl2") or package:config("sdl2_no_renderer"),
             sdl2_renderer    = package:config("sdl2_renderer"),
+            sdl3             = package:config("sdl3") or package:config("sdl3_no_renderer"),
```
qudix, line 182:
```suggestion
            sdl3             = package:config("sdl3"),
```

packages/i/imgui/port/xmake.lua
```diff
@@ -119,6 +124,28 @@ target("imgui")
         add_packages("libsdl2")
     end
 
+    if has_config("sdl3") then
+        if os.exists("backends/imgui_impl_sdl3.cpp") then
```
waruqi, line 128: we need not call `os.exists`

waruqi:
>  add_configs("sdl3",             {description = "Enable the sdl3 backend with sdl3_renderer", default = false, type = "boolean"})
    add_configs("sdl3_no_renderer", {description = "Enable the sdl3 backend without sdl3_renderer", default = false, type = "boolean"})
    add_configs("sdl3_renderer",    {description = "Enable the sdl3 renderer backend", default = false, type = "boolean"})

What is the difference between these three, can't they be simplified into two or one configuration?

and sdl2 too.

packages/i/imgui/xmake.lua
SnowinterCat: OK, I see.

packages/i/imgui/xmake.lua
SnowinterCat: I've found that sdl2_no_renderer=true does basically the same thing as sdl2_renderer=false. It looks like they should be combined into one

PR #6568, opened 2025-03-10
Auto-update aws-c-s3 to v0.7.12

waruqi:
New version of aws-c-s3 detected (package version: v0.7.7, last github version: v0.7.12)


Altina-oz:
maybe related：
https://stackoverflow.com/questions/65829295/aws-sdk-c-install-using-cmake-on-ubuntu-problem-with-installing-third-party-l

PR #6631, opened 2025-03-18
screen_capture_lite: add missing links for windows

Chi-EEE:


star-hengxing:
See https://github.com/xmake-io/xmake-repo/pull/6619

Issue #6049, opened 2024-12-25
Add steam-audio package


ethindp:
### Is your feature request related to a problem? Please describe.

I'd like to use steam audio in my software (I'm not familiar enough with xmake yet to get this added myself).

### Describe the solution you'd like

Steam audio should be in the repo.

### Additional context

The steam audio build process is a bit convoluted. It requires cmake and python, but it's fully documented [here](https://github.com/ValveSoftware/steam-audio/blob/master/core/doc/build-instructions.rst).

PR #6634, opened 2025-03-18
steam-audio: add package

star-hengxing:
```lua
add_requires("steam-audio[fft=pffft,mkl,embree,radeonrays]")
```

```c++
#include <phonon.h>
```

close https://github.com/xmake-io/xmake-repo/issues/6049

### DIFF
--- a/packages/s/steam-audio/patch.lua
+++ b/packages/s/steam-audio/patch.lua
@@ -1,4 +1,19 @@
+function _unsafe_patch(package)
+    if package:is_plat("macosx") then
+        io.replace("CMakeLists.txt", [[set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13")]], "", {plain = true})
+    end
+
+    -- clang unsupported this flag https://github.com/llvm/llvm-project/issues/25059
+    -- But build failed on ubuntu with libstdc++-13
+    io.replace("CMakeLists.txt", "add_compile_options(-fabi-version=6)", "", {plain = true})
+    if package:is_cross() then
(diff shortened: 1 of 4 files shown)


packages/s/steam-audio/patch.lua
```diff
@@ -0,0 +1,62 @@
+function _unsafe_patch(package)
+    if package:is_plat("macosx") then
+        io.replace("CMakeLists.txt", [[set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64")]], "", {plain = true})
+        io.replace("CMakeLists.txt", [[set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13")]], "", {plain = true})
+    end
+
+    -- clang unsupported this flag https://github.com/llvm/llvm-project/issues/25059
+    -- But build failed on ubuntu with libstdc++-13
+    io.replace("CMakeLists.txt", "add_compile_options(-fabi-version=6)", "", {plain = true})
```
star-hengxing, line 9: I’m not sure if these patches are correct. Any idea? @SirLynix

SirLynix: with that many patches I wonder if it wouldn't be better to port it to xmake?

otherwise I agree with this change

SirLynix: I guess it could be an issue if somehow this steam-audio binary ends up being used by another non-xmake binary (static linking), can't really tell

star-hengxing: I add an option to enable/disable the patch.

PR #6638, opened 2025-03-19
Test zoe

luadebug:
note: the following packages are unsupported on bsd/x86_64:
note: the following packages are unsupported on wasm/wasm32:
  -> libcurl 8.11.0 [2m[from:zoe, license:MIT]
Curl required dep.


luadebug:
Maybe there is better way to wrap android NDK 21 support, IDK.

PR #6653, opened 2025-03-20
Test melon

luadebug:
Windows has not DLL build support.

MinGW32/64 or UCRT would not work, but Clang64 is working. I do not know upon Mac OS X MinGW-W64 so Ill place up placeholder there.
```lua
        if is_subhost("macosx") then
            raise("package(melon) is unsupported on Mac OS X subhost.")
        end
```

### DIFF
--- a/packages/m/melon/xmake.lua
+++ b/packages/m/melon/xmake.lua
@@ -6,20 +6,23 @@ package("melon")
     add_urls("https://github.com/Water-Melon/Melon.git")
     add_versions("2025.01.18", "9df92922ab384295380d4414493e69983671dbf5")
 
-    if is_plat("mingw") then
-        add_deps("mman_win32")    
-    end
+    on_check(function (package)
+        if is_plat("mingw") and is_subhost("msys") then
+            raise("package(melon) is unsupported on MinGW64/UCRT64. Use CLANG64 Shell.")
+        end
+        if is_plat("android") then
+            import("core.tool.toolchain")
+            local ndk = toolchain.load("ndk", {plat = package:plat(), arch = package:arch()})
+            local ndk_sdkver = ndk:config("ndk_sdkver")
+            assert(ndk_sdkver and tonumber(ndk_sdkver) >= 26, "package(melon): need ndk api level >= 26 for android")
+        end
+    end)
 
     on_install(function (package)
         io.replace("include/mln_utils.h", 
         "#if defined(__APPLE__) || defined(MSVC) || defined(__wasm__) || defined(__FreeBSD__)", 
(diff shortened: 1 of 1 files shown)


packages/m/melon/xmake.lua
```diff
@@ -0,0 +1,71 @@
...
+    add_versions("2025.01.18", "9df92922ab384295380d4414493e69983671dbf5")
+
+    if is_plat("windows") then
+        add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
+    end
+
+    on_check(function (package)
+        if is_plat("mingw") and is_subhost("msys") then
+            raise("package(melon) is unsupported on MinGW64/UCRT64. Use CLANG64 Shell.")
+        end
```
luadebug, line 16: Can it be checked directly as CLANG64?

packages/m/melon/xmake.lua
```diff
@@ -0,0 +1,71 @@
...
+        if is_plat("android") then
+            import("core.tool.toolchain")
+            local ndk = toolchain.load("ndk", {plat = package:plat(), arch = package:arch()})
+            local ndk_sdkver = ndk:config("ndk_sdkver")
+            assert(ndk_sdkver and tonumber(ndk_sdkver) >= 26, "package(melon): need ndk api level >= 26 for android")
+        end
+    end)
```
luadebug, line 23: I am not sure upon android as
   73 |             pthread_cancel(t->tids[i]);
Is not into bionic. Might be good reason to limit android.

packages/m/melon/xmake.lua
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    on_check("mingw", function (package)
+        if is_subhost("macosx") or is_subhost("msys") then
+            raise("package(melon) is unsupported on MinGW64/UCRT64. Use CLANG64 Shell.")
+        end
+    end)
+
+    on_install("!android", function (package)
+        if is_plat("windows") then
```
waruqi, line 20: package:is_plat

packages/m/melon/xmake.lua
star-hengxing:
```lua
        msystem = os.getenv("MSYSTEM")
        if msystem then
            msystem = msystem:lower()
        end
```

packages/m/melon/xmake.lua
luadebug: Thanks, fixed.

packages/m/melon/xmake.lua
luadebug: OK, done. I dunno upon Mac OS X & MinGW. So Ill just place placeholder.

PR #6666, opened 2025-03-21
Test think-cell-library

luadebug:
```cpp
#include <tc/>
```
b2 does not support WASM.
Release tag on GitHub is not x.y.z it's like x.y in case of this header only library.

### DIFF
--- a/packages/t/think-cell-library/xmake.lua
+++ b/packages/t/think-cell-library/xmake.lua
@@ -17,7 +17,7 @@ package("think-cell-library")
     end)
 
     on_install("!wasm", function (package)
-        os.rm("tc/string/spirit")
+        os.rm("**.cpp", "tc/string/spirit")
         io.replace("tc/string/spirit.h", [[#include "spirit/x3.hpp"]], "#include <boost/spirit/home/x3.hpp>", {plain = true})
         os.cp("tc", package:installdir("include"))
     end)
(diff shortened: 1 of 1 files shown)


packages/t/think-cell-library/xmake.lua
```diff
@@ -0,0 +1,45 @@
...
+        io.writefile("xmake.lua", [[
+            add_rules("mode.release", "mode.debug")
+            set_languages("c++20")
+            add_requires("boost", {configs = {filesystem = true, container = true}})
+            target("think-cell-library")
+                set_kind("headeronly")
+                add_packages("boost")
+                add_headerfiles("(tc/**.h)")
+            ]])
+        import("package.tools.xmake").install(package)
```
star-hengxing, line 32: You can directly use `os.cp`

luadebug:
Maybe it should not install .t.cpp, only .h

packages/t/think-cell-library/xmake.lua
```diff
@@ -19,17 +19,7 @@ package("think-cell-library")
...
-            add_rules("mode.release", "mode.debug")
-            set_languages("c++20")
-            add_requires("boost", {configs = {filesystem = true, container = true}})
-            target("think-cell-library")
-                set_kind("headeronly")
-                add_packages("boost")
-                add_headerfiles("(tc/**.h)")
-            ]])
-        import("package.tools.xmake").install(package)
+        os.cp("tc", path.join(package:installdir("include"), "tc"))
```
waruqi:
```
os.cp("tc", package:installdir("include"))
```

luadebug: Done https://github.com/xmake-io/xmake-repo/pull/6666/commits/f7637cdf92938545b3985073b0fa6be197086022 installs headers only.

PR #6705, opened 2025-03-25
Test woff2

luadebug:
```xmake
if is_plat("windows") then
end
```
mingw@macosx https://stackoverflow.com/questions/8878676/compile-error-g-error-trying-to-exec-cc1plus-execvp-no-such-file-or-dir ?
https://github.com/xmake-io/xmake/issues/5191
Exactly 
```
x86_64-w64-mingw32-g++: fatal error: cannot execute 'cc1plus': posix_spawnp: No such file or directory
```

### DIFF
--- a/packages/w/woff2/xmake.lua
+++ b/packages/w/woff2/xmake.lua
@@ -8,10 +8,23 @@ package("woff2")
 
     add_versions("v1.0.2", "add272bb09e6384a4833ffca4896350fdb16e0ca22df68c0384773c67a175594")
 
-    add_deps("cmake", "brotli")
+    if is_plat("windows") then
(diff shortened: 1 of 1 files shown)


packages/w/woff2/xmake.lua
```diff
@@ -0,0 +1,44 @@
+package("woff2")
+    set_homepage("https://github.com/google/woff2")
+    set_description("Font compression reference code.")
+    set_license("MIT")
+
+    add_urls("https://github.com/google/woff2/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/google/woff2.git")
```
star-hengxing, line 7: If git url has submodule, you need to set it false
done. https://github.com/xmake-io/xmake-repo/commit/12c88339a38175f9d28d8896c6b2a32d1ab0d015

PR #6743, opened 2025-03-29
webui: add new config

luadebug:
Resolves https://github.com/xmake-io/xmake-repo/pull/6727


packages/w/webui/port/xmake.lua
```diff
@@ -0,0 +1,17 @@
+add_rules("mode.debug", "mode.release")
+
+option("log",              {showmenu = true,  default = true})
```
waruqi, line 3: remove spaces

PR #6764, opened 2025-03-31
add zstr.

Redbeanw44602:


packages/z/zstr/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("zstr")
+    set_kind("library", {headeronly = true})
+
+    set_homepage("https://github.com/mateidavid/zstr")
+    set_description("A C++ header-only ZLib wrapper.")
+    set_license("MIT")
+
+    add_urls("https://github.com/mateidavid/zstr.git")
+    add_versions("1.0.7", "755da7890ea22478a702e3139092e6c964fab1f5")
```
waruqi, line 9: please add a tar url.

https://github.com/mateidavid/zstr/archive/refs/tags/v1.0.7.tar.gz

Redbeanw44602:
Re-running should resolve the issue, the network infrastructure of the macOS runners shares a single IP, so it is more prone to errors.

PR #6765, opened 2025-03-31
funchook: add package

Redbeanw44602:


packages/f/funchook/xmake.lua
```diff
@@ -0,0 +1,57 @@
...
+    add_urls("https://github.com/kubo/funchook.git")
+    add_versions("1.1.3", "88388db3c69e16c1560fee65c6857d75f5ce6fd5")
+
+    add_patches("*", "patches/try-fix-function-visibility.patch", "5b505ad24332320f3970a6cb56b5f550b01b9c80aa14cea0fea74ac77f1fc8f3")
+
+    add_configs("disasm", {description = "Disassembler engine.", default = nil, type = "string", values = {"capstone", "distorm", "zydis"}})
+
+    add_deps("cmake")
+    on_load(function(package)
+        if is_plat("windows") then
```
luadebug, line 15:
```suggestion
        if package:is_plat("windows") then
```

packages/f/funchook/xmake.lua
```diff
@@ -0,0 +1,57 @@
...
+            package:add("syslinks", "psapi")
+        else
+            -- unix
+            package:add("syslinks", "dl")
+        end
+        if not package:config("disasm") then
+            -- default disasm engine.
+            if is_arch("arm64") then
```
luadebug, line 23:
```suggestion
            if package:is_arch("arm64") then
```

Redbeanw44602:
This is a bit strange, xmake didn't automatically install cmake on fedora and archlinux, I added the cmake installation in my own [test repo](https://github.com/Redbeanw44602/test-xmake-repo/pull/3/commits/0982bf491bd65085fc87c091eb1464886e19886c) and it solved the problem, I guess this may be an xmake bug?

luadebug:
Also why
```
    "C:\Users\runneradmin\AppData\Local\.xmake\packages\f\funchook\1.1.3\8e12501f49a545b19b9f40b55b8945b9\lib\funchook.lib",
    "C:\Users\runneradmin\AppData\Local\.xmake\packages\f\funchook\1.1.3\8e12501f49a545b19b9f40b55b8945b9\lib\funchook_dll.lib",
    "C:\Users\runneradmin\AppData\Local\.xmake\packages\f\funchook\1.1.3\8e12501f49a545b19b9f40b55b8945b9\bin\funchook.dll", 
```
I thought it should be exactly one `.lib` and one `.dll`

Redbeanw44602:
@luadebug I don't quite understand, `shared = true` will cause both dynamic and static libraries to be built under windows, and only static libraries to be built under linux, is that the case?

luadebug:
> @luadebug I don't quite understand, `shared = true` will cause both dynamic and static libraries to be built under windows, and only shared libraries to be built under linux, is that the case?
> 

Install should install only one target, not both targets.

luadebug:
"C:\Users\runneradmin\AppData\Local\.xmake\packages\f\funchook\1.1.3\8e12501f49a545b19b9f40b55b8945b9\lib\funchook_dll.lib",
    "C:\Users\runneradmin\AppData\Local\.xmake\packages\f\funchook\1.1.3\8e12501f49a545b19b9f40b55b8945b9\bin\funchook.dll", 
Well looks resolved now only one target being installed.

packages/f/funchook/xmake.lua
```diff
@@ -0,0 +1,61 @@
+package("funchook")
+    set_homepage("https://github.com/kubo/funchook")
+    set_description("Hook function calls by inserting jump instructions at runtime.")
+    set_license("GPLv2")
+
+    add_urls("https://github.com/kubo/funchook.git")
+    add_versions("1.1.3", "88388db3c69e16c1560fee65c6857d75f5ce6fd5")
```
luadebug, line 7:
```suggestion
   add_urls("https://github.com/kubo/funchook/archive/refs/tags/$(version).tar.gz",
             "https://github.com/kubo/funchook.git")

    add_versions("v1.1.3", "4b0195e70524237e222dc34c53ac25e12677bb936e64eefe33189931688444c4")
```

packages/f/funchook/xmake.lua
```diff
@@ -0,0 +1,61 @@
+package("funchook")
+    set_homepage("https://github.com/kubo/funchook")
+    set_description("Hook function calls by inserting jump instructions at runtime.")
+    set_license("GPLv2")
```
luadebug, line 4:
```suggestion
    set_license("GPL-2.0-or-later")
```

luadebug:
Well it looks like you have to de-bundle `distorm`.
https://github.com/kubo/funchook/blob/v1.1.3/CMakeLists.txt#L114-L137

Redbeanw44602:
I will create a patch to do this.

PR #6805, opened 2025-04-03
libxml2: resolve missing dependencies when external libraries are needed for optional features

Doekin:
~~This change enables `iconv` support by default for `libxml2`, aligning with its default configuration options.~~

This PR addresses an issue where enabling `iconv` can cause build failures for Android due to missing dependencies. The error typically appears as:

```console
ld: error: undefined symbol: iconv
>>> referenced by encoding.c:1122 (/home/runner/.xmake/cache/packages/2504/l/libxml2/v2.14.0/source/encoding.c:1122)
>>>               encoding.c.o:(xmlIconvConvert) in archive /home/runner/.xmake/packages/l/libxml2/v2.14.0/cd35469623984bc18d849f390d29bfb2/lib/libxml2.a
```


star-hengxing:
xrepo generally adheres to the principle of minimal dependencies and does not necessarily need to synchronize with upstream options.

Doekin:
OK, I'll revert the last commit.

luadebug:
Would it be convenient to define `local packagedeps = {}` and fill it accordingly to used options and pass it into `(package, configs, {packagedeps = packagedeps})`?

Doekin:
Thank you for the suggestion. It should generally be fine to reference non-existing dependencies. If a library is not installed as a dependency, it won't take effect or cause errors during the build process. Therefore, passing all possible dependencies might be a simpler and more concise approach. However, I'm open to further discussion if you think a more selective approach would be better.

packages/l/libxml2/xmake.lua
```diff
@@ -142,7 +139,7 @@ package("libxml2")
         if lzma and not lzma:config("shared") then
             table.insert(cxflags, "-DLZMA_API_STATIC")
         end
-        import("package.tools.cmake").install(package, configs, {cxflags = cxflags, shflags = shflags})
+        import("package.tools.cmake").install(package, configs, {cxflags = cxflags, shflags = shflags, packagedeps = {"libiconv", "icu4c", "zlib", "xz", "readline"}})
```
waruqi, line 142: We should pass dependencies as needed.

```lua
local packagedeps = {}
if package:config("zlib") then
    table.insert(packagedeps, "zlib")
end
```

In addition, packagedeps are usually not needed, so cmake will automatically find them through add_deps, and xmake will automatically set CMAKE_PREFIX_PATH and PKG_CONFIG_PATH to let cmake/pkg-config find these dependent libraries.

Only when cmake really cannot find a library, packagedeps will be used as a hack to support it.
So don't pass all packagedeps at once without any test verification.

Doekin: Understood. I'll investigate the root cause of the build failures for Android when enabling `iconv`, rather than using `packagedeps` without proper testing. Thank you for the guidance.

PR #6822, opened 2025-04-05
python: refactor & update.

Redbeanw44602:
- drop py2.7.18, py3.7.9 
> They are all end-of-life. And it failed to pass the test on macosx due to some problems.
- drop py3.11.9, py3.13.0
> They were not present in the pre-builds, which resulted in different versions being available on different platforms and tests failing.
- add py3.13.2.
- add config for all *features*.
- add config for potentially useful *packages*.
> See cpython configure script for details.
- setup pip (wheel, setuptool) only when `ensurepip` is enabled.
- use openssl3 by default.
- cross-compilation issues have been resolved.
- support for more platforms (android, iphoneos, wasm).
> android, iphoneos, wasm support should be complete, but are temporarily disabled due to issues with some dependencies in xrepo.

### DIFF
--- a/packages/p/python/constants.lua
+++ b/packages/p/python/constants.lua
@@ -1,6 +1,4 @@
-constants = {}
-
---- configure
+local constants = {}
 
 -- only options listed here are yes/no configurations, underscores will be replaced with hyphens.
 function constants.get_yn_features()
@@ -43,8 +41,6 @@ function constants.get_supported_packages()
         "emscripten_target"
     }
 end
-
---- module
 
 function main()
     return constants
--- a/packages/p/python/constants.lua
+++ b/packages/p/python/constants.lua
@@ -1,0 +1,51 @@
+constants = {}
+
+--- configure
+
+-- only options listed here are yes/no configurations, underscores will be replaced with hyphens.
+function constants.get_yn_features()
+    return {
+        "wasm_dynamic_linking", -- 3.11
+        "wasm_pthreads", -- 3.11
+        "shared",
+        "profiling",
+        "gil", -- 3.13
+        "pystats", -- 3.11
+        "optimizations", -- 3.6
+        "bolt", -- 3.12
+        "loadable_sqlite_extensions", -- 3.6
+        "ipv6",
+        "test_modules" -- 3.10
+    }
+end
+
+function constants.get_all_features()
+    return table.join(
+        constants.get_yn_features(),
+        {
+            "universalsdk",
+            "framework",
+            "experimental_jit", -- 3.13
+            "big_digits"
+        }
+    )
+end
+
+function constants.get_supported_packages()
+    return {
+        "framework_name",
+        "app_store_compliance",
+        "hash_algorithm",
+        "builtin_hashlib_hashes",
+        "ssl_default_suites",
+        "lto",
+        "ensurepip",
+        "emscripten_target"
+    }
+end
+
+--- module
+
+function main()
+    return constants
+end
--- a/packages/p/python/constants.lua
+++ b/packages/p/python/constants.lua
@@ -42,6 +42,6 @@ function constants.get_supported_packages()
     }
 end
 
-function main()
+function get_python_package_constants()
     return constants
 end
--- a/packages/p/python/xmake.lua
+++ b/packages/p/python/xmake.lua
@@ -5,7 +5,7 @@ package("python")
 
     -- enable-FEATURE
     includes(path.join(os.scriptdir(), "constants.lua"))
-    for _, feature in ipairs(constants.get_yn_features()) do
+    for _, feature in ipairs(get_python_package_constants()) do
         -- if the user doesn't pass it (nil), we won't pass it either.
         add_configs(feature, {description = "Enable " .. feature .. ".", default = nil, type = "boolean"})
     end
@@ -169,7 +169,7 @@ package("python")
 
     --- android, iphoneos, wasm unsupported: dependencies not resolved.
     on_install("macosx", "linux", "bsd", function (package)
-        local constants = import('constants')()
+        local constants = import('constants').get_python_package_constants()
         function opt2cfg(cfg)
             if type(cfg) == "boolean" then
                 return cfg and 'yes' or 'no'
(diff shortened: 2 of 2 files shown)


Redbeanw44602:
Any updates?

packages/p/python/xmake.lua
```diff
@@ -80,21 +87,53 @@ package("python")
...
+            package:add("deps", "libffi") -- py module 'ctypes'
+            if pkgver:ge("3.10") then -- sqlite3, py module 'sqlite3'
+                package:add("deps", "sqlite3 >=3.7.15")
+            elseif pkgver:ge("3.13") then
+                package:add("deps", "sqlite3 >=3.15.2")
+            else
+                package:add("deps", "sqlite3")
+            end
+        end
+        if is_plat("linux", "macosx") then
```
waruqi, line 112: use package:is_plat

packages/p/python/constants.lua
```diff
@@ -0,0 +1,51 @@
+constants = {}
+
+--- configure
```
waruqi, line 3: remove unused comments

packages/p/python/xmake.lua
```diff
@@ -3,27 +3,32 @@ package("python")
...
+
+    -- with-PACKAGE
+    add_configs("framework_name", {description = "(macOS) Specify the name for the python framework.", default = nil, type = "string"})
+    add_configs("app_store_compliance", {description = "(macOS) Enable any patches required for compiliance with app stores.", default = nil, type = "boolean"}) -- 3.13
+    add_configs("hash_algorithm", {description = "Select hash algorithm for use in Python/pyhash.c", default = nil, type = "string", values = {"fnv", "siphash13", "siphash24"}}) -- 3.4, 3.11
+    add_configs("builtin_hashlib_hashes", {description = "Builtin hash modules. (md5, sha1, sha2, sha3, blake2)", default = nil, type = "string"}) -- 3.9
+    add_configs("ssl_default_suites", {description = "Override default cipher suites string. (python, openssl)", default = nil, type = "string"}) -- 3.7, 3.10
+    add_configs("lto", {description = "Enable Link-Time-Optimization in any build.", default = nil, values = {true, false, "full", "thin", "no", "yes"}})
+    add_configs("ensurepip", {description = "'install' or 'upgrade' using bundled pip", default = nil, values = {true, false, "upgrade", "install", "no"}}) -- 3.6
```
SirLynix, line 24: it should be enabled by default to not break retrocompatibility

packages/p/python/xmake.lua
```diff
@@ -50,26 +70,13 @@ package("python")
...
-    end
-
-    if is_host("linux", "bsd") then
-        add_deps("libffi", "zlib", {host = true, private = true})
-        add_syslinks("util", "pthread", "dl")
-    end
-
-    on_load("@windows", "@msys", "@cygwin", function (package)
-
+    on_load("windows", "msys", "mingw", "cygwin", function (package)
```
SirLynix, line 79: why changing from host platforms to target platforms?

packages/p/python/xmake.lua
```diff
@@ -80,21 +87,53 @@ package("python")
         package:addenv("PATH", "Scripts")
     end)
 
-    on_load("@macosx", "@linux", "@bsd", function (package)
-        local version = package:version()
+    on_load("macosx", "linux", "bsd", function (package)
```
SirLynix, line 90: same

packages/p/python/xmake.lua
```diff
@@ -105,15 +144,11 @@ package("python")
...
-
-        if package:config("headeronly") then
-            package:set("links", "")
-        end
     end)
 
     on_fetch("fetch")
 
-    on_install("@windows|x86", "@windows|x64", "@msys", "@cygwin", function (package)
+    on_install("windows|x86", "windows|x64", "msys", "mingw", "cygwin", function (package)
```
SirLynix, line 151: same

packages/p/python/xmake.lua
```diff
@@ -124,25 +159,45 @@ package("python")
...
+            os.vrunv(python, {"-m", "pip", "install", "--upgrade", "--force-reinstall", "pip"})
+            os.vrunv(python, {"-m", "pip", "install", "--upgrade", "setuptools"})
+            os.vrunv(python, {"-m", "pip", "install", "wheel"})
+        end
     end)
 
-    on_install("@macosx", "@bsd", "@linux", function (package)
-        local version = package:version()
+    --- android, iphoneos, wasm unsupported: dependencies not resolved.
+    on_install("macosx", "linux", "bsd", function (package)
```
SirLynix, line 171: same

SirLynix (changes requested):

packages/p/python/xmake.lua
Redbeanw44602: If I understand correctly, using `@` actually breaks cross-compilation, and ensurepip is disabled by default for the same reason. Often we may only need to use CPython's API without a full Python environment.
Is there any way to achieve both ends?

SirLynix: It doesn't break cross compilation, it makes it target a host architecture. @linux means you target the linux host, even if it builds for Android or for Windows using mingw.
This is because the python package is about having a running python interpreter able to execute python programs such as meson, on the host.

I think we should make a different package for cpython API or anything the program links to

waruqi: if use python as cli program. we should use `@host` , please do not modify it.

Even for cross-compilation, it will always use the python on the host instead of the python of the cross-compilation platform, otherwise python will not run.

Redbeanw44602: I see what you mean. But the purpose of cross-compilation is not to "run on the host", but to generate code for other platforms.

Redbeanw44602: Maybe we should create another package focused on users who need the CPython API, how about `libpython`?

waruqi: Cross compile python? This will be a big job, as many of its dependent libraries will need to enable cross-compilation support.

If you want to do it, you can add libpython package to do it.

Redbeanw44602: hmm, it does it now at least on windows (precompiled), linux, macosx and bsd. As for android, iphoneos and wasm it is indeed temporarily disabled due to some dependency issues (see line 170).

I will make another pull request later and then close this one.

PR #6828, opened 2025-04-06
Auto-update hypre to v2.33.0

waruqi:
New version of hypre detected (package version: v2.32.0, last github version: v2.33.0)


apocelipes:
Close because of #9313.

PR #6829, opened 2025-04-06
Update drogon to v1.9.10

RigelNana:


RigelNana:
The only failing section is because the xrepo-mirror's NASM version is outdated (2.16.01), while this one is using 2.16.03, which could not be found.

RigelNana:
The only failing section is caused by a network error (which can be resolved by rerunning the job), and additionally, the NASM version in xmake-mirror is outdated (v2.16.01), resulting in the specified version (v2.16.03) being unavailable when using the mirror.

PR #6834, opened 2025-04-07
libpng: add older version

Elite-stay:


star-hengxing:
code too long, please reference
https://github.com/xmake-io/xmake-repo/blob/2ddcfb362ed95e055844d2813b97a1f8728cda6f/packages/z/zasm/xmake.lua#L40-L70

PR #6854, opened 2025-04-08
v8: fix and update to 13.5.212.10

poirierlouis:
**Requires:**
xmake v2.9.9 to use `target` option when building with GN tool.

**Features:**
- support latest stable version 13.5.212.10
- tested on Windows only
- fix syntax to configure "runtimes", only MT for now
- only build static monolith flavor
- patch to fix build configuration of a dependency (third_party/partition_alloc/)
- erase previous version 10.0.58

**Notes:**
- `add_patches` is not used, because afaik, it is not meant to run in this context. `gclient` is ran during `on_install` callback. `add_patches` would try to apply the patch before execution of `gclient`, that is before files to be patched are even cloned/pulled.

**Questions:**
(1) Is testing on Linux + macOS required to merge?
(2) Is backward compatibility with version 10.0.58 required to merge?

### DIFF
--- a/packages/v/v8/xmake.lua
+++ b/packages/v/v8/xmake.lua
@@ -53,17 +53,20 @@ package("v8")
         os.vrunv(gclient, {"sync", "--reset", "--delete_unversioned_trees", "-v"}, {envs = envs})
 
         -- Fix GN issue
-        local patchPath = path.join(package:scriptdir(), "patches", "fix_chromium.patch")
-        local sourcePath = path.join(package:cachedir(), "source", "v8")
-        local cwd = os.cd(sourcePath)
+        local patchPath = path.join(package:scriptdir(), "patches", package:version() .. ".patch")
 
-        os.execv("git", {"apply", patchPath})
+        if os.exists(patchPath) then
+            local sourcePath = path.join(package:cachedir(), "source", "v8")
+            local cwd = os.cd(sourcePath)
 
-        os.cd(path.join("third_party", "partition_alloc"))
-        os.execv("git", {"add", path.join("src", "partition_alloc", "BUILD.gn")})
-        os.execv("git", {"commit", "-m", "xmake-patch-01"})
+            os.execv("git", {"apply", patchPath})
 
-        os.cd(cwd)
+            os.cd(path.join("third_party", "partition_alloc"))
+            os.execv("git", {"add", path.join("src", "partition_alloc", "BUILD.gn")})
+            os.execv("git", {"commit", "-m", "xmake-patch-01"})
+
+            os.cd(cwd)
+        end
 
         -- Setup args.gn
         local configs = {
--- a/packages/v/v8/xmake.lua
+++ b/packages/v/v8/xmake.lua
@@ -1,9 +1,11 @@
 package("v8")
-    set_homepage("https://chromium.googlesource.com/v8/v8.git")
+    set_homepage("https://v8.dev")
     set_description("V8 JavaScript Engine")
 
     add_urls("https://github.com/v8/v8.git")
-    add_versions("10.0.58", "d75903764c8547b6fc35c7a3fe4991320be03135")
+    add_versions("13.3.415.19", "fe051262efbbd92479a08436f733eba9f756e008")
+    add_versions("13.4.114.19", "a57459aad9ec3ed5f78d9ded700d52e31029efd2")
+    add_versions("13.5.212.10", "e2591684c45463aa1e46ebefc3fd35deee63f37c")
 
     add_deps("depot_tools")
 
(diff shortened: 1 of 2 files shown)


packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,68 @@ package("v8")
...
+        -- Update repository and dependencies
+        -- Clean any local changes to apply patches
+        os.vrunv(gclient, {"sync", "--reset", "--delete_unversioned_trees", "-v"}, {envs = envs})
+
+        -- Fix GN issue
+        local patchPath = path.join(package:scriptdir(), "patches", "fix_chromium.patch")
+        local sourcePath = path.join(package:cachedir(), "source", "v8")
+        local cwd = os.cd(sourcePath)
+
+        os.execv("git", {"apply", patchPath})
```
waruqi, line 60: why not use `add_patches`?

poirierlouis: I mentioned that in `Notes`. Is it supposed to work only when `add_patches` is called, like I can call it instead of this custom code and it should work the same?

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,68 @@ package("v8")
...
+
+        -- Prevent long path issue on Windows
+        os.vrun("git config --global core.longpaths true")
+
+        -- Update repository and dependencies
+        -- Clean any local changes to apply patches
+        os.vrunv(gclient, {"sync", "--reset", "--delete_unversioned_trees", "-v"}, {envs = envs})
+
+        -- Fix GN issue
+        local patchPath = path.join(package:scriptdir(), "patches", "fix_chromium.patch")
```
waruqi, line 56: where is this patch file?

packages/v/v8/xmake.lua
waruqi: it will be patched before on_install

packages/v/v8/xmake.lua
waruqi: And different versions of patches should be different, we need to handle them by version

packages/v/v8/xmake.lua
poirierlouis: Hum, so it won't work like I said. Maybe an alternative would be to update/sync repository with `gclient` right after `on_download`? Can I implement this callback to execute the default behaviour (like a `super` call) and then run `gclient` sync? I guess a patch is applied between on_download and on_install?

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,71 @@ package("v8")
...
+
+        -- Prevent long path issue on Windows
+        os.vrun("git config --global core.longpaths true")
+
+        -- Update repository and dependencies
+        -- Clean any local changes to apply patches
+        os.vrunv(gclient, {"sync", "--reset", "--delete_unversioned_trees", "-v"}, {envs = envs})
+
+        -- Fix GN issue
+        local patchPath = path.join(package:scriptdir(), "patches", package:version() .. ".patch")
```
waruqi, line 56: please add a sub-directory for version. `patches/x.x.x/xxx.patch`

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,71 @@ package("v8")
...
+
+        if os.exists(patchPath) then
+            local sourcePath = path.join(package:cachedir(), "source", "v8")
```
waruqi, line 59: use package:sourcedir()

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,71 @@ package("v8")
...
+            local cwd = os.cd(sourcePath)
+
+            os.execv("git", {"apply", patchPath})
```
waruqi, line 62: please use lower case. and we need not switch directory.

```lua
os.execv("git", {"apply", patch_path}, {curdir = sourcedir})
```

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,71 @@ package("v8")
...
         if package:is_plat("windows") then
-            envs.DEPOT_TOOLS_WIN_TOOLCAHIN = "0"
+            envs.DEPOT_TOOLS_WIN_TOOLCHAIN = "0"
+            envs.GYP_MSVS_VERSION = "2022"
         end
         local gclient = is_host("windows") and "gclient.bat" or "gclient"
-        os.vrunv(gclient, {"sync", "-v"}, {envs = envs})
+
+        -- Prevent long path issue on Windows
+        os.vrun("git config --global core.longpaths true")
```
waruqi, line 49: please do not modify user global configuration.

packages/v/v8/xmake.lua
poirierlouis: I tried `package:sourcedir()` (actually writing the output in a test.log file to see the path). It returns `nil`. Also I couldn't find a definition for `package:sourcedir()` in documentation, hence why I used `cachedir()`.
What am I missing?

PR #6863, opened 2025-04-09
v8: fix and update to 13.5.212.10

poirierlouis:
**Requires:**
xmake v2.9.9 to use `target` option when building with GN tool.

**Features:**
- support latest stable version 13.5.212.10
- tested on Windows only
- fix syntax to configure "runtimes", only MT for now
- only build static monolith flavor
- patch to fix build configuration of a dependency (third_party/partition_alloc/)
- erase previous version 10.0.58

**Notes:**
- `add_patches` is not used, because afaik, it is not meant to run in this context. `gclient` is ran during `on_install` callback. `add_patches` would try to apply the patch before execution of `gclient`, that is before files to be patched are even cloned/pulled.

- see previous https://github.com/xmake-io/xmake-repo/pull/6854

**Questions:**
(1) Is testing on Linux + macOS required to merge?
(2) Is backward compatibility with version 10.0.58 required to merge?


poirierlouis:
@waruqi sorry for creating a new PR. Previous one was broken, GitHub couldn't process my last commits somehow.

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,79 @@ package("v8")
...
+        -- Update repository and dependencies
+        -- Clean any local changes to apply patches
+        try {
+            function()
+                os.vrunv(gclient, {"sync", "-v", "--reset", "--delete_unversioned_trees"}, {envs = envs})
+            end,
+
+            finally {
+                function()
+                    os.vrunv("git", {"config", "--local", "--unset", "core.longpaths"})
```
waruqi, line 61: we need not unset local configuration, It will not affect other packages.

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,79 @@ package("v8")
...
+                end
+            }
+        }
+
+        -- Fix GN issue
+        local version = package:version()
```
waruqi, line 67: use package:version_str()

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,79 @@ package("v8")
...
+        end
+
+        -- Setup args.gn
         local configs = {
             is_official_build = false,
             is_component_build = false,
             is_debug = package:debug(),
-            is_shared_library = package:config("shared"),
-            symbol_level = 0,
+            symbol_level = package:debug() and 2 or 0,
```
waruqi, line 84: use package:is_debug()

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,79 @@ package("v8")
...
+                function()
+                    os.vrunv("git", {"config", "--local", "--unset", "core.longpaths"})
+                end
+            }
+        }
+
+        -- Fix GN issue
+        local version = package:version()
+        local patch_path = path.join(package:scriptdir(), "patches", version, "fix.patch")
+
```
waruqi, line 69: remove empty line

packages/v/v8/xmake.lua
```diff
@@ -49,36 +38,67 @@ package("v8")
...
+        -- Fix GN issue
+        local version = package:version_str()
+        local patch_path = path.join(package:scriptdir(), "patches", version, "fix.patch")
+        if os.exists(patch_path) then
+            local source_path = path.join(package:cachedir(), "source", "v8")
+            local fix_path = path.join(source_path, "third_party", "partition_alloc")
+            os.execv("git", {"apply", patch_path}, {curdir = source_path})
+            os.execv("git", {"add", path.join("src", "partition_alloc", "BUILD.gn")}, {curdir = fix_path})
+            os.execv("git", {"commit", "-m", "xmake-patch-01"}, {curdir = fix_path})
```
waruqi, line 64: I think we need not to and and commit it.

poirierlouis: and this is why `gclient sync` use reset flags to make sure it starts with a clean worktree to apply the patch.

poirierlouis:
[Linux (ubuntu-latest, static, debug)](https://github.com/xmake-io/xmake-repo/actions/runs/14385662124/job/40365152881?pr=6863#step:5:2810):
```
ar: obj/libv8_monolith.a: error reading obj/v8_base_without_compiler/code.o: No space left on device
ninja: build stopped: subcommand failed.
```

Looks like CI is not scaled for V8: it does take time/CPU/RAM/SSD. I haven't looked at all logs but I bet most of them failed due to resource exhaustion.

[Windows (MSVC 2022, static, x64, MT)](https://github.com/xmake-io/xmake-repo/actions/runs/14385662123/job/40365153050?pr=6863):
```
2025-04-11T07:00:45.0227329Z error: patch failed: third_party/partition_alloc/src/partition_alloc/BUILD.gn:41
2025-04-11T07:00:45.0228022Z error: third_party/partition_alloc/src/partition_alloc/BUILD.gn: patch does not apply
2025-04-11T07:00:45.0257025Z [0m[2;31;1merror: [0m@programdir\core\sandbox\modules\os.lua:378: execv(git apply D:\a\xmake-repo\xmake-repo\packages\v\v8\patches\13.5.212.10\fix.patch) failed(1)
```
Builds passed for versions 13.3.415.19 and 13.4.114.19.
It fails for 13.5.212.10 when trying to apply the patch. Which I'm not sure why yet. I'll take a look.

Is it required to pass all CI tests?
Can it be reduced to only `static` / `release` / `MT` flags? And later increase compatibility range...

waruqi:
> Looks like CI is not scaled for V8: it does take time/CPU/RAM/SSD. I haven't looked at all logs but I bet most of them failed due to resource exhaustion.

If we cannot confirm that the package works properly due to CI resource limits.

One solution is to improve CI and try to increase the resource limits. If that is not possible, at least the contributors need to confirm that the tests pass on their local machines.

poirierlouis:
Is it possible to add some kind of CI flags in xmake scripts/test.lua?
Regarding disk space on Windows, max is set to 32GB. It currently builds 2 versions, but the last one don't have enough space.
Instead of increasing max size in CI, what about changing test.lua: such as when multiple versions are tested, it will clean cache folder one after the other (free storage space)?

e.g.
- test v1 => fail or pass => clean cache
- test v2 => fail or pass => clean cache
- ...

This way no need to increase storage space in CI, and it would be better in the long term anyway. The amount of versions will eventually be to high to request an even more growing storage space on CI side.



I'll take a look to update package so it supports only Linux x86_64 and Windows x64 with static/MT/release. I can build both environments locally though.

waruqi:
The default is to install them in parallel. This can greatly improve the efficiency of test installation for most small libraries. However, we cannot distinguish some large libraries and implement serial compilation. I think you can solve this problem by submitting one version at a time instead of modifying multiple versions at once.

waruqi:
I removed other two versions, if it works on ci, you can open two new prs to add left two versions.

poirierlouis:
Thanks, I was going to remove them eventually. Better start with the latest version for now anyway.

I fixed two issues this time:
- Git account required (email + name) to build on Linux
- patch is obsolete (is not applied), build issue seems to have been fixed by Google

poirierlouis:
> you can add on_check to disable it.

Thanks, I added some checks and restrictions (Linux / Windows / static / release).

> You can increase it in this pr if it can.

I'm using an action to free disk space (common technique for free/public plans). We'll see if it is enough, and hopefully don't break things.

packages/v/v8/xmake.lua
```diff
@@ -17,7 +17,41 @@ package("v8")
...
+                 using value_type = int;
+                 int operator*() const;
+                 SimpleInputIterator& operator++();
+                 void operator++(int) { ++*this; }
+             };
+             static_assert(std::input_iterator<SimpleInputIterator>);
+         ]]}, {configs = {languages = "c++20"}}), "package(v8): require at least C++20.")
+
+        -- Only configured and tested for:
+        assert(is_mode("release") and is_kind("static"), "package(v8): only configured for static + release usage")
```
waruqi, line 39: no these apis, use `not package:is_debug()` and `not package:config("shared")`

packages/v/v8/xmake.lua
```diff
@@ -17,7 +17,41 @@ package("v8")
...
+
+        if is_host("windows") then
+            -- Require MSVC / Visual Studio 2022
+            local msvc = toolchain.load("msvc", {plat = package:plat(), arch = package:arch()})
```
waruqi, line 43: package:toolchain("msvc")

poirierlouis:
I'll see if I can fix builds for Archlinux, Fedora and Linux arm64 (locally first).
I will also add a check regarding runtime, it only supports MT for now.

poirierlouis:
Attempt to fix Archlinux and Fedora.
It should also ignore compilation for MD runtime.

I'm wondering if the added action for ubuntu workflow can cause issues for other packages?

poirierlouis:
Sorry check for MT was too global instead of scoped for Windows.

packages/v/v8/xmake.lua
```diff
@@ -36,10 +36,11 @@ package("v8")
          ]]}, {configs = {languages = "c++20"}}), "package(v8): require at least C++20.")
 
         -- Only configured and tested for:
-        assert(not package:is_debug() and not package:config("shared") and package:config("runtimes") == "MT",
-            "package(v8): only configured for static + release + MT usage")
+        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage")
 
         if is_host("windows") then
+            assert(package:config("runtimes") == "MT", "package(v8): only configured for MT on Windows")
```
poirierlouis, line 42: However I saw actions running for MD, `package:config("runtimes") == "MT"` is not the way to check for this?

packages/v/v8/xmake.lua
```diff
@@ -1,36 +1,59 @@
...
-              "v8_base_without_compiler",
-              "v8_libplatform",
-              "v8_libbase",
-              "torque_base",
-              "torque_generated_definitions",
-              "cppgc_base",
-              "torque_ls_base")
-
-    on_install("linux", "macosx", "windows", function (package)
+    add_includedirs("include", {public = true})
```
waruqi, line 17: we need not to set `{public = true}`

packages/v/v8/xmake.lua
waruqi: use `package:has_runtime("MT") or package:has_runtime("MTd")`

packages/v/v8/xmake.lua
```diff
@@ -1,36 +1,59 @@
...
+    add_links("v8_monolith")
+
+    on_check(function (package)
+        import("core.tool.toolchain")
+        import("core.base.semver")
```
waruqi, line 22: unused imports

packages/v/v8/xmake.lua
poirierlouis: thanks, `MTd` is not used because configuration only support release mode for now.

packages/v/v8/xmake.lua
poirierlouis: done

packages/v/v8/xmake.lua
poirierlouis: right

packages/v/v8/xmake.lua
```diff
@@ -1,36 +1,56 @@
...
+                 void operator++(int) { ++*this; }
+             };
+             static_assert(std::input_iterator<SimpleInputIterator>);
+         ]]}, {configs = {languages = "c++20"}}), "package(v8): require at least C++20.")
+
+        -- Only configured and tested for:
+        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage")
+
+        if is_host("windows") then
+            assert(package:has_runtime("MT"), "package(v8): only configured for MT on Windows")
```
waruqi, line 39: we can use add_configs + readonly to set them as MT, you can see other packages

poirierlouis: It's already done that way for Windows.
iirc, it was still running builds for MD in CI

waruqi: If we use readonly to lock it, MT will be automatically used even if the user configures MD or the default configuration.

https://github.com/xmake-io/xmake-repo/blob/b570f5befa9d0222fa19024ebad84c0a51a7802c/packages/a/aqt/xmake.lua#L10

poirierlouis: Such as `has_runtime` check is obsolete/ignored (even during CI)?

waruqi: > Such as `has_runtime` check is obsolete/ignored (even during CI)?

if you add readonly runtimes, we need not to add check, you can remove it

poirierlouis:
Alright, so I removed the redundant check, only readonly runtimes is used.
I also reverted changes regarding Archlinux / Fedora, I couldn't fix these builds. So not supporting them for now.

Afaik, it is fine now. Let me know if some things needs to be changed before merging.

waruqi:
> Alright, so I removed the redundant check, only readonly runtimes is used. I also reverted changes regarding Archlinux / Fedora, I couldn't fix these builds. So not supporting them for now.
> 
> Afaik, it is fine now. Let me know if something needs to be changed before merging.

nothing, but please fix left ci errors.

packages/v/v8/xmake.lua
```diff
@@ -45,6 +47,10 @@ package("v8")
             else
                 assert(false, "package(v8): only configured for MSVC on Windows.")
             end
+
+            assert(not (winos.version == 11 and is_arch("arm64")), "package(v8): Windows 11 arm64 is not supported.")
```
waruqi, line 51: winos.version() and it return semver instance, not number

waruqi: is_arch -> package:is_arch

packages/v/v8/xmake.lua
```diff
@@ -33,7 +33,9 @@ package("v8")
          ]]}, {configs = {languages = "c++20"}}), "package(v8): require at least C++20.")
 
         -- Only configured and tested for:
-        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage")
+        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage.")
+        assert(not is_host("arch"), "package(v8): Archlinux is not supported.")
+        assert(not is_host("fedora"), "package(v8): Fedora is not supported.")
```
waruqi, line 38: and here

packages/v/v8/xmake.lua
```diff
@@ -33,7 +33,9 @@ package("v8")
          ]]}, {configs = {languages = "c++20"}}), "package(v8): require at least C++20.")
 
         -- Only configured and tested for:
-        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage")
+        assert(not package:is_debug() and not package:config("shared"), "package(v8): only configured for static + release usage.")
+        assert(not is_host("arch"), "package(v8): Archlinux is not supported.")
```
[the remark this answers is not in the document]
poirierlouis: I'll change to use `package:is_plat`.
But I don't understand about "only linux". How can I deduce whether it is Archlinux / Fedora to only disable them, to fix CI errors?

packages/v/v8/xmake.lua
waruqi: use version():ge

```
PS C:\Users\Downloads\test> xmake l winos.version
10.0.19044+1288
PS C:\Users\Downloads\test> xmake l
xmake> winos.version():ge("10.0.10000+126")
< true
```

waruqi: > satisfies should return true, as my win version is above 10.0.22000+194, no?

remove space `>=1.0`

packages/v/v8/xmake.lua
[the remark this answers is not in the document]
waruqi: use linuxos.name

https://github.com/xmake-io/xmake/blob/386725082bbea11b0fb3894c7b3ae7411b375ae0/xmake/core/base/linuxos.lua#L80

is_host("linux") + linuxos.name() 

BTW, Why are libraries still bundled with distributions? We should fix them.

packages/v/v8/xmake.lua
poirierlouis: Thanks, [documentation](https://xmake.io/#/manual/extension_modules?id=semversatisfies) mislead me.

packages/v/v8/xmake.lua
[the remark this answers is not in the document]
poirierlouis: Alright, I changed it.

> Why are libraries still bundled with distributions? We should fix them.

What do you mean? That building V8 for Archlinux / Fedora should be supported and that it should work?

PR #6868, opened 2025-04-10
thrift: fix deps

star-hengxing:


packages/t/thrift/xmake.lua
```diff
@@ -19,12 +19,8 @@ package("thrift")
...
     add_configs("compiler", {description = "Build compiler", default = false, type = "boolean"})
 
-    add_deps("cmake", "boost")
-    if is_plat("windows") then
-        add_deps("winflexbison")
-    else
-        add_deps("flex", "bison")
-    end
+    add_deps("cmake", "flex", "bison", {kind = "binary"})
+    add_deps("boost", {configs = {locale = true}})
```
zhuoqiang, line 23: seems icu is required as well:

```lua
add_deps("boost", {configs = {locale = true, icu = true}})
```

PR #6885, opened 2025-04-13
Change the source of llvm 18.1.1 on windows

neroll-2021:
llvm 18.1.1 on windows cannot be used as library because it lacks a lot of files.

Change its source to `llvm/llvm-project/releases/download/llvmorg-18.1.1/clang+llvm-18.1.1-x86_64-pc-windows-msvc.tar.xz` to get full header files and libraries.


neroll-2021:
[xmake-mirror/llvm-windows/releases](https://github.com/xmake-mirror/llvm-windows/releases) is not updated, so tests failed because of checksum.

neroll-2021:
It seems that the checksum of llvm 18.1.1 from [xmake-mirror/llvm-windows/releases](https://github.com/xmake-mirror/llvm-windows/releases) is different from the one I have calculated before. I have made an update.

PR #6897, opened 2025-04-14
Slang rewrite

exdal:
My alternative approach to properly adding slang-shader.
I've partially rewritten the build script and removed CMake dependencies.
Current version directly pulls dependencies from xmake-repo instead.
Previously implemented on https://github.com/xmake-io/xmake-repo/pull/4138 https://github.com/xmake-io/xmake-repo/pull/4416.

This currently only supports slang -> spir-v output (with slang-glslang optimizer).

I've renamed it to `shader-slang` because of a name collusion with another library named `slang` (see: https://github.com/shader-slang/slang/issues/4016).

Any help on improving this package is appreciated.

### DIFF
--- a/packages/s/shader-slang/patches/v2025.6.3.patch
+++ b/packages/s/shader-slang/patches/v2025.6.3.patch
@@ -1,0 +1,13 @@
+diff --git a/source/slang-glslang/slang-glslang.cpp b/source/slang-glslang/slang-glslang.cpp
+index 8d9f64fd5..011dbf256 100644
+--- a/source/slang-glslang/slang-glslang.cpp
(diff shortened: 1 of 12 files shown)


exdal:
`slang-glslang` requires `MachineIndependent/localintermediate` header to be accessible, but xrepo's glslang doesn't seem to allow it. CMake version works properly.

waruqi:
> `slang-glslang` requires `MachineIndependent/localintermediate` header to be accessible, but xrepo's glslang doesn't seem to allow it. CMake version works properly.

we can improve glslang package, use os.cp to copy install them, or use add_patches to patch cmakelists.

PR #6933, opened 2025-04-18
freerdp: add package

star-hengxing:


luadebug:
Is it export surface (__dllimport__/__dllexport__) issue?

star-hengxing:
> Is it export surface (**dllimport**/**dllexport**) issue? It feels like `static` symbols broke for `.lib` on MinGW.

I've also find a similar issue on MinGW: https://github.com/jbaldwin/libcoro/issues/283
We need to investigate the different effects of these function modifiers:

- `__attribute__((visibility("default")))`
- `__declspec((dllexport/dllimport))`
- `__attribute__((dllexport/dllimport))`

luadebug:
> > Is it export surface (**dllimport**/**dllexport**) issue? It feels like `static` symbols broke for `.lib` on MinGW.
> 
> I've also find a similar issue on MinGW: [jbaldwin/libcoro#283](https://github.com/jbaldwin/libcoro/issues/283) We need to investigate the different effects of these function decoration:
> 
>     * `__attribute__((visibility("default")))`
> 
>     * `__declspec((dllexport/dllimport))`
> 
>     * `__attribute__((dllexport/dllimport))`
> 
> 
> https://github.com/FreeRDP/FreeRDP/blob/e2e6e9fd8581b779d2accba4e054fa652a13df9e/include/freerdp/api.h
> 
> ```c
> #if defined _WIN32 || defined __CYGWIN__
>     #ifdef FREERDP_EXPORTS
>         #ifdef __GNUC__
>             #define FREERDP_API __attribute__((dllexport))
>         #else
>             #define FREERDP_API __declspec(dllexport)
>         #endif
>     #else
>         #ifdef __GNUC__
>             #define FREERDP_API __attribute__((dllimport))
>         #else
>             #define FREERDP_API __declspec(dllimport)
>         #endif
>     #endif
> #else
>     #if defined(__GNUC__) && (__GNUC__ >= 4)
>         #define FREERDP_API __attribute__((visibility("default")))
>     #else
>         #define FREERDP_API
>     #endif
> #endif
> ```

I tried to rely over existing implementation, it required to add two more syslinks. It seems passing.

PR #6948, opened 2025-04-20
libllvm: add package.

Redbeanw44602:
### Differs from `xmake-repo/llvm`
This package is provided as a *library*, not a *toolchain*. (Even if its kind is set to *library*, it still has problems in use.)

### Packaging principles
Keep llvm-project as "default" as possible, that is, keep xrepo as little as possible without making choices for the user.

### Should splitting packages?
I discussed splitting the package with @star-hengxing in the telegram group, but I later thought that LLVM was more like a whole, so I didn't split it. If there are more opinions, we can discuss it.

### DIFF
--- a/packages/l/libllvm/xmake.lua
+++ b/packages/l/libllvm/xmake.lua
@@ -1,0 +1,222 @@
+package("libllvm")
+    set_kind("library")
+    set_homepage("https://llvm.org/")
+    set_description("The LLVM Compiler Infrastructure.")
+
+    -- The LLVM shared library cannot be built under windows.
+    add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = is_plat("windows")})
+
+    add_configs("exception", {description = "Enable C++ exception support for LLVM.", default = true, type = "boolean"})
+    add_configs("rtti",      {description = "Enable C++ RTTI support for LLVM.", default = true, type = "boolean"})
+
+    add_configs("ms_dia",  {description = "Enable DIA SDK to support non-native PDB parsing. (msvc only)", default = true, type = "boolean"})
+    add_configs("libffi",  {description = "Enable libffi to support the LLVM interpreter to call external functions.", default = false, type = "boolean"})
+    add_configs("httplib", {description = "Enable cpp-httplib to support llvm-debuginfod serve debug information over HTTP.", default = false, type = "boolean"})
+    add_configs("libcxx",  {description = "Use libc++ as C++ standard library instead of libstdc++, ", default = false, type = "boolean"})
+
+    includes(path.join(os.scriptdir(), "constants.lua"))
+    for _, project in ipairs(get_llvm_known_projects()) do
+        add_configs(project:gsub("-", "_"), {description = "Build " .. project .. " project.", default = (project == "clang"), type = "boolean"})
+    end
+    for _, runtime in ipairs(get_llvm_all_runtimes()) do
+        add_configs(runtime:gsub("-", "_"), {description = "Build " .. runtime .. " runtime.", default = false, type = "boolean"})
+    end
+
+    if is_plat("windows") then
+        -- pre-built
+        if is_arch("x64") then
+            add_urls("https://github.com/xmake-mirror/llvm-windows/releases/download/$(version)/clang+llvm-$(version)-win64.zip")
+            add_versions("19.1.7", "c6e058c6012f499811caa1ec037cc1b5c2fd2f8c20cc3315cae602cbd6c81a5e")
+        end
+
+        add_configs("runtimes",   {description = "Set vs compiler runtime.", default = "MT", readonly = true})
+        add_configs("vs_runtime", {description = "Set vs compiler runtime.", default = "MT", readonly = true})
+        add_syslinks("psapi", "shell32", "ole32", "uuid", "advapi32", "ws2_32", "ntdll", "version")
+    else
+        -- self-built
+        add_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version)/llvm-project-$(version).src.tar.xz")
+        add_versions("19.1.7", "82401fea7b79d0078043f7598b835284d6650a75b93e64b6f761ea7b63097501")
+
+        add_deps("ninja")
+        add_deps("zlib", "zstd", {optional = true})
+        set_policy("package.cmake_generator.ninja", true)
+    end
+
+    -- workaround to fix "error: undefined symbol: __mulodi4" (armeabi-v7a, r22, windows)
+    if is_plat("android") then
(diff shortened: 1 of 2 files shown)


Redbeanw44602:
### Possible test failures
I have run the same tests in my own repo and based on my experience, the following test will not pass, but it actually works.
 - `android` (linux, ndk_r22) 4 tests failed. This version of ndk creates huge library files that fill up the disk when building, causing failures.
 - `freebsd` The compilation cannot be completed. In fact, the memory size given by xrepo to the freebsd vm is unreasonable. Ubuntu runner provides 16g of memory while freebsd only has 4g. It will be killed due to oom during tblgen.

packages/l/libllvm/xmake.lua
```diff
@@ -0,0 +1,222 @@
...
+        end
+        for _, runtime in ipairs(constants.get_llvm_all_runtimes()) do
+            if package:config(runtime:gsub("-", "_")) then
+                table.insert(runtimes_enabled, runtime)
+            end
+        end
+
+        local configs = {
+            -- common cmake
+            "-DCMAKE_BUILD_TYPE=Release",
```
star-hengxing, line 112: Why not `table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))`

packages/l/libllvm/xmake.lua
```diff
@@ -0,0 +1,222 @@
...
+            "-DLLVM_INCLUDE_EXAMPLES=OFF",
+            "-DLLVM_INCLUDE_TESTS=OFF",
+            "-DLLVM_INCLUDE_BENCHMARKS=OFF",
+            "-DLLVM_OPTIMIZED_TABLEGEN=ON",
+            "-DLLVM_ENABLE_PROJECTS=" .. table.concat(projects_enabled, ";"),
+            "-DLLVM_ENABLE_RUNTIMES=" .. table.concat(runtimes_enabled, ";"),
+
+            -- disable tools build - to save link time
+            "-DLLVM_BUILD_TOOLS=OFF",
+            "-DCLANG_BUILD_TOOLS=OFF",
```
star-hengxing, line 126: Even if this option is disabled, some tools will still be built. If lto is enabled, building these tools will waste a amount of time. You can reference this patch.
https://github.com/clice-project/llvm-binary/blob/84226882fc2a00f838856514f92f7e00585b4b6f/xmake.lua#L48-L69

packages/l/libllvm/xmake.lua
```diff
@@ -0,0 +1,222 @@
...
+            add_versions("19.1.7", "c6e058c6012f499811caa1ec037cc1b5c2fd2f8c20cc3315cae602cbd6c81a5e")
+        end
+
+        add_configs("runtimes",   {description = "Set vs compiler runtime.", default = "MT", readonly = true})
+        add_configs("vs_runtime", {description = "Set vs compiler runtime.", default = "MT", readonly = true})
+        add_syslinks("psapi", "shell32", "ole32", "uuid", "advapi32", "ws2_32", "ntdll", "version")
+    else
+        -- self-built
+        add_urls("https://github.com/llvm/llvm-project/releases/download/llvmorg-$(version)/llvm-project-$(version).src.tar.xz")
+        add_versions("19.1.7", "82401fea7b79d0078043f7598b835284d6650a75b93e64b6f761ea7b63097501")
```
star-hengxing, line 38: Please add git url.

packages/l/libllvm/xmake.lua
```diff
@@ -0,0 +1,222 @@
+package("libllvm")
+    set_kind("library")
+    set_homepage("https://llvm.org/")
+    set_description("The LLVM Compiler Infrastructure.")
+
+    -- The LLVM shared library cannot be built under windows.
+    add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = is_plat("windows")})
```
star-hengxing, line 7:
```lua
if is_plat("windows","mingw") then
end
```

packages/l/libllvm/xmake.lua
```diff
@@ -0,0 +1,222 @@
...
+
+    -- error: undefined symbol: backtrace
+    if is_plat("bsd") then
+        add_syslinks("execinfo")
+    end
+
+    add_deps("cmake")
+    on_load(function (package)
+        local constants = import('constants')
+        package:addenv("PATH", "bin")
```
star-hengxing, line 58: Why?

packages/l/libllvm/xmake.lua
Redbeanw44602: Is there a way to customize git tags? (similar to `version=...`).
For example, I want to add `llvmorg-` before the tag.🤔

packages/l/libllvm/xmake.lua
Redbeanw44602: Forgot why this line exists. 😵 Removed.

Redbeanw44602:
Can someone merge #7175 first? This saves one test.

Redbeanw44602:
Looks like https://github.com/xmake-io/xmake/commit/d98c1ab687d99d955866e3de56b78eaa5ab1af40  breaks the package, we need to undefine `BUILD_SHARED_LIBS`.

star-hengxing:
> Looks like [xmake-io/xmake@d98c1ab](https://github.com/xmake-io/xmake/commit/d98c1ab687d99d955866e3de56b78eaa5ab1af40) broke this package, we need to undefine `BUILD_SHARED_LIBS`.

```lua
import("package.tools.cmake")

local opt = {}
opt.envs = cmake.buildenvs(package)
opt.envs.CMAKE_BUILD_TYPE = nil
cmake.install(package, configs, opt)
```

Redbeanw44602:
Thanks for your reply. But this doesn't seem to work, except modifying the xmake source code, we can't undefine that variable.🙁

But... I found that simply keeping `BUILD_SHARED_LIBS=OFF` seemed to work as expected, so now I run a test.

Redbeanw44602:
I think the package is ready to merge, all the tests failing are with code 143.

https://github.com/actions/runner-images/issues/6680

luadebug:
[660/4518] This one is very huge?
I am sorry, that I am unsure that producing very long build package is okay without splitting it into smaller parts. :/

Redbeanw44602:
> [660/4518] This one is very huge?

When a job is interrupted by github, the log may not be dumped completely.

Try viewing the raw logs and you'll get:
```
Failed to generate URL to download logs.
```

So, it is hard to determine what caused the job to be interrupted. If you run it on local machine (same environment), it will succeed.

luadebug:
https://github.com/xmake-io/xmake-repo/pull/7286
Github CI Worker has been promoted from win2019 to win2025.

Issue #6898, opened 2025-04-14
glslang: cannot access `MachineIndependent`


exdal:
### Xmake Version

xmake v2.9.9+20250408,

### Operating System Version and Architecture

All

### Describe Bug

glslang package doesn't install `MachineIndependent` directory required by `slang-shader` package (see: https://github.com/xmake-io/xmake-repo/pull/6897#issuecomment-2801374258)

### Expected Behavior

```cpp
#include "glslang/MachineIndependent/localintermediate.h"  // should be accessible.
```

### Project Configuration

```lua
add_requires("glslang 1.4.309+0")
target("slang-glslang")
    add_packages("glslang")
```

### Additional Information and Error Logs

```console
error: @programdir/core/main.lua:329: @programdir/actions/build/main.lua:146: @programdir/modules/async/runjobs.lua:331: @programdir/modules/private/action/build/object.lua:100: @programdir/modules/core/tools/gcc.lua:997: source/slang-glslang/slang-glslang.cpp:5:10: fatal error: glslang/MachineIndependent/localintermediate.h: No such file or directory
    5 | #include "glslang/MachineIndependent/localintermediate.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

PR #6982, opened 2025-04-24
glslang: expose MachineIndependent headers

exdal:
fixes and closes https://github.com/xmake-io/xmake-repo/issues/6898

### DIFF
--- a/packages/g/glslang/xmake.lua
+++ b/packages/g/glslang/xmake.lua
@@ -91,6 +91,9 @@ package("glslang")
             package:add("links", "glslang", "glslang-default-resource-limits")
         end
 
+        os.cp("glslang/MachineIndependent/**.h", path.join(package:installdir("include"), "glslang", "MachineIndependent"))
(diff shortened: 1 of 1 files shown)


packages/g/glslang/xmake.lua
```diff
@@ -91,6 +91,9 @@ package("glslang")
             package:add("links", "glslang", "glslang-default-resource-limits")
         end
 
+        os.cp("glslang/MachineIndependent/**.h", path.join(package:installdir("include"), "glslang", "MachineIndependent"))
+        os.cp("glslang/Include/**.h", path.join(package:installdir("include"), "glslang", "Include"))
```
SirLynix, line 95:
```suggestion
        os.mv("glslang/MachineIndependent", path.join(package:installdir("include"), "glslang"))
        os.mv("glslang/Include", path.join(package:installdir("include"), "glslang"))
```
wouldn't moving the folder just work?

PR #6989, opened 2025-04-25
update cpuinfo

apocelipes:
- replace `package:config("vs_runtime")` with `package:runtimes()`
- support more arm cpus

### DIFF
--- a/packages/c/cpuinfo/xmake.lua
+++ b/packages/c/cpuinfo/xmake.lua
@@ -58,13 +58,19 @@ package("cpuinfo")
         end
 
         if package:is_plat("windows") then
-            table.insert(configs, "-DCPUINFO_RUNTIME_TYPE=" .. (package:runtimes():startswith("MT") and "static" or "shared"))
+            table.insert(configs, "-DCPUINFO_RUNTIME_TYPE=default")
+            if package:is_arch("arm64") then
+                table.insert(configs, "-DCPUINFO_TARGET_PROCESSOR=arm64")
+            end
             local vs_sdkver = import("core.tool.toolchain").load("msvc"):config("vs_sdkver")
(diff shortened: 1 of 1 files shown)


packages/c/cpuinfo/xmake.lua
```diff
@@ -57,7 +58,7 @@ package("cpuinfo")
         end
 
         if package:is_plat("windows") then
-            table.insert(configs, "-DCPUINFO_RUNTIME_TYPE=" .. (package:config("vs_runtime"):startswith("MT") and "static" or "shared"))
+            table.insert(configs, "-DCPUINFO_RUNTIME_TYPE=" .. (package:runtimes():startswith("MT") and "static" or "shared"))
```
star-hengxing, line 61: I think we can use `-DCPUINFO_RUNTIME_TYPE=default`
```cmake
MACRO(CPUINFO_TARGET_RUNTIME_LIBRARY target)
  IF(MSVC AND NOT CPUINFO_RUNTIME_TYPE STREQUAL "default")
    IF(CPUINFO_RUNTIME_TYPE STREQUAL "shared")
      TARGET_COMPILE_OPTIONS(${target} PRIVATE
        "/MD$<$<CONFIG:Debug>:d>")
    ELSEIF(CPUINFO_RUNTIME_TYPE STREQUAL "static")
      TARGET_COMPILE_OPTIONS(${target} PRIVATE
        "/MT$<$<CONFIG:Debug>:d>")
    ENDIF()
  ENDIF()
ENDMACRO()
```

apocelipes: いいですね！アドバイスありがとうございます。

apocelipes:
Windows arm ci always tries to compile x86 code
image

PR #1131, opened 2022-04-02
add exosip, osip, jrtplib, jthread, resip

hsinor:
* add osip
* add exosip
* add jthread
* add jrtplib
* add resip

PR #7002, opened 2025-04-26
resiprocate: add package

luadebug:
https://cgit.freebsd.org/src/plain/docs/UsingLibcxx.rst?h=vendor/libc%2b%2b/libc%2b%2b-release_701-r349250&id=74c4bc8d0eddcb4786594f1c6b598094fac43859
Resolves: https://github.com/xmake-io/xmake-repo/pull/1131

### DIFF
--- a/packages/r/resip/xmake.lua
+++ b/packages/r/resip/xmake.lua
@@ -1,5 +1,4 @@
 package("resip")
-
     set_homepage("https://resiprocate.org/Main_Page")
     set_description("C++ implementation of SIP, ICE, TURN and related protocols.")
 
@@ -13,7 +12,12 @@ package("resip")
         add_links("resip", "dum", "rutil", "resipares")
         add_deps("autotools")
         add_deps("openssl", "c-ares")
+    end
+
+    if is_plat("linux", "bsd") then
         add_syslinks("pthread")
+    elseif is_plat("macosx", "iphoneos") then
+        add_deps("pkg-config")
     end
 
     on_load("windows", function(package)
@@ -58,7 +62,7 @@ package("resip")
         os.cp("*/*/rutil.lib", package:installdir("lib"))
     end)
 
-    on_install("!windows", function(package)
+    on_install("!windows and !wasm", function(package)
         local configs = {}
         table.insert(configs, "--enable-shared=" .. (package:config("shared") and "yes" or "no"))
         table.insert(configs, "--enable-static=" .. (package:config("shared") and "no" or "yes"))
--- a/packages/r/resip/xmake.lua
+++ b/packages/r/resip/xmake.lua
@@ -78,7 +78,7 @@ package("resip")
             table.insert(configs, "--enable-debug")
         end
         if package:is_plat("bsd", "android") then
-            opt.cxxflags = "-std=c++11"
+            opt.cxflags = "-D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR"
         end
         import("package.tools.autoconf").install(package, configs)
     end)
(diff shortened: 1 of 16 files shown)


star-hengxing:
> For Windows OS in order to build `3-4` shared libs it would require to create 3-4 `.DEF` files & store them into xrepo?

In software engineering, trade-offs are essential, and maintaining multiple `.def` files for each version is impractical.

luadebug:
https://github.com/conan-io/conan-center-index/blob/c29c3a06d0c5d4fd98529a34586c4f60ab00f659/recipes/resiprocate/all/conanfile.py#L15-L21
Name package as `resiprocate`?

star-hengxing:
> https://github.com/conan-io/conan-center-index/blob/c29c3a06d0c5d4fd98529a34586c4f60ab00f659/recipes/resiprocate/all/conanfile.py#L15-L21 Name package as `resiprocate`?

Sure.

---

https://github.com/resiprocate/resiprocate/blob/ceda26fef4d5586f730527ed9c7546e44cab2acd/CMakeLists.txt

Why not use cmake for build?

packages/r/resip/xmake.lua
```diff
@@ -0,0 +1,94 @@
...
+        add_deps("autotools")
+        add_deps("openssl", "c-ares")
+    end
+
+    if is_plat("linux", "bsd") then
+        add_syslinks("pthread")
+    end
+    if is_plat("macosx", "iphoneos", "bsd") then
+        add_deps("pkg-config")
+    end
```
star-hengxing, line 22: I think this way of code is not very clear, I suggest
```lua
-- configs
if is_plat()

-- deps
if is_plat()

-- syslinks
if is_plat()
```

packages/r/resip/xmake.lua
```diff
@@ -0,0 +1,94 @@
+package("resip")
+    set_homepage("https://resiprocate.org/Main_Page")
+    set_description("C++ implementation of SIP, ICE, TURN and related protocols.")
+
```
star-hengxing, line 4: Missing license.
https://github.com/conan-io/conan-center-index/blob/c29c3a06d0c5d4fd98529a34586c4f60ab00f659/recipes/resiprocate/all/conanfile.py#L21

packages/r/resip/xmake.lua
```diff
@@ -0,0 +1,94 @@
...
+    add_urls("https://github.com/resiprocate/resiprocate/archive/refs/tags/resiprocate-$(version).tar.gz")
+    add_versions("1.12.0", "aa8906082e4221bffbfab3210df68a6ba1f57ba1532d89ea4572b4fa9877914f")
```
star-hengxing, line 6: Missing git url.

luadebug:
> https://github.com/resiprocate/resiprocate/blob/ceda26fef4d5586f730527ed9c7546e44cab2acd/CMakeLists.txt
> 
> Why not use cmake for build?
That is because `resiprocate-1.12.0` does not provide CMakeLists.txt

star-hengxing:
> > https://github.com/resiprocate/resiprocate/blob/ceda26fef4d5586f730527ed9c7546e44cab2acd/CMakeLists.txt
> > Why not use cmake for build?
> 
> That is because `resiprocate-1.12.0` (1.12.0) does not provide `CMakeLists.txt` file.

Why not use lastest version?

luadebug:
> > > https://github.com/resiprocate/resiprocate/blob/ceda26fef4d5586f730527ed9c7546e44cab2acd/CMakeLists.txt
> > > Why not use cmake for build?
> > 
> > 
> > That is because `resiprocate-1.12.0` (1.12.0) does not provide `CMakeLists.txt` file.
> 
> Why not use lastest version?

Because this tries to resolve https://github.com/xmake-io/xmake-repo/pull/1131 adds 1.10.2 & 1.12.0. (1.10.2 is too old).
IMHO, it is good idea to keep one old legacy version with Makefile build system, and I could not find newer than 1.12.0, as versioning being done like 1.12.0 -> next version 1.13.1.

PR #7023, opened 2025-04-29
adding wgsl-validator package (and rust cross-compilation support)

Kbz-8:


SirLynix:
Waiting for https://github.com/xmake-io/xmake/pull/6392 to be merged to fix the error on arm64/x86

Kbz-8:
the rust package don't seem to handle Android and iOS platforms ?

SirLynix:
Yes it seems some instructions are missing in the rust package for android/ios

Kbz-8:
even some Windows targets

SirLynix:
one of the big issue now is that xmake doesn't support cross-compilation when installing a cargo package, I think the same target should be passed as well but I don't have time to test it right now

SirLynix (changes requested):

SirLynix:
on_check doesn't seem to be called with the package configs.

It should be called twice, one when testing the package by itself (default config, which is done) and one when testing wgsl-validator with the right configs.

SirLynix:
@waruqi 
```
error: @programdir/core/sandbox/modules/os.lua:273: error: toolchain '1.86.0-x86_64-unknown-linux-gnu' does not support target 'arm-unknown-linux-gnu'
note: you can see a list of supported targets with `rustc --print=target-list`
```

rustc supports the following cross targets:
```
arm-unknown-linux-gnueabi
arm-unknown-linux-gnueabihf
arm-unknown-linux-musleabi
arm-unknown-linux-musleabihf
```
but I don't know how can I extract the last part (e.g. gnueabihf) with xmake to install the correct toolchain target?

SirLynix:
It should be good, I moved a lot of code to a separate rustlib package to serve as a template for future Rust libraries. I'm also pointing on dev branch for cross as it's required so xmake doesn't pass the --target in the toolchain.

Other than the cross.yml change it can be merged.

SirLynix (approved):

packages/r/rustlib/xmake.lua
```diff
@@ -0,0 +1,72 @@
...
+            rcfile:print("add_requireconfs(\"cargo::naga\", {arch = \"%s\", override = true})", cross_target)
+            rcfile:print("add_rcflags(\"--target=%s\")", cross_target)
+        end
+        rcfile:close()
+
+        local envs = import("package.tools.xmake").buildenvs(package)
+        table.insert(envs.XMAKE_RCFILES, rcfile_path)
+
+        return envs
+    end)
```
waruqi, line 72: no on_test

SirLynix: It's a template, it's meant to simplify packages using it as a base

packages/r/rustlib/xmake.lua
```diff
@@ -0,0 +1,72 @@
...
+        if cross_target then
+            rcfile:print("add_requireconfs(\"cargo::naga\", {arch = \"%s\", override = true})", cross_target)
+            rcfile:print("add_rcflags(\"--target=%s\")", cross_target)
+        end
+        rcfile:close()
+
+        local envs = import("package.tools.xmake").buildenvs(package)
+        table.insert(envs.XMAKE_RCFILES, rcfile_path)
+
+        return envs
```
waruqi, line 71: ? why?

packages/r/rustlib/xmake.lua
```diff
@@ -0,0 +1,72 @@
...
+    on_load(function (package)
+        local toolchainconfigs = {}
+        toolchainconfigs.target_plat = package:plat()
+        toolchainconfigs.target_arch = package:arch()
+
+        if package:is_plat("cross") then
+            -- detect cross configuration from the compiler, if possible
+            local compiler, toolname = package:tool("cc")
+            if toolname == "clang" or toolname == "gcc" then
+                local outdata, errdata = os.iorunv(compiler, {"-v"})
```
waruqi, line 14: we should not call os.iorunv in on_load

SirLynix: Unfortunately that's the only way I found to extract the compiler target (which is a useful information for cross platform)

packages/r/rustlib/xmake.lua
SirLynix: So it can be passed to the child package, but maybe data_set is better here

packages/r/rustlib/xmake.lua
SirLynix: I took a look at the way xmake handles tools, I was thinking of making a pull request to add the support of an `:info()` method, or maybe a complete `:target()` or `:target_abi()`, but I think this is something you probably want to think a bit before. Do you want me to open a feature request?

waruqi: You can submit a feature request first. I haven't figured out whether it is needed and how to design the API, so it may not be added for the time being.

PR #7043, opened 2025-04-30
tracy: add `xmake` buildsystem & use by default

luadebug:


packages/t/tracy/xmake.lua
```diff
@@ -12,80 +12,138 @@ package("tracy")
...
-        -- collect tracy defines from cmake configs
-        for _, config in ipairs(configs) do
-            local define, value = config:match("-D(TRACY_%S+)=(.*)")
-            if define and value and value == "ON" then
-                package:add("defines", define)
+            -- collect tracy defines from cmake configs
+            for _, config in ipairs(configs) do
+                local define, value = config:match("-D(TRACY_%S+)=(.*)")
+                if define and value and value == "ON" then
+                    package:add("defines", define)
```
star-hengxing, line 93: It will missing defines when use xmake to build.

luadebug: Indeed, this would require to define mappings I guess, in order to resolve this.

packages/t/tracy/xmake.lua
```diff
@@ -12,80 +12,190 @@ package("tracy")
...
-    on_install("windows|x64", "macosx", "linux|x86_64", function (package)
-        local configs = {}
-        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
-        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+    on_load(function (package)
+        if package:config("cmake") then
+            package:add("deps", "cmake")
+        end
+        if package:is_plat("windows", "mingw") then
+            package:add("defines", "_WIN32")
```
star-hengxing, line 57: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

PR #7065, opened 2025-05-04
add opencv android

kekopom:
- Add support of Android opencv in xrepo
- Disabling Android tests `-DBUILD_ANDROID_EXAMPLES=OFF` as it causes issues. see https://github.com/opencv/opencv/issues/15769#issuecomment-549570072
- Got issues from Android-Determine.cmake (cmake 3.28) : Android: Unknown processor CMAKE_SYSTEM_PROCESSOR=arm64-v8a, looking into the file it appears that we should map arm64-v8a to aarch64 by passing it to -DCMAKE_SYSTEM_PROCESSOR=aarch64
[Need to way a clean way to handle it] ==> done: setting the `DCMAKE_ANDROID_ARCH_ABI` instead of `DCMAKE_SYSTEM_PROCESSOR` for Android seems the right way to do it.

Testing locally by running `xmake l scripts/test.lua -v -D --shallow -p android --ndk=~/android-ndk-r28b --ndk_sdkver=32 -a arm64-v8a opencv`
and for shared:
`xmake l scripts/test.lua -v -D --shallow -p android --ndk=~/android-ndk-r27c --ndk_sdkver=32 -a arm64-v8a -k shared opencv`

### DIFF
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -241,11 +241,11 @@ package("opencv")
             package:add("links", reallink)
         end
         if package:is_plat("android") then
-            local linkdir = (package:config("shared") and "libs" or "staticlibs")
-            local libname = "lib*." .. (package:config("shared") and "so" or "a")
-            local libfiles = {}
-            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native", linkdir, package:targetarch(), libname)))
-            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native/3rdparty/libs", package:targetarch(), libname)))
+            for _, suffix in ipairs({"*.a", "*.so"}) do
+                for _, f in ipairs(os.files(path.join(package:installdir(path.join("sdk/native/3rdparty/libs", package:targetarch())), suffix))) do
+                    package:add("links", path.basename(f):match("lib(.+)"))
+                end
+            end
             for _, f in ipairs(libfiles) do
                 if not f:match("opencv_.+") then
                     package:add("links", path.basename(f):match("lib(.+)"))
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -179,7 +179,7 @@ package("opencv")
             table.insert(configs, "-DBUILD_WITH_STATIC_CRT=" .. (package:has_runtime("MT", "MTd") and "ON" or "OFF"))
             if package:is_arch("arm64") then
                 -- https://github.com/opencv/opencv/issues/25052
-                table.insert(configs, "-DCPU_BASELINE=NEON")
+                table.insert(configs, "-DCPU_NEON_FP16_SUPPORTED=OFF")
                 -- https://github.com/opencv/opencv/issues/24235
                 table.insert(configs, "-DOPENCV_SKIP_SYSTEM_PROCESSOR_DETECTION=ON")
                 table.insert(configs, "-DAARCH64=ON")
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -77,7 +77,7 @@ package("opencv")
         end
     end)
 
-    on_load("linux", "macosx", "windows", "mingw@windows,msys", function (package)
+    on_load("android", "linux", "macosx", "windows", "mingw@windows,msys", function (package)
         if package:is_plat("windows") then
             local arch = "x64"
             if     package:is_arch("x86")   then arch = "x86"
@@ -97,6 +97,10 @@ package("opencv")
             local arch = (package:is_arch("x86_64") and "x64" or "x86")
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             package:add("linkdirs", path.join(arch, "mingw", linkdir))
+        elseif package:is_plat("android") then
+            local arch = package:targetarch()
+            local linkdir = (package:config("shared") and "lib" or "staticlibs")
+            package:add("linkdirs", path.join("sdk/native", linkdir, package:targetarch()))
         elseif package:version():ge("4.0") then
             package:add("includedirs", "include/opencv4")
             package:add("linkdirs", "lib", "lib/opencv4/3rdparty")
@@ -136,7 +140,7 @@ package("opencv")
         end)
     end
 
-    on_install("linux", "macosx", "windows", "mingw@windows,msys", function (package)
+    on_install("android", "linux", "macosx", "windows", "mingw@windows,msys", function (package)
         io.replace("cmake/OpenCVUtils.cmake", "if(PKG_CONFIG_FOUND OR PkgConfig_FOUND)", "if(NOT WIN32 AND (PKG_CONFIG_FOUND OR PkgConfig_FOUND))", {plain = true})
         local configs = {"-DCMAKE_OSX_DEPLOYMENT_TARGET=",
                          "-DBUILD_PERF_TESTS=OFF",
@@ -179,8 +183,18 @@ package("opencv")
                 table.insert(configs, "-DCMAKE_SYSTEM_NAME=Darwin")
             elseif package:is_plat("linux") then
                 table.insert(configs, "-DCMAKE_SYSTEM_NAME=Linux")
+            elseif package:is_plat("android") then
+                table.insert(configs, "-DCMAKE_SYSTEM_NAME=Android")
+                -- from https://github.com/opencv/opencv/issues/15769#issuecomment-549570072
+                table.insert(configs, "-DBUILD_ANDROID_EXAMPLES=OFF")
             end
-            table.insert(configs, "-DCMAKE_SYSTEM_PROCESSOR=" .. package:targetarch())
+
+            -- In case of android we prefer to set CMAKE_ANDROID_ARCH_ABI rather than CMAKE_SYSTEM_PROCESSOR
+            if package:is_plat("android") then
+                table.insert(configs, "-DCMAKE_ANDROID_ARCH_ABI=" .. package:targetarch())
+            else
+                table.insert(configs, "-DCMAKE_SYSTEM_PROCESSOR=" .. package:targetarch())
+            end
         end
         local resourcedir = package:resourcedir("opencv_contrib")
         if resourcedir then
@@ -223,7 +237,18 @@ package("opencv")
             end
             package:add("links", reallink)
         end
-        if package:is_plat("windows") then
+        if package:is_plat("android") then
+            local linkdir = (package:config("shared") and "lib" or "staticlibs")
+            local libfiles = {}
+            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native", linkdir, package:targetarch(), "lib*.a")))
+            for _, f in ipairs(libfiles) do
+                if f:match("libopencv_.+") then
+                    package:add("links", path.basename(f))
+                end
+            end
+            package:addenv("PATH", "bin")
+
+        elseif package:is_plat("windows") then
             local arch = "x64"
             if     package:is_arch("x86")   then arch = "x86"
             elseif package:is_arch("arm64") then arch = "ARM64"
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -193,6 +193,7 @@ package("opencv")
                 table.insert(configs, "-DCMAKE_SYSTEM_NAME=Android")
                 -- from https://github.com/opencv/opencv/issues/15769#issuecomment-549570072
                 table.insert(configs, "-DBUILD_ANDROID_EXAMPLES=OFF")
+                table.insert(configs, "-DBUILD_ANDROID_PROJECTS=OFF")
             end
 
             -- In case of android we prefer to set CMAKE_ANDROID_ARCH_ABI rather than CMAKE_SYSTEM_PROCESSOR
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -85,12 +85,6 @@ package("opencv")
             if     package:is_arch("x86")   then arch = "x86"
             elseif package:is_arch("arm64") then arch = "ARM64"
             end
-            -- Workaround for arm64
-            if package:is_arch("arm64") and not os.isdir("ARM64") then
-                if     os.isdir("x64") then arch = "x64"
-                elseif os.isdir("x86") then arch = "x86"
-                end
-            end
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             local vs = package:toolchain("msvc"):config("vs")
             local vc_ver = "vc13"
@@ -259,6 +253,12 @@ package("opencv")
             local arch = "x64"
             if     package:is_arch("x86")   then arch = "x86"
             elseif package:is_arch("arm64") then arch = "ARM64"
+            end
+            -- Workaround for arm64
+            if package:is_arch("arm64") and not os.isdir("ARM64") then
+                if     os.isdir("x64") then arch = "x64"
+                elseif os.isdir("x86") then arch = "x86"
+                end
             end
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             local vs = package:toolchain("msvc"):config("vs")
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -179,7 +179,7 @@ package("opencv")
             table.insert(configs, "-DBUILD_WITH_STATIC_CRT=" .. (package:has_runtime("MT", "MTd") and "ON" or "OFF"))
             if package:is_arch("arm64") then
                 -- https://github.com/opencv/opencv/issues/25052
-                table.insert(configs, "-DCPU_NEON_FP16_SUPPORTED=OFF")
+                table.insert(configs, "-DCPU_BASELINE=NEON")
                 -- https://github.com/opencv/opencv/issues/24235
                 table.insert(configs, "-DOPENCV_SKIP_SYSTEM_PROCESSOR_DETECTION=ON")
                 table.insert(configs, "-DAARCH64=ON")
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -100,7 +100,6 @@ package("opencv")
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             package:add("linkdirs", path.join(arch, "mingw", linkdir))
         elseif package:is_plat("android") then
-            local arch = package:targetarch()
             local linkdir = (package:config("shared") and "libs" or "staticlibs")
             package:add("linkdirs", path.join("sdk/native", linkdir, package:targetarch()))
             package:add("linkdirs", path.join("sdk/native/3rdparty/libs", package:targetarch()))
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -251,7 +251,6 @@ package("opencv")
                     package:add("links", path.basename(f):match("lib(.+)"))
                 end
             end
-            package:addenv("PATH", "bin")
 
         elseif package:is_plat("windows") then
             local arch = "x64"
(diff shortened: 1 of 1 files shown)


kekopom:
current state: 

CMake Error at /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake:113 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

packages/o/opencv/xmake.lua
```diff
@@ -223,7 +237,18 @@ package("opencv")
...
+        if package:is_plat("android") then
+            local linkdir = (package:config("shared") and "lib" or "staticlibs")
+            local libfiles = {}
+            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native", linkdir, package:targetarch(), "lib*.a")))
+            for _, f in ipairs(libfiles) do
+                if f:match("libopencv_.+") then
+                    package:add("links", path.basename(f))
+                end
+            end
+            package:addenv("PATH", "bin")
```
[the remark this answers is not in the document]
luadebug: 244 -> 248 line https://xmake.io/#/manual/package_dependencies?id=packageadd_links
When you add links it should be one line `add_links("mbedtls", "mbedx509", "mbedcrypto")` which shows proper link sequence. Dependencies earlier.
Why would you do `package:addenv("PATH", "bin")`?

kekopom:
> > current state:
> > CMake Error at /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake:113 (cmake_minimum_required): Compatibility with CMake < 3.5 has been removed from CMake.
> > Update the VERSION argument value. Or, use the ... syntax to tell CMake that the project requires at least but has been updated to work with policies introduced by or earlier.
> > Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
> > ==> Looks like adding table.insert(configs, "-DCMAKE_POLICY_VERSION_MINIMUM=3.5") fixed the issue
> 
> If you update xmake by `xmake update dev`, would it resolve this issue, so there would not need to add `"-DCMAKE_POLICY_VERSION_MINIMUM=3.5"`?

Looks like it didn't work :(

FAILED: unix-install/opencv4.pc /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd/unix-install/opencv4.pc 
cd /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd && /home/kekopom/.xmake/packages/c/cmake/4.0.1/ee0342c3438b4e54a11f5d4d26f5299a/bin/cmake -DCMAKE_HELPER_SCRIPT=/home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd/OpenCVGenPkgConfig.info.cmake -P /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake
CMake Error at /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake:113 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

packages/o/opencv/xmake.lua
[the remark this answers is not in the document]
kekopom: In line 103 since we do `package:add("linkdirs", path.join("sdk/native", linkdir, package:targetarch()))`
I think we don't need to specify again each lib that is in path.join("sdk/native", linkdir, package:targetarch()) 
as we already include it in package:add_linkdirs right?

I might have misunderstood your comment if that's the case I'm sorry 🙏 thanks a lot for your reply though :)

luadebug:
> > > current state:
> > > CMake Error at /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake:113 (cmake_minimum_required): Compatibility with CMake < 3.5 has been removed from CMake.
> > > Update the VERSION argument value. Or, use the ... syntax to tell CMake that the project requires at least but has been updated to work with policies introduced by or earlier.
> > > Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
> > > ==> Looks like adding table.insert(configs, "-DCMAKE_POLICY_VERSION_MINIMUM=3.5") fixed the issue
> > 
> > 
> > If you update xmake by `xmake update dev`, would it resolve this issue, so there would not need to add `"-DCMAKE_POLICY_VERSION_MINIMUM=3.5"`?
> 
> Looks like it didn't work :( using xmake --version xmake v3.0.0+dev.3d3f6c3
> 
> FAILED: unix-install/opencv4.pc /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd/unix-install/opencv4.pc cd /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd && /home/kekopom/.xmake/packages/c/cmake/4.0.1/ee0342c3438b4e54a11f5d4d26f5299a/bin/cmake -DCMAKE_HELPER_SCRIPT=/home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/bd/OpenCVGenPkgConfig.info.cmake -P /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake CMake Error at /home/kekopom/.xmake/cache/packages/2505/o/opencv/4.11.0/source/cmake/OpenCVGenPkgconfig.cmake:113 (cmake_minimum_required): Compatibility with CMake < 3.5 has been removed from CMake.
> 
> Update the VERSION argument value. Or, use the ... syntax to tell CMake that the project requires at least but has been updated to work with policies introduced by or earlier.
> 
> Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
> 
> ==> Looks like the warning has became an error since CMake 4.0.0 which is pulled by XMake if I am not misleading

https://github.com/xmake-io/xmake/blob/ea13d87b271fbfe6ec96b90cb6b315220f4c53e9/xmake/modules/package/tools/cmake.lua#L951-L953 This supposed to pass this -DCMAKE_POLICY_VERSION_MINIMUM=3.5 during build with CMake 4.0 & newer though.

kekopom:
@luadebug indeed I can see that the flag is well propagated to cmake 
You will find below the full outputs of the command `xmake l scripts/test.lua -v -D --shallow -p android --ndk=~/android-ndk-r27c --ndk_sdkver=32 -a arm64-v8a opencv`
On my side I cannot understand what's wrong 😞 

[opencv_android.log](https://github.com/user-attachments/files/20072622/opencv_android.log)

luadebug:
> leads to a different error, I assume that it fixed the error for now
I suppose it leads to
```
error: D:\xm\xmake-repo\packages\o\opencv\xmake.lua:311: ...amdir\core\sandbox\modules\import\core\tool\compiler.lua:84: @programdir\modules\core\tools\gcc.lua:1020: C:\Users\Admin\AppData\Local\Temp\.xmake\250507\_C595EFB0B9474EA38934169C8482FD35.cpp:2:10: fatal error: 'opencv2/opencv.hpp' file not found
    2 | #include <opencv2/opencv.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```
Since include folder is empty.

kekopom:
Yep I fixed it by adding `package:add("includedirs", "sdk/native/jni/include")`
I think it does the same as your suggestion? :)

kekopom:
I think I've been able to finally build it
[0m[38;2;255;255;0m  => [0minstall opencv 4.11.0 .. [38;2;0;255;0;1mok[0m

packages/o/opencv/xmake.lua
```diff
@@ -97,6 +99,12 @@ package("opencv")
             local arch = (package:is_arch("x86_64") and "x64" or "x86")
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             package:add("linkdirs", path.join(arch, "mingw", linkdir))
+        elseif package:is_plat("android") then
+            local arch = package:targetarch()
+            local linkdir = (package:config("shared") and "lib" or "staticlibs")
```
luadebug, line 104: `sdk\native\libs\arm64-v8a\libopencv_signal.so`
So this should be
```suggestion
            local linkdir = (package:config("shared") and "libs" or "staticlibs")
```
Though CI is probably going to work since it static only, ```sdk/native/staticlibs/arm64-v8a/libopencv_bgsegm.a```

packages/o/opencv/xmake.lua
```diff
@@ -223,7 +241,19 @@ package("opencv")
             end
             package:add("links", reallink)
         end
-        if package:is_plat("windows") then
+        if package:is_plat("android") then
+            local linkdir = (package:config("shared") and "lib" or "staticlibs")
+            local libfiles = {}
+            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native", linkdir, package:targetarch(), "lib*.a")))
+            table.join2(libfiles, os.files(path.join(package:installdir(), "sdk/native/3rdparty/libs", package:targetarch(), "lib*.a")))
```
luadebug, line 248: This would skip `.so` files during shared build.

packages/o/opencv/xmake.lua
```diff
@@ -223,7 +241,19 @@ package("opencv")
...
+            for _, f in ipairs(libfiles) do
+                if not f:match("opencv_.+") then
+                    package:add("links", path.basename(f):match("lib(.+)"))
+                end
+            end
```
luadebug, line 253: It looks like current intention is to add 3rdparty into `links` so build would work, try this
```xmake
package:add("links", "ade", "cpufeatures", "libjpeg-turbo", "libpng", "libprotobuf", "libtiff", "libwebp", "quirc", "tegra_hal", "zlib")
```
instead of 245 -> 253

packages/o/opencv/xmake.lua
```diff
@@ -223,7 +241,19 @@ package("opencv")
...
+            package:addenv("PATH", "bin")
```
luadebug, line 254: I'm not sure this need, since nothing being installed into `bin` folder.

packages/o/opencv/xmake.lua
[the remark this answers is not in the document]
luadebug: Moved to https://github.com/xmake-io/xmake-repo/pull/7065#discussion_r2076939469

packages/o/opencv/xmake.lua
```diff
@@ -97,6 +99,12 @@ package("opencv")
             local arch = (package:is_arch("x86_64") and "x64" or "x86")
             local linkdir = (package:config("shared") and "lib" or "staticlib")
             package:add("linkdirs", path.join(arch, "mingw", linkdir))
+        elseif package:is_plat("android") then
+            local arch = package:targetarch()
```
luadebug, line 103: Is not being used.

kekopom: done thank you!

packages/o/opencv/xmake.lua
kekopom: added a new local libname that is set to lib*.a or lib*.so according to shared or not.
switched to libs instead of lib,
thanks!

packages/o/opencv/xmake.lua
kekopom: I did just as it was done for others platforms, your solution seems elegant but what if opencv adds a new lib?
since you hardcoded the libs, you would need to add it there, is that what we want?

kekopom: Also I think the user can select the feature`s he wants to embedded:
see `local features and `local default_features`

To me it seems that we want to link only the libs the user wants, isn't it?

kekopom:
Both (shared / static) builds are OK

packages/o/opencv/xmake.lua
luadebug: > seems elegant but what if opencv adds a new lib?

OK, it was just test, but indeed in real scenario it would be good to follow existing implementation.
```xmake
        if package:is_plat("android") then
            for _, suffix in ipairs({"*.a", "*.so"}) do
                for _, f in ipairs(os.files(path.join(package:installdir(path.join("sdk/native/3rdparty/libs", package:targetarch())), suffix))) do
                    package:add("links", path.basename(f):match("lib(.+)"))
                end
            end
            package:addenv("PATH", "bin")
```
So it would be done just like for others platforms.
> that we want to link only the libs the user wants

It builds only libs that user specified by options, the idea of add_links is to place libraries into correct link order, so library could be built successfully.

packages/o/opencv/xmake.lua
kekopom: Builds ok for shared, can we assume that if it builds ok we can just remove it?
Or we need to actually test the opencv lib in real scenario?

luadebug: > Builds ok for shared, can we assume that if it builds ok we can just remove it?

I think it can be safely removed, no bin folder with executable being either created. This indeed is part of windows/mingw handler so on_test would work, since they produce executable during build they test built executables like
```xmake
        if not package:is_cross() then
            -- FIXME, it will crash on action ci/2022/MD, but it works on local machine
            if not package:has_runtime("MD", "MDd") then
                if package:debug() and package:is_plat("windows", "mingw") then
                    os.vrun("opencv_versiond")
                else
                    os.vrun("opencv_version")
                end
            end
        end
```

>Or we need to actually test the opencv lib in real scenario?

That would be good, but it is not mandatory, as idea of `on_test` is to check functions from library, so symbols would get resolved from built linked libraries. Sometimes to check if built executable is working as well.

kekopom: alright removed, thanks a lot!

packages/o/opencv/xmake.lua
kekopom: Applied your suggestion that works like a charm, thanks!

kekopom:
seems related to  #1228

kekopom:
I've been able to build and run this exemple https://docs.opencv.org/4.x/d9/d15/samples_2cpp_2cout_mat_8cpp-example.html

On my Android phone!

packages/o/opencv/xmake.lua
```diff
@@ -223,7 +240,19 @@ package("opencv")
...
+            for _, suffix in ipairs({"*.a", "*.so"}) do
+                for _, f in ipairs(os.files(path.join(package:installdir(path.join("sdk/native/3rdparty/libs", package:targetarch())), suffix))) do
+                    package:add("links", path.basename(f):match("lib(.+)"))
+                end
+            end
+            for _, f in ipairs(libfiles) do
+                if not f:match("opencv_.+") then
+                    package:add("links", path.basename(f):match("lib(.+)"))
+                end
+            end
```
luadebug, line 253: Two for loops they look like repeating the same thing?

kekopom: Good catch **thanks**! Builds are still ok + runtime ok without the second loop

kekopom:
@luadebug thanks for helping fixing windows build :D 
with your patch, it complains to not find links

Could it be linked with `elseif package:is_arch("arm64") then arch = "ARM64"` line 255?
Since you added `-DAARCH64=ON` should it be now aarch64 right?

luadebug:
It looks like linkdir for windows `arm64` is `x64` for `local arch`. Need some workaround like new implementation would support old implementation as well..
C:/Users/runneradmin/AppData/Local/.xmake/packages/o/opencv/4.11.0/954bb460d4834f52b7ff0e8b3ff4d938/x64/vc17/staticlib/opencv_dpm4110.lib

kekopom:
@luadebug do you know what's happening?
`C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\wchar.h(254): error C2664: '__n64 __uint64x1_t_to_n64(uint64x1_t)': cannot convert argument 1 from 'uint16x4_t' to 'uint64x1_t'`

Only related topic I could find is https://developercommunity.visualstudio.com/t/wcharh-header-fails-compilation-on-ARM6/10868106?sort=newest
with no solution in it ..

Btw since we haven't changed anything for Windows ARM I would have expected it to work, I am a bit surprised

luadebug:
> @luadebug do you know what's happening? `C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\wchar.h(254): error C2664: '__n64 __uint64x1_t_to_n64(uint64x1_t)': cannot convert argument 1 from 'uint16x4_t' to 'uint64x1_t'`
> Only related topic I could find is https://developercommunity.visualstudio.com/t/wcharh-header-fails-compilation-on-ARM6/10868106?sort=newest with no solution in it ..

Why is this being mentioned, is it some existing Github CI failure?
The only mention is https://github.com/microsoft/vcpkg/issues/44310 as it is Windows SDK version breaking changes.

> Btw since we haven't changed anything for Windows ARM I would have expected it to work, I am a bit surprised

Maybe new version update broke something.

kekopom:
So what do we do?
can we disable windows arm support for now?

luadebug:
> So what do we do? can we disable windows arm support for now?

It looks like this package has windows@arm64 introduced exactly for 4.11.0...
```
    if on_check then
        on_check("windows|arm64", function (package)
            if package:version() and package:version():lt("4.10.0") then
                raise("current opencv version does not support windows/arm64!")
            end
            local vs = package:toolchain("msvc"):config("vs")
            assert(tonumber(vs) >= 2022, "opencv requires Visual Studio 2022 and later for arm targets")
        end)
    end
```

I wish we could limit Windows SDK that being used by 2022 worker to be newer version and at same time keep mine workaround, and find fix links for Windows-11 arm worker at same time.

luadebug (changes requested):
Need to fix links for Windows-11 arm worker I guess so.

luadebug:
>  Since you added `-DAARCH64=ON` should it be now aarch64 instead of "ARM64"? (in case of arm only)

Afaik  -DARM in case of `arm` but not `arm64` for arm64 it is -DAARCH64

> Could it be linked with `elseif package:is_arch("arm64") then arch = "ARM64"` line 255?

I resolved this by os.trymv but now executable issues for /MT vs runtime...

kekopom:
reverting as it is worst xd

luadebug:
Well now need to check if MSVC Windows-2025 worker is being skipped as well as MSVC Windows-2022 worker, since they have same issue... https://github.com/opencv/opencv/issues/27401
Expected windows-11-arm worker succeed.

waruqi:
Is it ready now?

packages/o/opencv/xmake.lua
```diff
@@ -141,8 +141,6 @@ package("opencv")
             end
             local vs = package:toolchain("msvc"):config("vs")
             assert(tonumber(vs) >= 2022, "opencv requires Visual Studio 2022 and later for arm targets")
-            local vs_sdkver = package:toolchain("msvc"):config("vs_sdkver")
-            assert(vs_sdkver and semver.match(vs_sdkver):eq("10.0.26100") and os.arch() == "arm64", "package(opencv): requires host arch to be arm for vs_sdkver 10.0.26100")
```
luadebug: I am unsure if this is proper way for recipe, but yet there is no another way since it is MSVC recent update toolchain bug. 
Github Worker does grab `10.0.26100.0` but it is too buggy for OpenCV & some different projects.
Github Worker does not grab `10.0.26100.4188 (released 5/2025)` to either have a chance to build it. https://github.com/microsoft/vcpkg/issues/44310#issuecomment-2944966048
Maybe would improve situation https://github.com/actions/runner-images/pull/12363

luadebug: Maybe it is good to wait for https://github.com/actions/runner-images/issues/12316...

luadebug: Awaiting for https://github.com/actions/runner-images/issues/12506

luadebug: Resolved by https://github.com/actions/runner-images/issues/12506

luadebug:
> Is it ready now?

Yes it is ready.

PR #7049, opened 2025-05-01
(gettext) add version 0.22.5 0.23.1 0.24.1 and 0.25 and fix libiconv

Arthapz:
* Before adding new features and new modules, please go to issues to submit the relevant feature description first.
* Write good commit messages and use the same coding conventions as the rest of the project.
* Please commit code to dev branch and we will merge into master branch in feature
* Ensure your edited codes with four spaces instead of TAB.

------

* 增加新特性和新模块之前，请先到issues提交相关特性说明，经过讨论评估确认后，再进行相应的代码提交，避免做无用工作。
* 编写友好可读的提交信息，并使用与工程代码相同的代码规范，代码请用4个空格字符代替tab缩进。
* 请提交代码到dev分支，如果通过，我们会在特定时间合并到master分支上。
* 为了规范化提交日志的格式，commit消息，不要用中文，请用英文描述。

PR #7088, opened 2025-05-07
mpg123: update to 1.32.10

luadebug:
https://github.com/xmake-io/xmake-repo/pull/7049/files

### DIFF
--- a/packages/m/mpg123/xmake.lua
+++ b/packages/m/mpg123/xmake.lua
@@ -40,8 +40,8 @@ package("mpg123")
             -- using execv instead of vrunv is needed when compiling on windows, otherwise it fail with
             -- ../libtool: line 3331: 0: Bad file descriptor
             -- when -v is not supplied to xmake
-            os.vrunv("make", argv, {shell = true})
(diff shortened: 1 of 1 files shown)


packages/m/mpg123/xmake.lua
```diff
@@ -4,23 +4,47 @@ package("mpg123")
...
+            local njob = option.get("jobs") or tostring(os.default_njob())
+            local argv = {"-j" .. njob}
+            if option.get("verbose") then
+                table.insert(argv, "V=1")
+            end
+            -- using execv instead of vrunv is needed when compiling on windows, otherwise it fail with
+            -- ../libtool: line 3331: 0: Bad file descriptor
+            -- when -v is not supplied to xmake
+            os.execv("make", argv, {shell = true})
+            os.execv("make", table.join({"install"}, argv), {shell = true})
```
waruqi, line 44: we should use package.tools.make to do it.

and make is not shell, we should not set `shell = true`

luadebug:
```
./libtool: line 3112: 0: Bad file descriptor
make: *** [Makefile:3581: src/libmpg123/libmpg123.la] Error 1
```

PR #7117, opened 2025-05-14
vulkanscenegraph: add v1.1.10

ecoezen:
version update and version-dependent config additions.


ecoezen:
removed unnecessary CMAKE_PREFIX_PATH handling.

ecoezen:
fails due to unhandled vulkansdk, so that failing find_package(Vulkan). any hints how to fix this in CI @waruqi ? local tests for Android, macos, windows and linux are fine.

ecoezen:
I think this can be merged as vulkansdk is not installed on CI, thus, cannot be tested. local tests are okay. fyi @waruqi

PR #7124, opened 2025-05-14
readline: extend platform support

Doekin:
This PR improves platform support for `readline`.

**Side Note: Impact of `CFLAGS` on `autotools` Optimization Levels**

A noteworthy behavior was identified with `autotools`-based builds:
Setting the `CFLAGS` environment variable overrides the default compiler flags that `autoconf` would otherwise establish. Critically, if these user-supplied `CFLAGS` lack optimization specifications (e.g., `-O2`), the build process defaults to the compiler's base optimization level, typically `-O0`.

This is distinct from CMake's approach, where build-type-specific flags (e.g., `CMAKE_C_FLAGS_RELEASE`) are additive to `CMAKE_C_FLAGS`, generally ensuring optimized release builds. The `autotools` behavior can inadvertently result in packages being built without intended optimizations if `CFLAGS` is set.

### DIFF
--- a/packages/n/ncurses/xmake.lua
+++ b/packages/n/ncurses/xmake.lua
@@ -45,6 +45,8 @@ package("ncurses")
             "--with-gpm=no",
             "--without-tests",
             "--without-ada",
+            "--enable-pc-files",
+            "--with-pkg-config-libdir=" .. path.unix(package:installdir("lib", "pkgconfig")):gsub("^(%a):", "/%1"),
         }
 
         table.insert(configs, "--with-debug=" .. (package:is_debug() and "yes" or "no"))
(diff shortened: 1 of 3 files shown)


packages/n/ncurses/xmake.lua
```diff
@@ -20,6 +20,9 @@ package("ncurses")
     end
 
     on_load(function (package)
+        if package:is_cross() then
+            package:add("deps", "ncurses~host", {kind = "binary", private = true})
```
waruqi, line 24: we need not add `~host` label, and modify kind, you can see https://github.com/xmake-io/xmake-repo/blob/1002ec9bf6d4be1bde579e073f73f61f1d61ccd4/packages/l/linux-headers/xmake.lua#L26

Issue #7127, opened 2025-05-14
`mbedtls` has wrong sha256sum inherited from upstream


mirageAlchemy:
### Xmake Version

2.9.9

### Operating System Version and Architecture

Linux login0 5.15.0-117-generic #127-Ubuntu SMP Fri Jul 5 20:13:28 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Describe Bug

-> % cat mbedtls-3.6.1-sha256sum.txt
fc8bef0991b43629b7e5319de6f34f13359011105e08e3e16eed3a9fe6ffd3a3
 mirage@anonymous [~] 
-> % sha256sum v3.6.1.zip                                                                               
bca632585454dac63de0c425d251671594c7f2b3c86a56fc3bade51b70d160e9  v3.6.1.zip

---
I understand that we in xrepo directly copied the sha256sum from the [upstream](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.1). But the thing is that upstream seems to calculate the sha256sum from a otherwhere released tar ball, not the one in github that xrepo pulls. See https://github.com/Mbed-TLS/mbedtls/issues/6984 .

### Expected Behavior

when we download the mbedtls it should download and verify the tarball successfully(if no network is comprimised).

### Project Configuration

/

### Additional Information and Error Logs

/

luadebug:
Download order would be `.tar.bz` if it failed to download, it would fallback to `.zip`, but intended behavior is to download `.tar.bz` & build on top of `.tar.bz` rather download `.zip`, hash itself directly points to `tar.bz` checksum.

> I understand that we in xrepo directly copied the sha256sum from the [upstream](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.1).

https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.1
> # Checksum
> The SHA256 hashes for the archives are:
> fc8bef0991b43629b7e5319de6f34f13359011105e08e3e16eed3a9fe6ffd3a3 mbedtls-3.6.1.tar.bz2

It supposed to work if `.tar.bz2` was pulled successfully.
> => download https://github.com/Mbed-TLS/mbedtls/releases/download/v3.6.1/mbedtls-3.6.1.tar.bz2 .. ok

But indeed, `.zip` indeed would have different size/checksum. On top of that `.zip` would consist of different files packed inside, according to note.
>Note
[...]

PR #7128, opened 2025-05-14
mbedtls: fix url

luadebug:
https://github.com/xmake-io/xmake-repo/issues/7127


packages/m/mbedtls/xmake.lua
```diff
@@ -4,7 +4,11 @@ package("mbedtls")
     set_license("Apache-2.0")
 
     add_urls("https://github.com/Mbed-TLS/mbedtls/releases/download/$(version).tar.bz2", {version = function (version)
-        return string.format("%s/mbedtls-%s", version, tostring(version):sub(2))
+        if version:lt("3.6.1") then
+            return string.format("%s/mbedtls-%s", version, tostring(version):sub(2))
```
waruqi, line 8: we can use `version:sub(2)` directly, need not tostring

Issue #7136, opened 2025-05-15
Add replxx


Willaaaaaaa:
### Is your feature request related to a problem? Please describe.

Although the GNU Readline library (GPLv3) is already included in xmake-repo, some projects are not GPLv3, so a C/C++ library with a loose license is needed, such as replxx (BSD).

replxx: A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.

### Describe the solution you'd like

I will submit a pr later.

PR #7137, opened 2025-05-15
replxx: add package

Willaaaaaaa:
#7136

### DIFF
--- a/packages/r/replxx/xmake.lua
+++ b/packages/r/replxx/xmake.lua
@@ -12,7 +12,7 @@ package("replxx")
     add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = "true"})
     
     on_load(function(package)
-        if is_plat("linux", "bsd") then
+        if package:is_plat("linux", "bsd") then
             package:add("syslinks", "pthread")
         end
         if not package:config("shared") then
@@ -23,7 +23,7 @@ package("replxx")
     on_install("windows", "linux", "macosx", "bsd", function (package)
         local configs = {"-DCMAKE_CXX_STANDARD=11"}
         
-        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DREPLXX_BUILD_EXAMPLES=" .. (package:config("examples") and "ON" or "OFF"))
         table.insert(configs, "-DREPLXX_BUILD_PACKAGE=" .. (package:config("package") and "ON" or "OFF"))
@@ -37,7 +37,7 @@ package("replxx")
                 #include <replxx.hxx>
                 int main() {
                     replxx::Replxx rx;
-                    return 0;
+                    rx.invoke(replxx::Replxx::ACTION::CLEAR_SELF, 0);
                 }
             ]]
         }, {configs = {languages = "c++11"}}))
--- a/packages/r/replxx/xmake.lua
+++ b/packages/r/replxx/xmake.lua
@@ -9,15 +9,24 @@ package("replxx")
     add_deps("cmake")
     add_configs("examples",{description = "Build the examples", default = false, type = "boolean"})
     add_configs("package", {description = "Generate package target", default = false, type = "boolean"})
-    add_configs("shared", {description = "Build shared library", default = false, type = "boolean"})
+    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = "true"})
+    
+    on_load(function(package)
+        if is_plat("linux", "bsd") then
+            package:add("syslinks", "pthread")
+        end
+        if not package:config("shared") then
+            package:add("defines", "REPLXX_STATIC")
+        end
+    end)
 
-    on_install(function (package)
+    on_install("windows", "linux", "macosx", "bsd", function (package)
         local configs = {"-DCMAKE_CXX_STANDARD=11"}
-
+        
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DREPLXX_BUILD_EXAMPLES=" .. (package:config("examples") and "ON" or "OFF"))
-        table.insert(configs, "-DREPLXX_BUILD_PACKAGE=" .. (package:config("package")  and "ON" or "OFF"))
+        table.insert(configs, "-DREPLXX_BUILD_PACKAGE=" .. (package:config("package") and "ON" or "OFF"))
 
         import("package.tools.cmake").install(package, configs)
     end)
--- a/packages/r/replxx/xmake.lua
+++ b/packages/r/replxx/xmake.lua
@@ -8,6 +8,14 @@ package("replxx")
 
     add_deps("cmake")
 
+    if on_check then
+        on_check("android", function (package)
+            local ndk = package:toolchain("ndk")
+            local ndk_sdkver = ndk:config("ndk_sdkver")
+            assert(ndk_sdkver and tonumber(ndk_sdkver) > 23, "package(replxx): need ndk api level > 23")
+        end)
+    end
+
     on_load(function(package)
         if package:is_plat("linux", "bsd") then
             package:add("syslinks", "pthread")
--- a/packages/r/replxx/xmake.lua
+++ b/packages/r/replxx/xmake.lua
@@ -7,7 +7,7 @@ package("replxx")
     add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
 
     add_deps("cmake")
-    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = "true"})
+    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = true})
     
     on_load(function(package)
         if package:is_plat("linux", "bsd") then
@@ -20,10 +20,8 @@ package("replxx")
 
     on_install("windows", "linux", "macosx", "bsd", function (package)
         local configs = {"-DCMAKE_CXX_STANDARD=11"}
-        
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
-
         import("package.tools.cmake").install(package, configs)
     end)
 
--- a/packages/r/replxx/xmake.lua
+++ b/packages/r/replxx/xmake.lua
@@ -1,0 +1,35 @@
+package("replxx")
+    set_homepage("https://github.com/AmokHuginnsson/replxx")
+    set_description("A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+    add_configs("examples",{description = "Build the examples", default = false, type = "boolean"})
+    add_configs("package", {description = "Generate package target", default = false, type = "boolean"})
+    add_configs("shared", {description = "Build shared library", default = false, type = "boolean"})
+
+    on_install(function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
+
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        table.insert(configs, "-DREPLXX_BUILD_EXAMPLES=" .. (package:config("examples") and "ON" or "OFF"))
+        table.insert(configs, "-DREPLXX_BUILD_PACKAGE=" .. (package:config("package")  and "ON" or "OFF"))
+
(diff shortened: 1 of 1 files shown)


packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+    add_configs("examples",{description = "Build the examples", default = false, type = "boolean"})
+    add_configs("package", {description = "Generate package target", default = false, type = "boolean"})
+    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = "true"})
+    
+    on_load(function(package)
+        if is_plat("linux", "bsd") then
```
luadebug, line 15:
```suggestion
        if package:is_plat("linux", "bsd") then
```

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+        end
+        if not package:config("shared") then
+            package:add("defines", "REPLXX_STATIC")
+        end
+    end)
+
+    on_install("windows", "linux", "macosx", "bsd", function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
+        
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:debug() and "Debug" or "Release"))
```
luadebug, line 26:
```suggestion
        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
```
>Wether the the package gets built with debug mode (deprecated: use [package:is_debug](https://xmake.io/#/manual/package_instance?id=packageis_debug) instead)

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({
+            test = [[
+                #include <replxx.hxx>
+                int main() {
+                    replxx::Replxx rx;
+                    return 0;
```
luadebug, line 40:
```suggestion
                    rx.invoke(replxx::Replxx::ACTION::CLEAR_SELF, 0);
```
We need to improve test, so it would test some function, `return 0` does not need here.

Willaaaaaaa:
Thank you for all the suggestions above! They really get the point.😍👍
There is one thing I don't quite understand: when should I test the function? I have noticed that some tests only need to include headers. Does this only apply to header-only libraries?

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+    set_homepage("https://github.com/AmokHuginnsson/replxx")
+    set_description("A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+    add_configs("examples",{description = "Build the examples", default = false, type = "boolean"})
+    add_configs("package", {description = "Generate package target", default = false, type = "boolean"})
```
star-hengxing, line 11: We don't need to build examples and package, Please remove it.
```suggestion
```

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,44 @@
...
+        end
+    end)
+
+    on_install("windows", "linux", "macosx", "bsd", function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
+        
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        table.insert(configs, "-DREPLXX_BUILD_EXAMPLES=" .. (package:config("examples") and "ON" or "OFF"))
+        table.insert(configs, "-DREPLXX_BUILD_PACKAGE=" .. (package:config("package") and "ON" or "OFF"))
```
luadebug, line 29:
```suggestion

```

Willaaaaaaa:
Got it! I thought setting the default value to false was enough.🥲🫡

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,40 @@
+package("replxx")
+    set_homepage("https://github.com/AmokHuginnsson/replxx")
+    set_description("A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = "true"})
```
waruqi, line 10: `"true"` -> true, not string

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,40 @@
...
+            package:add("defines", "REPLXX_STATIC")
+        end
+    end)
+
+    on_install("windows", "linux", "macosx", "bsd", function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
+        
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+
```
waruqi, line 26: remove empty line

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,40 @@
...
+            package:add("syslinks", "pthread")
+        end
+        if not package:config("shared") then
+            package:add("defines", "REPLXX_STATIC")
+        end
+    end)
+
+    on_install("windows", "linux", "macosx", "bsd", function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
+
```
waruqi, line 23: remove empty line

Willaaaaaaa:
So many tiny details that need to pay attention. Thank you for your patient review and correction!❤

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,38 @@
+package("replxx")
+    set_homepage("https://github.com/AmokHuginnsson/replxx")
+    set_description("A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+    add_configs("shared",  {description = "Build shared library", default = false, type = "boolean", readonly = true})
```
star-hengxing, line 10: I try shared build on linux, it work fine. If unsupported windows, use
```lua
if is_plat("windows") then
    add_configs("shared", {description = "Build shared library", default = false, type = "boolean", readonly = true})
end
```

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,38 @@
...
+        if package:is_plat("linux", "bsd") then
+            package:add("syslinks", "pthread")
+        end
+        if not package:config("shared") then
+            package:add("defines", "REPLXX_STATIC")
+        end
+    end)
+
+    on_install("windows", "linux", "macosx", "bsd", function (package)
+        local configs = {"-DCMAKE_CXX_STANDARD=11"}
```
star-hengxing, line 22: There is no need to explicitly set `CMAKE_CXX_STANDARD`.

---

https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L29-L38
We can disable these options to speed up compilation.
```lua
local configs = {"-DREPLXX_BUILD_EXAMPLES=OFF", "-DREPLXX_BUILD_PACKAGE=OFF"}
```

packages/r/replxx/xmake.lua
```diff
@@ -0,0 +1,37 @@
+package("replxx")
+    set_homepage("https://github.com/AmokHuginnsson/replxx")
+    set_description("A readline and libedit replacement that supports UTF-8, syntax highlighting, hints and Windows and is BSD licensed.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/AmokHuginnsson/replxx.git")
+    add_versions("2021.11.25", "1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2")
+
+    add_deps("cmake")
+
```
luadebug, line 10:
```suggestion

    if on_check then
        on_check("android", function (package)
            local ndk = package:toolchain("ndk")
            local ndk_sdkver = ndk:config("ndk_sdkver")
            assert(ndk_sdkver and tonumber(ndk_sdkver) > 23, "package(replxx): need ndk api level > 23")
        end)
    end

```

packages/r/replxx/xmake.lua
luadebug:
```
add_patches("2021.11.25", "https://github.com/AmokHuginnsson/replxx/compare/master%2E%2E%2ESpaceIm:fix/windows-dllexport.diff", "0fa4fbce85f4e73de32117ed6ab6bf0c79f7f9172797a2ddc882ac785f97b160")
```
This would add missing ones, but it is not recommended.

Willaaaaaaa: > I try shared build on linux, it work fine. If unsupported windows, use
> 
> ```lua
> if is_plat("windows") then
>     add_configs("shared", {description = "Build shared library", default = false, type = "boolean", readonly = true})
> end
> ```

It seems that shared build on windows works, too. I think I add this config just because [the initial cmakelists set shared as default](https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L40-L43).
image

Sry for this line, I think I mislead you guys to think this repo dont support shared build.

packages/r/replxx/xmake.lua
Willaaaaaaa: I explicitly set `CMAKE_CXX_STANDARD=11` mainly because cmakelists set default c++ std=11(see: https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L10-L12).
So I just added this line. I thought this line would help🥲

---

Yes! I forgot these dependent options. I deleted `add_configs("package")` & `add_configs("example")`, and forgot to add these configs back.🥲

Willaaaaaaa: Thanks a lot!! I got it! Even if the user uses `--cxxflags="-std=c++23"`, xmake will not pass it to cmake in this particular situation. So no need to worry about errors when `std>=c++20`. (`replxx` redefines keyword 'char8_t')
Finally know why there is no need to explicitly set it.
image

packages/r/replxx/xmake.lua
Willaaaaaaa: I am not familiar with Android dev. Thank you very much for your modification and the attached link. Really helped me understand!

packages/r/replxx/xmake.lua
Willaaaaaaa: > ```lua
> add_patches("2021.11.25", "https://github.com/AmokHuginnsson/replxx/compare/master%2E%2E%2ESpaceIm:fix/windows-dllexport.diff", "0fa4fbce85f4e73de32117ed6ab6bf0c79f7f9172797a2ddc882ac785f97b160")
> ```
> 
> This would add missing ones, but it is not recommended, since allowed is only to download & store diff for unmerged PR. Like this
> 
> Details

I have tried to shared build on both linux and windows. All worked fine.
As a newcomer, I don't know if I need to add this patch. But this patch really makes sense.❤

star-hengxing: > It seems that shared build on windows works, too. I think I add this config just because [the initial cmakelists set shared as default](https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L40-L43).

`add_configs("shared", {description = "Build shared library", default = false, type = "boolean", readonly = true})`

`readonly` will permanently set this configuration to `false`. If you need to set a default option value, simply use `default`. However, in `xrepo`, it's not always necessary to stay synchronized with the upstream options.

packages/r/replxx/xmake.lua
luadebug: C/CXX std are being set according to https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L10-L16 there is no need to explicitly set them.

> So no need to worry about errors when std>=c++20. (replxx redefines keyword 'char8_t')

It should emit warning `warning: identifier ‘char8_t’ is a keyword in C++20 [-Wc++20-compat]` & succeed, but do not prevent compilation & emit error... 🤔🤔🤔 what errors are you mentioning?

packages/r/replxx/xmake.lua
luadebug: C/CXX std are being set according to https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L10-L16 there is no need to explicitly set them.

> So no need to worry about errors when std>=c++20. (replxx redefines keyword 'char8_t')

It should emit warning `warning: identifier ‘char8_t’ is a keyword in C++20 [-Wc++20-compat]` & succeed, but do not prevent compilation & emit error... 🤔🤔🤔 what errors are you mentioning?

Willaaaaaaa: > It should emit warning `warning: identifier ‘char8_t’ is a keyword in C++20 [-Wc++20-compat]` & succeed, but do not prevent compilation & emit error... 🤔🤔🤔 what errors are you mentioning?

Yes, you are totally right. This is a misunderstanding caused by my carelessness. I have reproduced it carefully and found that what you said is completely true.
In fact, the source code has guaranteed to [avoid errors](https://github.com/AmokHuginnsson/replxx/blob/1f149bfe20bf6e49c1afd4154eaf0032c8c2fda2/src/conversion.hxx#L12). So, no errors anyway. I do apologize for confusing you!😭👉👈

Willaaaaaaa: > `readonly` will permanently set this configuration to `false`. If you need to set a default option value, simply use `default`. However, in `xrepo`, it's not always necessary to stay synchronized with the upstream options.

Noted! Thanks for sharing your understandings. I'm aware of it.🌹

PR #7168, opened 2025-05-19
Update gtest to v1.17.0.

Mr-Second:


star-hengxing:
see https://github.com/xmake-io/xmake-repo/issues/7152#issuecomment-2888954013

PR #7195, opened 2025-05-23
sqlite3-lemon: add new package

luadebug:
https://archlinux.org/packages/core/x86_64/lemon/
TODO: limit platforms.

### DIFF
--- a/packages/s/sqlite3-lemon/xmake.lua
+++ b/packages/s/sqlite3-lemon/xmake.lua
@@ -1,5 +1,4 @@
 package("sqlite3-lemon")
-
     set_kind("binary")
     set_homepage("https://sqlite.org/")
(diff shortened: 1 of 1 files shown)


packages/s/sqlite3-lemon/xmake.lua
```diff
@@ -0,0 +1,49 @@
...
+        io.writefile("xmake.lua", [[
+            add_rules("mode.debug", "mode.release")
+            set_encodings("utf-8")
+
+            target("lemon")
+                set_kind("binary")
+                add_files("tool/lemon.c")
+        ]])
+        import("package.tools.xmake").install(package)
+        package:addenv("PATH", "bin")
```
star-hengxing, line 44: binary package will automatically add this

PR #7253, opened 2025-05-31
add: libhat

SMlc666:


luadebug (approved):

packages/l/libhat/xmake.lua
```diff
@@ -0,0 +1,81 @@
...
+        local CMakeLists_content = io.readfile("CMakeLists.txt")
+        io.writefile("CMakeLists.txt", CMakeLists_content .. [[
+            if(LIBHAT_MODULE_TARGET AND TARGET libhat_module)
+                install(TARGETS libhat_module
+                    EXPORT libhat_module-targets FILE_SET CXX_MODULES DESTINATION include RUNTIME DESTINATION "bin" ARCHIVE DESTINATION "lib" LIBRARY DESTINATION "lib")
+            endif()
+        ]])
+        local opt = {}
+        if package:config("module") and package:has_tool("cxx", "cl") then
+            opt.cxflags = {"/EHsc", "/experimental:module"}
```
qudix: `/experimental:module` is deprecated and will likely be removed in a future VS version

luadebug: > `/experimental:module` is deprecated and will likely be removed in a future VS version

Yet it lacks of info how to guard it properly, since it is not removed yet. https://learn.microsoft.com/en-us/cpp/build/reference/experimental-module?view=msvc-170&viewFallbackFrom=msvc-170.

qudix: From what I can see in the [libhat source](https://github.com/BasedInc/libhat/blob/70e92a1e19fbaa9bb8cf8c34246eb591d6337bec/module/libhat.cppm#L42), it's not using msvc's dead experimental modules, just the standard ones, so that doesn't matter in this case.

luadebug: Well some users might still have outdated MSVC, so they would stick with /MD runtime, /EHsc and /experimental:module at this case anyway? Yet I decided to remove this as it has not been released tag and is sort of nightly feature.

PR #7206, opened 2025-05-25
ncnn: new package

std-microblock:
https://github.com/Tencent/ncnn
>ncnn is a high-performance neural network inference framework optimized for the mobile platform.

PR #7267, opened 2025-06-01
libomp: add mingw platform support

luadebug:
Attempt to add `MinGW` support to `omp` package.
Might improve https://github.com/xmake-io/xmake-repo/pull/7206 MinGW build.
Note, MacOS@MinGW much more likely supposed to use uasm.exe from MinGW environment rather use uasm from Mac OS environment. Yet I believe according to our recent discussion MinGW support is in a progress to be improved, so I decided to fallback for uasm by Mac OS environment. (I do not have another way to solve this).

### DIFF
--- a/packages/l/libomp/xmake.lua
+++ b/packages/l/libomp/xmake.lua
@@ -11,6 +11,11 @@ package("libomp")
 
     add_resources("19.1.0", "llvm_cmake", "https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.0/cmake-19.1.0.src.tar.xz", "dc78b6a9ac8a097ca6ac0f23c06821d65e6ea3bf666026f529994c1d01056ae7")
 
+    if is_plat("mingw", "msys") then
+        add_patches("19.1.0", "patches/001-cast-to-make-gcc-happy.patch", "11352ffbe7559a7170f2abd52b3552c877fbcf8fc82cff77b421e8b130a4dd66")
+        add_patches("19.1.0", "patches/002-hacks-for-static-linking.patch", "08e39ea52a99204528740196a13cc29daf1b65a6e230fbd7bdd745dde5d11ef3")
+    end
+
     add_configs("shared", {description = "Build shared library.", default = true, type = "boolean"})
 
     on_fetch("macosx", "linux", function (package, opt)
@@ -20,8 +25,12 @@ package("libomp")
     end)
 
     add_deps("cmake")
+    if is_plat("mingw", "msys") then
+        add_deps("uasm")
+    end
 
     add_links("omp")
+
     if is_plat("macosx") then
         add_extsources("brew::libomp")
     elseif is_plat("linux") then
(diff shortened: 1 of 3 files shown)


packages/l/libomp/xmake.lua
```diff
@@ -56,6 +65,12 @@ package("libomp")
 
         local configs = {"-DOPENMP_STANDALONE_BUILD=ON", "-DOPENMP_ENABLE_LIBOMPTARGET=OFF", "-DLIBOMP_INSTALL_ALIASES=OFF"}
         local shared = package:config("shared")
+        if package:is_plat("mingw", "msys") then
+            if package:is_arch("x86_64", "x64") then
+                table.insert(configs, "-DLIBOMP_ASMFLAGS=-win64")
+            end
+            table.insert(configs, "-DCMAKE_ASM_MASM_COMPILER=" .. path.join(package:dep("uasm"):installdir("bin"), "uasm.exe"))
```
luadebug: Probably would not work for MinGW@MacOS because it builds not .exe but relies over Mac OS build for uasm, possible workaround to remove .exe in case of MacOS subhost.

luadebug:
UASM v2.57, Jun  1 2025, Masm-compatible assembler.
Portions Copyright (c) 1992-2002 Sybase, Inc. All Rights Reserved.
```
Source code is available under the Sybase Open Watcom Public License.

Fatal error A1106: Cannot open file: "/safeseh" [ENOENT]
Fatal error A1106: Cannot open file: "/coff" [ENOENT]
z_Windows_NT-586_asm.asm: 1298 lines, 2 passes, 1784 ms, 0 warnings, 0 errors
```
It looks like binary of UASM works for MacOS@MinGW but it does not work for i386 for some reason, one of possible workaround is to enforce build & run of uasm through MinGW instead of MacOS...

luadebug:
```CMake
# Assembler flags
function(libomp_get_asmflags asmflags)
  set(asmflags_local)
  libomp_append(asmflags_local "-x assembler-with-cpp" LIBOMP_HAVE_X_ASSEMBLER_WITH_CPP_FLAG)
  # Architectural assembler flags
  if(${IA32})
    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
      libomp_append(asmflags_local -m32 LIBOMP_HAVE_M32_FLAG)
    endif()
    libomp_append(asmflags_local /safeseh LIBOMP_HAVE_SAFESEH_MASM_FLAG)
    libomp_append(asmflags_local /coff LIBOMP_HAVE_COFF_MASM_FLAG)
  elseif(${MIC})
    libomp_append(asmflags_local -mmic LIBOMP_HAVE_MMIC_FLAG)
  endif()
  set(asmflags_local ${asmflags_local} ${LIBOMP_ASMFLAGS})
  libomp_setup_flags(asmflags_local)
  set(${asmflags} ${asmflags_local} PARENT_SCOPE)
endfunction()
```
So once I replaced:
/safeseh => -safeseh
/coff => -coff
It seems working now for MacOS@MinGW i386 as well. Looks like Windows MinGW i386 converts /safeseh to -safeseh I guess (since executable is being run into MinGW), unlike Mac OS unable to do such fixup?
Program expects every argument being past with **-** not with **/**

PR #7283, opened 2025-06-03
add openmvs, fix some packages

choyy:


luadebug:
I gave up with MSVC 2019(is already dropped)/2022 ARM64 **OLD** workers support for OpenCV4. See https://github.com/xmake-io/xmake-repo/pull/7065 due these failure build logs. :/ But I kept **NEWEST** (Native ARM64 CPU build) ARM64 github CI worker building OpenCV4.
Their recent solution is to use CLang? https://github.com/microsoft/vcpkg/issues/44310
In TL;DR it is better to deal with opencv4 in first place, I guess?

luadebug:
https://github.com/xmake-io/xmake-repo/pull/7286
Github CI Worker has been promoted from win2019 to win2025.

choyy:
@star-hengxing 求助，看不懂了

image

star-hengxing:
目前这两个 linux ci 用的 gcc15，估计与此相关

luadebug:
I'm pretty sure MinGW would keep gcc 15 as well, besides MinGW might drop MINGW32 support this year.
image

PR #7291, opened 2025-06-04
Fix a dylib problem that fails to find onnxruntime

KnightJun:
在 macos 的M2 芯片环境上调试 x86-64 架构的时候会出现dyld[12794]: Library not loaded: @rpath/libonnxruntime.1.16.1.dylib

### DIFF
--- a/packages/o/onnxruntime/xmake.lua
+++ b/packages/o/onnxruntime/xmake.lua
@@ -5,6 +5,7 @@ package("onnxruntime")
 
     add_configs("gpu", {description = "Enable GPU support on windows|x64 and linux|x86_64", default = false, type = "boolean"})
     add_configs("cuda_version", {description = "Specify which CUDA version to use for GPU support", default = "11", type = "string"})
+    add_configs("shared", {description = "Download shared binaries.", default = true, type = "boolean", readonly = true})
 
     if is_plat("windows") then
         if is_arch("x64") then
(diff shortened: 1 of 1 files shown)


star-hengxing:
如果是 shared 包本来就会加。
https://github.com/xmake-io/xmake/blob/d75240a009a7f295e31568c83fe9e93de1be431d/xmake/core/package/package.lua#L1136-L1141

失败的原因是，因为目前 onnxruntime 是下的预编译二进制是 shared，但包描述里并没有强制设置为 shared。
`add_configs("shared", {description = "Download shared binaries.", default = true, type = "boolean", readonly = true})`
但 xrepo 默认包为静态库，所以没有添加到环境变量。

packages/o/onnxruntime/xmake.lua
```diff
@@ -110,6 +110,9 @@ package("onnxruntime")
             os.mv("lib/*.dll", package:installdir("bin"))
         end
         os.cp("*", package:installdir())
+        if package:is_plat("macosx") then
+            package:addenv("DYLD_LIBRARY_PATH", path.join(package:installdir(), "lib"))
```
waruqi: use relative path

luadebug: Just `package:addenv("DYLD_LIBRARY_PATH", "lib")` ?

waruqi: right

waruqi: and add `package:mark_as_pathenv("DYLD_LIBRARY_PATH")`

waruqi: btw, we need not add lib to DYLD_LIBRARY_PATH

xmake has added it for shared library.


https://github.com/xmake-io/xmake/blob/6989184ed0b7e61614d3f41bdbcdac4c6075ac77/xmake/core/package/package.lua#L1139

star-hengxing: I think I already explained this above...

PR #7296, opened 2025-06-05
Auto-update uasm to v213

waruqi:
New version of uasm detected (package version: v2.57r, last github version: v213)


luadebug:
image
image
How?

Issue #7288, opened 2025-06-04
Detach the nlohmann/json dependency from libhv


zixiai:
### Is your feature request related to a problem? Please describe.

Detach and replace [[libhv/cpputil/json.hpp](https://github.com/ithewei/libhv/blob/master/cpputil/json.hpp)](https://github.com/ithewei/libhv/blob/master/cpputil/json.hpp) with the nlohmann_json library in xrepo, as the original implementation is too outdated.

### Describe the solution you'd like

https://github.com/zixiai/xmake-repo/blob/ce8fb023e8eba0093cd2784db8d65ddd32574a89/packages/l/libhv/xmake.lua#L49

https://github.com/zixiai/xmake-repo/blob/ce8fb023e8eba0093cd2784db8d65ddd32574a89/packages/l/libhv/xmake.lua#L73-L90

### Describe alternatives you've considered

I'm not sure whether to add an option for nlohmann_json, since the default version is v3.9.1. After replacing it, some users might not prefer the latest version.

PR #7299, opened 2025-06-05
lodepng: add `cpp` config & libhv: debundle nlohmann_json

luadebug:
adds cpp config for `lodepng` package
https://github.com/xmake-io/xmake-repo/issues/7298
debundles nlohmann_json 
https://github.com/xmake-io/xmake-repo/issues/7288


packages/l/lodepng-с/xmake.lua
```diff
@@ -0,0 +1,26 @@
+package("lodepng-с")
+    set_homepage("https://lodev.org/lodepng/")
+    set_description("PNG encoder and decoder in C and C++.")
+    set_license("zlib")
+
+    add_urls("https://github.com/lvandeve/lodepng.git")
+    add_versions("2025.05.06", "17d08dd26cac4d63f43af217ebd70318bfb8189c")
+
+    on_install(function (package)
+        os.mv("lodepng.cpp", "lodepng.c")
```
star-hengxing: Why?

luadebug: https://github.com/lvandeve/lodepng/blob/master/README.md
> Building
> 
> Only two files are needed to encode and decode PNGs:
> 
>     lodepng.cpp (or renamed to lodepng.c)
>     lodepng.h
> 
> All other files are just source code examples, tests, misc utilities, etc..., which are normally not needed in projects using this.
> 
> You can include the files directly in your project's source tree and its makefile, IDE project file, or other build system. No library is necessary.
> 
> In addition to C++, LodePNG also supports ANSI C (C89), with all the same functionality: C++ only adds extra convenience API.
> 
> For C, rename lodepng.cpp to lodepng.c.
> 
> Consider using git submodules to include LodePNG in your project.

It looks like vcpkg just kept lodepng-c port. https://github.com/microsoft/vcpkg/tree/master/ports/lodepng https://github.com/microsoft/vcpkg/tree/master/ports/lodepng-c

star-hengxing: > In addition to C++, LodePNG also supports ANSI C (C89), with all the same functionality: C++ only adds extra convenience API.

There's no need to create a new package description. You can just add new configs directly within the `lodepng` package.

star-hengxing:
I don't understand why you did this, but generally, one issue should correspond to one PR, unless these two packages are related to each other.

PR #7341, opened 2025-06-11
glib: update to 2.85.0

luadebug:

------

### DIFF
--- a/packages/g/glib/xmake.lua
+++ b/packages/g/glib/xmake.lua
@@ -1,5 +1,4 @@
 package("glib")
-
     set_homepage("https://docs.gtk.org/glib/")
     set_description("Low-level core library that forms the basis for projects such as GTK+ and GNOME.")
     set_license("LGPL-2.1")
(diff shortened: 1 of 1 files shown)


packages/g/glib/xmake.lua
```diff
@@ -40,8 +40,11 @@ package("glib")
     elseif is_plat("macosx") then
         add_deps("libiconv", {system = true})
         add_deps("libintl")
-    elseif is_plat("windows") then
-        add_deps("libintl", "pkgconf")
+    elseif is_plat("windows", "mingw") then
+        add_deps("libintl")
+        if is_plat("windows") then
```
star-hengxing: is_subhost

luadebug: Why? If we go for `is_subhost("windows")` this would skip `Mac OS X @ MinGW`?

star-hengxing: Except for subhost windows, we default to other subhosts having pkg-config. But there will be exceptions, so I think using this is best. Do you have any other suggestions?
https://github.com/xmake-io/xmake-repo/blob/77845d621e81f941a259559838504cad71947b66/packages/l/libgit2cpp/xmake.lua#L14-L18

luadebug: I thought everything is simple here, since its inside of `is_plat("windows", "mingw")`. We just need add pkgconf for windows, MinGW has pkg-config installed already to use?

In this example its pkgconf to add for Windows / Windows @ MinGW / Windows @ WASM ... etc...?
Else add pkg-config to every another?

star-hengxing: See https://github.com/xmake-io/xmake/issues/5474

luadebug: OK so in this case, workaround is
```lua
if is_subhost("windows") then
    add_deps("pkgconf")
else
    add_deps("pkg-config")
end
```
?

PR #7346, opened 2025-06-11
base64-terrakuh: add new package

luadebug:
Needs for `cpp-mcp` https://github.com/hkr04/cpp-mcp


SirLynix:
Name is a bit generic :-/

luadebug:
> Name is a bit generic :-/

terrakuh-base64 or base64-terrakuh ?

SirLynix:
I'd say base64-terrakuh is probably better except if more terrakuh libraries are expected

PR #7365, opened 2025-06-13
taocpp-json: add new package

luadebug:
Latest pegtl release wont work, since too old.
```
C:\Users\Admin\AppData\Local\.xmake\packages\j\json\2025.03.11\36a027cd729542b880966f8d2d7625bd\include/tao/json/jaxn/internal/grammar.hpp: In static member function 'static bool tao::json::jaxn::internal::rules::sor_value_impl<String, Binary>::match_must(Input&, States&& ...)':
C:\Users\Admin\AppData\Local\.xmake\packages\j\json\2025.03.11\36a027cd729542b880966f8d2d7625bd\include/tao/json/jaxn/internal/grammar.hpp:245:77: error: 'optional' is not a member of 'tao::pegtl::rewind_mode' [-Wtemplate-body]
  245 |             return Control< must< Rule > >::template match< A, rewind_mode::optional, Action, Control >( in, st... );
```

### DIFF
--- a/packages/t/taocpp-json/xmake.lua
+++ b/packages/t/taocpp-json/xmake.lua
@@ -17,7 +17,7 @@ package("taocpp-json")
 pkg_check_modules(pegtl REQUIRED pegtl)]], {plain = true})
         io.replace("CMakeLists.txt",
             [[target_link_libraries(taocpp-json INTERFACE taocpp::pegtl)]],
-            [[target_include_directories(taocpp-json ${pegtl_INCLUDE_DIRS})]], {plain = true})
+            [[target_include_directories(taocpp-json PRIVATE ${pegtl_INCLUDE_DIRS})]], {plain = true})
         local configs = {
             "-DTAOCPP_JSON_BUILD_TESTS=OFF",
             "-DTAOCPP_JSON_BUILD_EXAMPLES=OFF",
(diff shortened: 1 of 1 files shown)


packages/t/taocpp-json/xmake.lua
```diff
@@ -0,0 +1,37 @@
...
+    set_homepage("https://github.com/taocpp/json")
+    set_description("C++ header-only JSON library")
+    set_license("MIT")
+
+    add_urls("https://github.com/taocpp/json.git", {submodules = false})
+    add_versions("2025.03.11", "11a31e12eda35c1322f9cf6ebb4cca0653d579dc")
+
+    add_deps("cmake")
+    add_deps("pkgconf")
+    add_deps("pegtl 54a2e32bf4593ed86782b4882702286cc8d621f9")
```
waruqi, line 12: why not use version?

luadebug: 🦺 Well it does not define `rewind_mode::optional` so compilation would fail. https://github.com/taocpp/json/blob/11a31e12eda35c1322f9cf6ebb4cca0653d579dc/include/tao/json/internal/grammar.hpp#L169
Does it work for you?
So my workaround is stick to latest changes, rather latest released tag.
https://github.com/taocpp/PEGTL/blob/3.2.8/include/tao/pegtl/rewind_mode.hpp

waruqi: we can add patches for latest released version.

luadebug: Too much changes between 3.2.8 & unreleased 4.0.0 to just use patch.

waruqi: ok

PR #7372, opened 2025-06-13
Fix meson 1.8.2 hash

SirLynix:
@luadebug what do you use to compute checksum? I think it's the second time fix it after you


luadebug:
image
7z. Im pretty sure I mistake over `https://github.com/mesonbuild/meson/releases/download/1.8.2/meson-1.8.2.tar.gz` & I download wrong one `https://github.com/mesonbuild/meson/archive/refs/tags/1.8.2.tar.gz`

PR #7375, opened 2025-06-13
protobuf-cpp: update to 31.1

luadebug:


packages/p/protobuf-cpp/xmake.lua
```diff
@@ -180,7 +181,7 @@ package("protobuf-cpp")
         table.insert(configs, "-Dprotobuf_BUILD_LIBUPB=" .. (package:config("upb") and "ON" or "OFF"))
 
         local opt = {}
-        opt.buildir = "build"
+        opt.builddir = "build"
```
star-hengxing: It maybe break xmake 2.x user

packages/p/protobuf-cpp/xmake.lua
```diff
@@ -58,7 +59,7 @@ package("protobuf-cpp")
 
     on_load(function (package)
         -- Fix MSVC 2019 arm64 error LNK2019: unresolved external symbol __popcnt referenced in function _upb_log2_table_size
-        if package:is_plat("windows") then
+        if package:version():eq("31.0") and package:is_plat("windows") then
```
star-hengxing: `package:version() and package:version():eq()`
When use git clone, `package:version()` maybe nil

PR #7461, opened 2025-06-21
update microsoft-proxy v3.4.0

ecoezen:
update microsoft-proxy v3.4.0


ecoezen:
It's a false fail due to ci failure. there is no freebsd build anyway. can be merged if you all are okay with th3 current state @waruqi

PR #7497, opened 2025-06-25
Update imgui to v1.92.0

AidanSun05:


AidanSun05:
The previous CI errors (mingw, shared build) were caused by conflicting declarations in the imgui source. I've included the patch I sent on the imgui repo to fix this.

PR #7509, opened 2025-06-27
ppqsort: add package

Cappecasper03:

------

### DIFF
--- a/packages/p/ppqsort/xmake.lua
+++ b/packages/p/ppqsort/xmake.lua
@@ -8,18 +8,15 @@ package("ppqsort")
 
     add_versions("v1.0.5", "39a973a680eb0af3bd0bdd5f4e9fa81d484915f3141e3a4568930647a328ba12")
 
+    on_check(function (package)
+        assert(package:has_cxxincludes("syncstream", {configs = {languages = "c++20"}}), "package(ppqsort): need <syncstream> header.")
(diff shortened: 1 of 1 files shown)


packages/p/ppqsort/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+        local has_cxx20 = package:has_cxxflags("-std=c++20")
+        local has_syncstream = package:has_cxxincludes("syncstream", {configs = {languages = "c++20"}})
+        local has_ranges = package:has_cxxincludes("ranges", {configs = {languages = "c++20"}})
+
+        if has_cxx20 and has_syncstream and has_ranges then
+            assert(package:has_cxxincludes("ppqsort.h", {configs = {languages = "c++20"}}))
+        else
+            wprint("C++20 not supported by the current toolchain, skipping test.")
+        end
+    end)
```
luadebug: I think on_test should test package before use in any case.

Cappecasper03: I could of course be wrong about the reason they fail, but how would we fix ci actions that fail because of not having these features?

star-hengxing: https://github.com/xmake-io/xmake-repo/blob/11e1b752a7cb9fca8d05991667a0c4ec5928d8ea/packages/r/reflect-cpp/xmake.lua#L39-L58

PR #7510, opened 2025-06-27
arduinojson: update to v7.4.2

apocelipes:
v7.4.1 -> v7.4.2


packages/a/arduinojson/xmake.lua
```diff
@@ -28,7 +29,7 @@ package("arduinojson")
             #include <ArduinoJson.h>
             void test() {
                 char json[] = "{\"sensor\":\"gps\",\"time\":1351824120,\"data\":[48.756080,2.302038]}";
-                DynamicJsonDocument doc(1024);
+                JsonDocument doc;
```
star-hengxing, line 35: It will break old version?

apocelipes: > It will break old version?

Breaks 6.x, fixed.

PR #7515, opened 2025-06-27
standardlib: update to 2024.03.25

luadebug:
It looks like releases and commit history was cleaned by upstream, and it is now just mirror of his so called *foundationallib*.

### DIFF
--- a/packages/s/standardlib/xmake.lua
+++ b/packages/s/standardlib/xmake.lua
@@ -30,5 +30,5 @@ package("standardlib")
                 const char *suffix = "text";
                 printf("Does the string end with \"%s\"? %s\n", suffix, ends_with(text, suffix) ? "Yes" : "No");
             }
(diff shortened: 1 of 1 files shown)


packages/s/standardlib/xmake.lua
```diff
@@ -1,24 +1,34 @@
...
+        assert(package:check_csnippets({test = [[
+            #include <foundationallib.h>
             #include <stdio.h>
             void test() {
                 const char *text = "This is a sample text";
                 const char *suffix = "text";
                 printf("Does the string end with \"%s\"? %s\n", suffix, ends_with(text, suffix) ? "Yes" : "No");
             }
-        ]]}))
+        ]]}, {configs = {languages = "c20"}}))
```
star-hengxing: c20?

luadebug: strdup is probably c2X

PR #7531, opened 2025-06-28
wxwidgets: do not warn user if unable to fetch minor of version in order to concatenate it with major and patch strings

luadebug:
Trying to resolve https://github.com/xmake-io/xmake-repo/issues/7491

### DIFF
--- a/packages/w/wxwidgets/xmake.lua
+++ b/packages/w/wxwidgets/xmake.lua
@@ -140,12 +140,11 @@ package("wxwidgets")
 
     on_install("macosx", "linux", function (package)
         -- Notify the user about issues caused by the CMake version.
-        local cmake = package:dep("cmake")
-        local cmake_fetch = cmake:fetch()
-        local major, minor, patch = cmake_fetch.version:match("^(%d+)%.(%d+)%.(%d+)$")
-        local cmake_version = tonumber(major.. minor.. patch)
-        if cmake_version > 3280 then
-            wprint("\ncmake may not find Cmath detail in https://github.com/prusa3d/PrusaSlicer/issues/12169\n")
+        if package:dep("cmake") and package:dep("cmake"):version() then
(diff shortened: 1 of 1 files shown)


packages/w/wxwidgets/xmake.lua
```diff
@@ -140,12 +140,11 @@ package("wxwidgets")
 
     on_install("macosx", "linux", function (package)
         -- Notify the user about issues caused by the CMake version.
-        local cmake = package:dep("cmake")
-        local cmake_fetch = cmake:fetch()
-        local major, minor, patch = cmake_fetch.version:match("^(%d+)%.(%d+)%.(%d+)$")
-        local cmake_version = tonumber(major.. minor.. patch)
-        if cmake_version > 3280 then
-            wprint("\ncmake may not find Cmath detail in https://github.com/prusa3d/PrusaSlicer/issues/12169\n")
+        if package:dep("cmake") and package:dep("cmake"):version() then
```
waruqi: if cmake is found on system, package:dep("cmake"):version() will return nil. it is only for remote cmake package.

we should always use fetch api to find system or remote package.

luadebug: Thank you for review, I updated this to
```lua
    on_install("macosx", "linux", function (package)
        import("core.base.semver")
        local cmake = package:dep("cmake")
        if cmake then
            local cmake_fetch = cmake:fetch()
            -- Notify the user about issues caused by the CMake version.
            if cmake_fetch and cmake_fetch.version ~= nil then
                local cmake_version = semver.new(cmake_fetch.version)
                if cmake_version:major() > 3 or 
                    (cmake_version:major() == 3 and cmake_version:minor() > 28) then
                    print("CMake may not find CMath detail in https://github.com/prusa3d/PrusaSlicer/issues/12169")
                end
            end
        end
```

PR #7549, opened 2025-06-30
CI: fix root repo only automations running on forks

godot42x:


.github/workflows/deploy.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    if: ${{ github.repository == 'xmake-io/xmake-repo' }} # only runs on root repository
```
apocelipes: You can omit the external `${{ }}`.
[doc](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution)

image

.github/workflows/sync.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         os: [ubuntu-latest]
 
     runs-on: ${{ matrix.os }}
-    if: ${{ github.repository }} == "xmake-io/xmake-repo"
+    if: ${{ github.repository == 'xmake-io/xmake-repo' }} # only runs on root repository
```
apocelipes: You can omit the external `${{ }}`.

Doc:
image

apocelipes:
You can omit the external `${{ }}`.

Doc:
image

godot42x:
sure🚀

PR #7551, opened 2025-06-30
libuvc: add Linux/MacOS X platforms

choyy:


choyy:
Can macosx be fixed?

luadebug:
> Can macosx be fixed?

Maybe it could, but that would require to edit `libusb` package file at same time to find solution?

luadebug (approved):

PR #7558, opened 2025-07-01
libcoro: add package

DIG-:
Add libcoro package.
Maybe there is better way to set cxxflags based on compiler

### DIFF
--- a/packages/l/libcoro/xmake.lua
+++ b/packages/l/libcoro/xmake.lua
@@ -25,13 +25,6 @@ package("libcoro")
 
     on_check(function (package)
         assert(package:check_cxxsnippets({test = [[
-            #if __has_include(<version>)
-            #  ifndef __cpp_lib_jthread
-            #      error "package(libcoro): Feature-test macro for jthread missing in <version>"
-            #  endif
-            #endif
-        ]]}, {configs = {languages = "c++20"}}), "package(libcoro): Feature-test macro for jthread missing in <version>")
-        assert(package:check_cxxsnippets({test = [[
             #if defined(__clang__)
             #  if __clang_major__ < 16
             #      error "package(libcoro): Clang version too low, need at least 16.0.0"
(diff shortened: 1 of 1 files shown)


DIG-:
Most error is related to unsupported compiler version.
One looks like to require clang v20+, due to not fully support std::jthread since v17.

packages/l/libcoro/xmake.lua
```diff
@@ -0,0 +1,105 @@
...
+        table.insert(configs, "-DLIBCORO_FEATURE_NETWORKING=" .. (package:config("networking") and "ON" or "OFF"))
+        table.insert(configs, "-DLIBCORO_FEATURE_TLS=" .. (package:config("tls") and "ON" or "OFF"))
+
+        local opt = {}
+        if package:is_plat("mingw") and package:config("shared") then
+            opt.shflags = "-Wl,--export-all-symbols"
+        end
+        import("package.tools.cmake").install(package, configs, opt)
+
+        if package:is_plat("windows") and package:is_debug() then
```
star-hengxing: It can be removed. https://github.com/xmake-io/xmake/issues/5566

packages/l/libcoro/xmake.lua
```diff
@@ -0,0 +1,105 @@
...
+            package:add("deps", "openssl")
+        end
+
+        if not package:config("shared") then
+            package:add("defines", "CORO_STATIC_DEFINE")
+        end
+
+        _, toolname, _ = package:tool("cxx")
+        if toolname == "gxx" then
+            package:add("cxxflags", "-fcoroutines")
```
star-hengxing, line 65: Please use `has_tool("cxx", "gcc")` in `on_install` , `on_load` can't get toolchain info.

luadebug: Should it be into `opt` rather than into `package:add`?

DIG-: These flags are required not only to build the package, but to use it.
It is mandatory to the project inherit these flags.

luadebug:
cmake not found! during arch linux build but this issue does not appear during msys2/mingw for some reason.
Probably because cmake was not pre-installed on arch linux. So we might need some another solution into on_check.

luadebug (approved):

PR #7640, opened 2025-07-13
zeus_expected: update to v1.3.1

Willaaaaaaa:
CI in #7533 has shown that the source code of this repo has an error like this:
image

It demonstrated that the compiler can't access a private variable by a `friend` function.
Although I am confused by this error, I fixed it with a patch.


Willaaaaaaa:
Thanks for the branch changing. completely forgot to check the target branch.

The error message in the **previous** CI indicated that the **friend** function `operator==` and `!=` could **NOT** access the private member `m_val` within `zeus::unexpected` class.
In my opinion, this should **not** be an error. I don't understand whether it is a compiler bug or a repo author's bug, so I'm not sure if this is the best way to package. Please review it when you are free. Thanks!

The code in question: https://github.com/zeus-cpp/expected/blob/main/include/zeus/expected.hpp#L199-L211
To work around, I used this member function to get `m_val`: https://github.com/zeus-cpp/expected/blob/main/include/zeus/expected.hpp#L185

star-hengxing:
You can first report this error as an issue in the library repository.
After the original author fixes it, if they release a new version, we can ignore this version. If not, we can merge this patch.

Willaaaaaaa:
> You can first report this error as an issue in the library repository. After the original author fixes it, if they release a new version, we can ignore this version. If not, we can merge this patch.

Sounds good. I'll open an issue with a reproducible example for the author asap.

X1aomu:
https://github.com/zeus-cpp/expected/releases/tag/v1.3.1

A fixed version was just released. Thanks for your work!

PR #7649, opened 2025-07-14
Superlu: update to v7.0.1

Willaaaaaaa:
#7464 
From v7.0.1, `superlu` no longer provides a **static** `superlu_config.h` file. This `.h` file must be dynamically generated from the `superlu_config.h.in` template through CMake macros. The old `xmake.lua` completely bypassed CMake, resulting in `superlu_config.h` never being created in v7.0.1, which in turn caused a "file not found" error when executing `io.replace`.

Solution:
Use `io.writefile` to overwrite.


Willaaaaaaa:
Fixed macos CI. It seems that the old macos CI failed to find `gfortran`.

Willaaaaaaa:
Ready for review.

Issue #5989, opened 2024-12-16
Luau fails to pass tests with extern_c = true


raxracks:
### Xmake Version

2.9.6

### Operating System Version and Architecture

Linux laptop 6.11.0-1001-asahi-arm #1-Ubuntu SMP PREEMPT_DYNAMIC Sat Sep 28 14:42:09 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux

### Describe Bug

The test in the Luau package fails to link because the `extern_c` config flag adds `extern "C"` to all the functions meaning they don't have mangled names and can't be found when linking unless the test also has `extern "C"`, I would submit my own pull request but I don't know how to change the test based on config flags.

```/usr/bin/ld: /tmp/.xmake1000/241216/_30972899FDE349008EBF238727CE0510.o: in function `test()':
_9B8070A7849241B0B487B0797A8791EE.cpp:(.text+0x8): undefined reference to `luaL_newstate()'
/usr/bin/ld: _9B8070A7849241B0B487B0797A8791EE.cpp:(.text+0x14): undefined reference to `luaL_openlibs(lua_State*)'
/usr/bin/ld: _9B8070A7849241B0B487B0797A8791EE.cpp:(.text+0x1c): undefined reference to `lua_close(lua_State*)'```

### Expected Behavior

The test should pass, as it does with the `extern_c` flag disabled

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")

add_requires("luau", { configs = { extern_c = true } })

target("luau-test")
    set_kind("binary")
    add_files("src/*.c")
    add_packages("luau")
```

### Additional Information and Error Logs

```console
> /usr/bin/g++ -o /tmp/.xmake1000/241216/_E19865D3ADAD483080527B34368CEB50.b /tmp/.xmake1000/241216/_E19865D3ADAD483080527B34368CEB50.o -L/home/sam/.xmake/packages/l/luau/0.643/c73bc7e59e274cd6b9cc974db0c13fcd/lib -lLuau.VM -lLuau.CodeGen -lLuau.EqSat -lLuau.Analysis -lLuau.Config -lLuau.Compiler -lLuau.Ast
/usr/bin/ld: /tmp/.xmake1000/241216/_E19865D3ADAD483080527B34368CEB50.o: in function `test()':
_9B8070A7849241B0B487B0797A8791EE.cpp:(.text+0x8): undefined reference to `luaL_newstate()'
[...]

PR #7688, opened 2025-07-19
[WIP] luau: Update to 0.683

Willaaaaaaa:
- Update `luau` to the latest version **0.683** (from 0.643). 
- Fix `extern_c` config #5989 

> Since there are too many versions in between, maybe I will add some versions between them.


Willaaaaaaa:
The ci result above has shown that `luau` can be successfully built under the condition of `extern_c = true`. Now I'll change the default `extern_c` value to false.
And should I add more version to fill the gap between version 0.643 and 0.683? Just add 0.6x9 or sth else or nothing?

**Ready for review.**

packages/l/luau/xmake.lua
```diff
@@ -27,7 +28,7 @@ package("luau")
         ]], {plain = true})
```
Willaaaaaaa, line 28: Just asking, is it neccesary to change `L24-L28` to `table.insert(configs, "-DCMAKE_POLICY_DEFAULT_CMP0057=NEW")`? as the latter one is cleaner and does not require io operations. Maybe faster?

Willaaaaaaa:
Just asking, is it neccesary to change 
https://github.com/xmake-io/xmake-repo/blob/ddce8e81df1e3afbbcd9e19518a69c7f4c3a7a3f/packages/l/luau/xmake.lua#L24-L27
to
```xmake
table.insert(configs, "-DCMAKE_POLICY_DEFAULT_CMP0057=NEW")
```
as the latter one is cleaner and does not require io operations? less io operations maybe faster?

luadebug:
It is just better to insert value into configs table, rather replacing cmake_policy line that might be changed in next versions or either did not exist in early versions (or refered git commit checkout). As well as that looks simple, we just trying to avoid io.replace and use patch if possible most of time, as per we keep *port* cleaner if we do not use io.replace at all.
When you insert value into configs it is guaranteed to work so it is safe approach.

Willaaaaaaa:
> It is just better to insert value into configs table, rather replacing cmake_policy line that might be changed in next versions or either did not exist in early versions (or refered git commit checkout). As well as that looks simple, we just trying to avoid io.replace and use patch if possible most of time, as per we keep _port_ cleaner if we do not use io.replace at all.

Thanks a lot. Your point is clearly more convincing. Edited.

Willaaaaaaa:
@waruqi 
I think there are 16 PRs that should be closed if we needn't add the old versions.
see: https://github.com/xmake-io/xmake-repo/pulls?q=is%3Apr+is%3Aopen+luau

PR #7703, opened 2025-07-21
improve ninja

waruqi:
we do not need additional python dependencies


star-hengxing:
Remove windows & macosx precompiled binary url?

waruqi:
> Remove windows & macosx precompiled binary url?

暂时不删了，既然有二进制可用，不是更好么，都不用编译了。后面如果有遇到问题，再切源码

PR #7705, opened 2025-07-21
hpx: update to v1.11.0

Willaaaaaaa:
update hpx:
1. support more plats
2. opti configs according to docs
3. **[NEED HELP]** fix link errors


star-hengxing:
You can use **draft** pr instead WIP title name.

Willaaaaaaa:
The current `xmake.lua` still encounters **link errors** in the `on_test` scope.

My idea is to replicate CMake's link behavior. And I try to use string manipulation to get all the necessary libraries and link flags by parsing the file `HPXInternalTargets.cmake` which is generated after hpx installation.

This latest submission is **the best I can do**. Directly scanning lib and hard-coding it into the link list may be able to solve this problem, but this method cannot guarantee the link order and is not suitable.
In addition, I also noticed the `.pc` file, but I am not familiar with it, and hpx does not seem to **provide detailed `.pc` files for each lib**. I took a rough look at `.pc` file and it seems not feasible.

Please help me see if the current solution is correct and tell me any better workaround. Thank you very much!

Willaaaaaaa:
## My thinkings:

### 1. get hpx installed

### 2. take a look at `HPXConfig.cmake` file
  <details>
    <summary> I found the following code: </summary>

    <code>
    # Set legacy variables for linking and include directories
    set(HPX_LIBRARIES "HPX::hpx;")
    # All properties are propagated from HPX::hpx, so the following can be empty
    set(HPX_INCLUDE_DIRS "")
    set(HPX_LIBRARY_DIR "")
    set(HPX_LINK_LIBRARIES "")
    set(HPX_LINKER_FLAGS "")
    </code>

  </details>

### 3. So search for `HPX::hpx`
found 
```
set_target_properties(HPX::hpx PROPERTIES
  INTERFACE_LINK_LIBRARIES "HPXInternal::hpx_full;HPXInternal::hpx_interface"
)
```
in `HPXTargets.cmake`

### 4. Search for `HPXInternal::hpx_full` and `HPXInternal::hpx_interface`
found
```
set_target_properties(HPXInternal::hpx_full PROPERTIES
...
INTERFACE_LINK_LIBRARIES...
```
and `HPXInternal::hpx_interface` in `HPXInternalTargets.cmake`

### So use string manipulation to get all the libs and flags.

Willaaaaaaa:
Just to make a plan for moving forward. If there isn't any suggestion by **tomorrow noon**, I will proceed with a another plan.

My plan is to solve the linker errors directly by using **`nm`** command to find which library provides each missing symbol. I'll then add the required libraries based on this analysis. Even though this is a more "brute-force" method, I will avoid hard-coding as much as I can. 

The way I think obviously have some sort of logical flaw (although I think it's right). Anyway, I will try to find an elegant solution through existing `xmake.lua`s in the meantime.

star-hengxing:
There has always been no universal solution to the link order error caused by gnu ld.
- For libraries with fewer links, manually using `add_links` ensures dependency order.
- Read certain configuration files to obtain dependencies for topological sorting, such as the abseil package, which reads the `.pc` file.

I also think it can be done by building the shared library and then using `ldd` to get the dependency relationship, which can also detect some unbundle dependencies.

Willaaaaaaa:
Thank you sooooooo much for the great suggestions!!!

-   **Manual sorting:** 
  `hpx` will end up with ~100 `.a` files, which is obviously straightforward but tedious to handle. This can the last resort.

-   **`.pc` files:** 
  Actually, I've already got inspired by `abseil` and took a brief look at `.pc`. And they seemed incomplete for `hpx`'s static libs. I'll take a deeper look later to see if there is a way.

-   **`ldd` with shared libs:**
  This, is absolutely a wonderful idea I hadn't thought of! I will give that a try.

I will take a deeper look at `.pc` files. If it can work, then use `.pc`, otherwise continue to try cmake with `ldd`.

star-hengxing:
```lua
import("utils.binary.deplibs")
```
example
https://github.com/xmake-io/xmake/blob/5670477e415e24962c57eac21194c63946ee5e1e/xmake/modules/target/action/install/main.lua#L55

star-hengxing:
related issue: https://github.com/xmake-io/xmake/issues/6650

Willaaaaaaa:
I don't understand. I have already used `import("utils.binary.deplibs")`, but the function failed to analyze the deps.
I think I have use the module in the right way, but the outputs suggest that this function is not working.

And when I tested locally, I'm confused by the errors in win+shared, too. Why I get a "LINK : fatal error LNK1561: 必须定义入口点" error. I don't know how to fix this, because it works in linux when `-k shared`.

---

It seems that I need to fix `hwloc` first. Working on it now. It may take some time because I'm busy these days. But I will make time for this and get it done within my ability.

packages/h/hpx/xmake.lua
```diff
@@ -46,16 +60,105 @@ package("hpx")
...
+        if not package:config("shared") then
+            table.insert(configs, "-DHPX_WITH_STATIC_LINKING=ON")
+        end
+        if package:config("shared") and package:is_plat("windows") then
+            table.insert(configs, "-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON")
+        end
+        -- On almost all platforms except Linux (x86) and Windows, HPX requires Boost.Context.
+        -- https://hpx-docs.stellar-group.org/latest/html/manual/building_hpx.html#most-important-cmake-options
+        -- See: https://github.com/STEllAR-GROUP/hpx/issues/4829
+        if is_plat("linux", "windows") and is_arch("x86", "x86_64", "x64") then
```
waruqi: package:is_plat, package:is_arch

packages/h/hpx/xmake.lua
```diff
@@ -46,16 +60,105 @@ package("hpx")
...
+                        dep_linkname = dep_basename:gsub("lib(.-)%.a", "%1")
+                    end
+                    if hpx_libs_map[dep_linkname] then
+                        cprint("  ${cyan}%s ${yellow}-> ${green}%s", linkname, dep_linkname)
+                        dag:add_edge(linkname, dep_linkname)
+                    end
+                end
+            end
+
+            local sorted_links = dag:topological_sort()
```
waruqi: if enable recursive, it will get all deps and use toposort on dev version

https://github.com/xmake-io/xmake/blob/b2628769d3206449e9301cb86bd87df9c789ccd5/xmake/modules/utils/binary/deplibs.lua#L382-L395

and topological_sort is deprecated, we should use topo_sort

Willaaaaaaa:
Existing issues:
1. **cannot find `-l<libname>` error**:
  Currently, I'm using the hard-coded method to link libs. But in `debug` mode, `hpx` generates library files with a letter `d` suffix (e.g. `libhpx_iostreams**d**` instead of `libhpx_iostreams`). This one can be solved either by iterating through all lib files in installdir or handling strings in `.pc` file. (at least I suppose so.

2. **cmake errors**(under `Windows`):
  boost cmake failed & hpx cmake failed
  For both, I have no idea why. I'll take a closer look tomorrow. QAQ

3. **Undefined symbols**(under `macos`):
  like this log shows: https://github.com/xmake-io/xmake-repo/actions/runs/16862048205/job/47763662198?pr=7705#step:4:1200
  Maybe this can be fixed by `add_framework`? I think this may be because the frameworks that the `hwloc` library depends on on macos are not being passed down correctly. But idk why. Should this be passed down automatically?

I'm so sorry for taking such a long time to package this. As I said, this pkg is probably too complex for me : ( . And I'd appreciate any advice you guys can offer!

PR #7752, opened 2025-07-26
onnxruntime: update to 1.22.0

choyy:


choyy:
If a version is only for specific platforms, how should it be handled?

luadebug:
> If a version is only for specific platforms, how should it be handled?

Maybe it is better to split this into two PRs. One PR would contain WIndows OS versions that have been added. Another PR would contain another OS represented. I think it is being discussed and merged without using on_check and with windows os pipeline fails.

waruqi:
Does it ready now?

choyy:
> Does it ready now?

Yes, ready.

PR #7753, opened 2025-07-26
itk: update to 5.4.4

choyy:


packages/i/itk/xmake.lua
```diff
@@ -7,36 +7,50 @@ package("itk")
...
         for _, lib in ipairs(libs) do
             package:add("links", lib .. "-" .. ver)
         end
+        if package:is_plat("windows", "macosx") and package:config("shared") then
+            package:add("deps", "hdf5[shared,cpplib]")
```
waruqi: Please use a consistent style in xmake-repo

```lua
package:add("deps", "hdf5", {configs = {shared = true, cpplib = true}})
```

packages/i/itk/xmake.lua
```diff
@@ -7,36 +7,50 @@ package("itk")
...
+        else
+            package:add("deps", "hdf5[cpplib]")
```
waruqi: and here

PR #7754, opened 2025-07-26
openssl3: fix find by cmake

ChrisCatCP:
https://discourse.cmake.org/t/findopenssl-cmake-does-not-search-lib64-directory/9429

### DIFF
--- a/packages/o/openssl3/xmake.lua
+++ b/packages/o/openssl3/xmake.lua
@@ -135,6 +135,9 @@ package("openssl3")
         os.vrunv("perl", configs, {envs = buildenvs})
         import("package.tools.make").build(package)
         import("package.tools.make").make(package, {"install_sw"})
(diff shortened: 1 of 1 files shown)


packages/o/openssl3/xmake.lua
```diff
@@ -135,6 +135,9 @@ package("openssl3")
         os.vrunv("perl", configs, {envs = buildenvs})
         import("package.tools.make").build(package)
         import("package.tools.make").make(package, {"install_sw"})
+        if package:is_arch("x86_64") then
+            os.trycp(path.join(package:installdir("lib64"), "*"), package:installdir("lib"))
```
star-hengxing, line 139: More Robustness?
```lua
if os.isdir(package:installdir("lib64")) then
    os.cp(package:installdir("lib64/*"), package:installdir("lib"))
end
```

PR #7759, opened 2025-07-27
pciaccess: add package

Willaaaaaaa:
New package
hpx -> hwloc -> pciaccess

### DIFF
--- a/packages/l/libpciaccess/xmake.lua
+++ b/packages/l/libpciaccess/xmake.lua
@@ -1,0 +1,39 @@
+package("libpciaccess")
+    set_homepage("https://gitlab.freedesktop.org/xorg/lib/libpciaccess")
+    set_description("Generic PCI access library")
+    set_license("MIT")
+
+    add_urls("https://gitlab.freedesktop.org/xorg/lib/libpciaccess/-/archive/libpciaccess-$(version)/libpciaccess-libpciaccess-$(version).tar.gz")
+    add_urls("https://gitlab.freedesktop.org/xorg/lib/libpciaccess.git")
+    add_versions("0.18.1", "a395317730e0e8d5e71419d4d1256a89e32c2fa793607b63c4d0fb497ae34602")
+
+    add_configs("zlib", {description = "Enable zlib support to read gzip compressed pci.ids.", default = false, type = "boolean"})
+    add_configs("linux_rom_fallback", {description = "Enable support for falling back to /dev/mem for roms on Linux.", default = false, type = "boolean"})
+    add_configs("pci_ids_path", {description = "Path to pci ids. If relative is assumed relative to $datadir.", default = "hwdata", type = "string"})
+
+    add_deps("meson", "ninja")
+
+    on_load(function (package)
+        if package:config("zlib") then
+            package:add("deps", "zlib")
+        end
+    end)
+
+    on_install(function (package)
+        local configs = {}
+        if package:is_debug() then
+            table.insert(configs, "-Dbuildtype=debug")
+        else
+            table.insert(configs, "-Dbuildtype=release")
+        end
+
+        table.insert(configs, "-Dzlib=" .. (package:config("zlib") and "enabled" or "disabled"))
+        table.insert(configs, "-Dlinux-rom-fallback=" .. (package:config("linux_rom_fallback") and "true" or "false"))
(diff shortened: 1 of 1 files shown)


packages/l/libpciaccess/xmake.lua
```diff
@@ -0,0 +1,33 @@
...
+    set_license("MIT")
+
+    add_urls("https://gitlab.freedesktop.org/xorg/lib/libpciaccess/-/archive/libpciaccess-$(version)/libpciaccess-libpciaccess-$(version).tar.gz",
+             "https://gitlab.freedesktop.org/xorg/lib/libpciaccess.git")
+
+    add_versions("0.18.1", "a395317730e0e8d5e71419d4d1256a89e32c2fa793607b63c4d0fb497ae34602")
+
+    add_configs("zlib", {description = "Enable zlib support to read gzip compressed pci.ids.", default = false, type = "boolean"})
+    add_configs("linux_rom_fallback", {description = "Enable support for falling back to /dev/mem for roms on Linux.", default = false, type = "boolean"})
+    add_configs("pci_ids_path", {description = "Path to pci ids. If relative is assumed relative to $datadir.", default = "hwdata", type = "string"})
```
star-hengxing: `add_deps("hwdata")`?

Willaaaaaaa: I'm not sure about this. I think `hwdata` is a runtime dependence. Therefore, it should be set up by users.

And the option `pci_ids_path` should be a relative path to `get_option('prefix') + get_option('datadir')`. If we use `add_deps("hwdata")`, meson may fail to get the right path, unless we change the build logic in `meson`.

Willaaaaaaa: Yes... According to `meson.build`, I think the description here (which is copied from `meson_options.txt`) is not accurate enough. It must be a reletive path. The current description might mislead us into thinking this can be an absolute path.

```
config.set_quoted(
  'PCIIDS_PATH',
  join_paths(get_option('prefix'), get_option('datadir'), get_option('pci-ids'))
)
```

Willaaaaaaa: In my laptop, `hwdata` will be installed at `/home/willaaaaaaa/.xmake/packages/h/hwdata/v0.397/086c9ab82bcb4aa8a8f32a2a838b8815/share/hwdata` when using `xrepo install hwdata`.

If we use add_deps("hwdata"), maybe we don't need to change `meson.build`. Instead, we can add a `--prefix=` parameter? But idk how to specify this parameter. Is there a way? 

But still, I don't think we ought to add `add_deps("hwdata")`. Or is it a good practice if we change the `hwdata` package logic to make it install at the path we need?

waruqi: we should use `add_deps("hwdata", {private = true})` and tell its path to this package.

and please remove pci_ids_path

```lua
add_deps("hwdata", {private = true})
on_install(function (package)
    -- ...
    table.insert(configs, "-Dpci-ids=" .. package:dep("hwdata"):installdir("share/hwdata")) -- you can use correct path.
end)
```

waruqi: > Or is it a good practice if we change the hwdata package logic to make it install at the path we need?

No, each package is stored in an isolated sandbox and we cannot modify its path.

waruqi: > we can add a --prefix= parameter? But idk how to specify this parameter. Is there a way?

We should not specify `--prefix` to modify any installation paths.

Willaaaaaaa: > we should use `add_deps("hwdata", {private = true})` and tell its path to this package.
> 
> and please remove pci_ids_path
> 
> ```lua
> add_deps("hwdata", {private = true})
> on_install(function (package)
>     -- ...
>     table.insert(configs, "-Dpci-ids=" .. package:dep("hwdata"):installdir("share/hwdata")) -- you can use correct path.
> end)
> ```

Perfect! Fully understood. Thanks for your kind and detailed explaination!!🌹

Willaaaaaaa:
Add bsd support for dependence `hwdata`.
Reference: https://github.com/freebsd/freebsd-ports/blob/main/misc/hwdata/files/patch-Makefile

PR #7781, opened 2025-07-30
ndk: refactor & upgrade.

Redbeanw44602:
- add versions from r22 to r27.
 - add `macosx|arm64` support.
 - ci always uses the latest version of r22 and r27.

---

There were some issues with ndk 22.0 that caused my tests to fail, so it was necessary to upgrade.
https://github.com/android/ndk/issues/1448


Redbeanw44602:
I'm not sure if I should change the `elseif` to `else` to make the test pass for FreeBSD, as it seems to me that this compromises semantic correctness. Xmake should check `on_install` first to avoid this error.

waruqi:
> I'm not sure if I should change the `elseif` to `else` to make the test pass for FreeBSD, as it seems to me that this compromises semantic correctness. Xmake should check `on_install` first to avoid this error.

If FreeBSD does not support this package, you should not add urls without any versions.

PR #7797, opened 2025-08-01
openssl3: fix bugs and update to 3.5.1

Redbeanw44602:


Redbeanw44602:
The `cannot open program database` error seems to occur at random times (re-running the test always fixes it),  but the probability is not low. `/FS` is passed and I don't know what went wrong. 🤯

star-hengxing:
You can reference
https://github.com/xmake-io/xmake/issues/5566
https://github.com/xmake-io/xmake/issues/5353

Redbeanw44602:
> You can reference [xmake-io/xmake#5566](https://github.com/xmake-io/xmake/issues/5566) [xmake-io/xmake#5353](https://github.com/xmake-io/xmake/issues/5353)

I managed to generate the pdb to a certain path, but the problem is still not solved. Am I understanding this wrongly as to why the pdb would randomly become unwritable?

star-hengxing:
> I managed to generate the pdb to a certain path, but the problem is still not solved. Am I understanding this wrongly as to why the pdb would randomly become unwritable?

I don't know. If you can find a reproducible example, I can debug it. For now, we could consider disabling jom for certain versions and switching back to single-threaded compilation with nmake.

There is still another issue: xrepo package defaults to debug `false`, so why is a pdb file being generated here? Maybe we need to remove the debug flags.

Redbeanw44602:
I can't find a stable way to reproduce this.

But there are some points worth noting:
 - This problem only occurs on Windows 2025 and has never occurred on 2022.
 - I reran the tests several times and sometimes they passed perfectly. Most of the failures were on 3.3.2~3.4.2, which I think was because they were tested first. I observed one failure on 3.2.5.

I would run a test without jom at all to confirm if the problem lies with it.

Redbeanw44602:
I think this PR is ready to be merged 👀

PR #7853, opened 2025-08-10
libsolv: add package.

Redbeanw44602:
- libsolv
   - libzchunk

### DIFF
--- a/packages/l/libsolv/xmake.lua
+++ b/packages/l/libsolv/xmake.lua
@@ -1,0 +1,181 @@
+package("libsolv")
+    set_homepage("https://github.com/openSUSE/libsolv")
+    set_description("Library for solving packages and reading repositories.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/openSUSE/libsolv/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/openSUSE/libsolv.git")
+
+    add_versions("0.7.34", "fd9c8a75d3ca09d9ff7b0d160902fac789b3ce6f9fb5b46a7647895f9d3eaf05")
+
+    add_patches("<=0.7.34", "patches/fix-msvc-c2036.patch", "a924517033d4f8ba18e922e892953834d3ca1a4fa5a69ae04fd308df40d1b2e8")
+    add_patches("<=0.7.34", "patches/fix-compile-on-mingw-w64.patch", "e6ba565110c918363a4499a4fc949f29777e0a189f192c231c81a47da821d21d")
+
+    -- when using mingw, cmake cannot force export of all symbols.
+    if is_plat("mingw", "msys", "cygwin") then
+        add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
+    end
+
+     -- needs rpm, rpmdb, rpmio, rpmmisc, db
+    add_configs("rpmdb",             {description = "Build with rpm database support.", default = false, type = "boolean"})
+    add_configs("rpmdb_librpm",      {description = "Use librpm to access the rpm database.", default = false, type = "boolean"})
+    add_configs("rpmdb_bdb",         {description = "Use BerkeleyDB to access the rpm database.", default = false, type = "boolean"})
+    add_configs("rpmdb_byrpmheader", {description = "Build with rpmdb Header support.", default = false, type = "boolean"})
+    add_configs("rpmpkg",            {description = "Build with rpm package support.", default = false, type = "boolean"})
+    add_configs("rpmpkg_librpm",     {description = "Use librpm to access rpm header information.", default = false, type = "boolean"})
+    add_configs("pubkey",            {description = "Build with pubkey support.", default = false, type = "boolean"})
+    add_configs("rpmmd",             {description = "Build with rpmmd repository support.", default = false, type = "boolean"})
+    add_configs("rpm5",              {description = "Enabling RPM 5 support.", default = false, type = "boolean"})
+
+    add_configs("suserepo",  {description = "Build with suse repository support.", default = false, type = "boolean"})
+    add_configs("comps",     {description = "Build with fedora comps support.", default = false, type = "boolean"})
+    add_configs("helixrepo", {description = "Build with helix repository support.", default = false, type = "boolean"})
+    add_configs("debian",    {description = "Build with debian package/repository support.", default = false, type = "boolean"})
+    add_configs("mdkrepo",   {description = "Build with mandriva/mageia repository support.", default = false, type = "boolean"})
+    add_configs("archrepo",  {description = "Build with archlinux repository support.", default = false, type = "boolean"})
+    add_configs("apk",       {description = "Build with apk package/repository support.", default = false, type = "boolean"})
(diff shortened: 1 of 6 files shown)


packages/l/libzchunk/xmake.lua
```diff
@@ -0,0 +1,73 @@
...
+
+    add_versions("1.5.1", "2c187055e2206e62cef4559845e7c2ec6ec5a07ce1e0a6044e4342e0c5d7771d")
+
+    add_patches("<=1.5.1", "patches/fix-cdecl.patch", "7ca1cbabe8516152e5d4e5cd5dc7c14b2fd0134f8ad7a8fa64159e07998ebeb4")
+
+    if is_plat("wasm") then
+        add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
+    end
+
+    add_configs("with_zstd", {description = "Enable compression support.", default = false, type = "boolean"})
```
star-hengxing: Please remove `with_`

packages/l/libzchunk/xmake.lua
```diff
@@ -0,0 +1,73 @@
...
+        table.insert(configs, "-Dwith-openssl=" .. (package:config("with_openssl") and "enabled" or "disabled"))
+
+        io.replace("meson.build", "subdir('src')", "subdir('src/lib')", {plain = true})
+        io.replace("meson.build", "not argplib.found()", "false", {plain = true})
+        if is_plat("windows") then
+            -- fix dll name & fix install dir
+            io.replace("src/lib/meson.build", "soversion: so_version,", "install_dir: get_option('libdir'),", {plain = true})
+            io.replace("src/lib/meson.build", "version: meson.project_version(),", "", {plain = true})
+            -- fix lib name
+            io.replace("src/lib/meson.build", "lib_suffix = 'lib'", "", {plain = true})
```
star-hengxing: Generally, we only need to fix the library name if it affects other projects' find_package. Why do we need to fix it here?
https://github.com/xmake-io/xmake-repo/blob/8817c1086230b6ff68536544a7875180f95f213c/packages/z/zstd/xmake.lua#L54-L62

packages/l/libzchunk/xmake.lua
```diff
@@ -0,0 +1,73 @@
...
+        end
+
+        import("package.tools.meson").install(package, configs)
+        if is_plat("windows") and not package:config("shared") then
+            -- fix lib name
+            os.cd(package:installdir("lib"))
+            os.trymv("libzck.a", "zck.lib")
```
star-hengxing: And here

packages/l/libzchunk/xmake.lua
Redbeanw44602: I forgot, let me test what happens if I remove them first😵‍💫

Redbeanw44602: Everything looks fine😂

packages/l/libsolv/xmake.lua
```diff
@@ -0,0 +1,181 @@
...
+            if package:config(option) then
+                if not table.contains(no_prefix_options, option) then
+                    table.insert(configs, ("-DENABLE_%s=ON"):format(option:upper()))
+                else
+                    table.insert(configs, ("-D%s=ON"):format(option:upper()))
+                end
+            end
+        end
+
+        io.replace("CMakeLists.txt", "ADD_SUBDIRECTORY (tools)", "", {plain = true})
```
star-hengxing: Maybe we can make a config.
https://github.com/xmake-io/xmake-repo/blob/cb92bde1da66aab3fbb376a744808bfa66b0fd76/packages/l/libmobi/xmake.lua#L15
https://github.com/xmake-io/xmake-repo/blob/cb92bde1da66aab3fbb376a744808bfa66b0fd76/packages/l/libmobi/xmake.lua#L39-L41

packages/l/libsolv/xmake.lua
```diff
@@ -0,0 +1,181 @@
...
+    add_configs("appdata",   {description = "Build with AppStream appdata support.", default = false, type = "boolean"})
+    add_configs("multi_semantics", {description = "Build with support for multiple distribution types.", default = false, type = "boolean"})
+
+    add_configs("lzma_compression",   {description = "Build with lzma/xz compression support.", default = false, type = "boolean"})
+    add_configs("bzip2_compression",  {description = "Build with bzip2 compression support.", default = false, type = "boolean"})
+    add_configs("zstd_compression",   {description = "Build with zstd compression support.", default = false, type = "boolean"})
+    add_configs("zchunk_compression", {description = "Build with zchunk compression support.", default = false, type = "boolean"})
+
+    add_configs("with_system_zchunk", {description = "Use system zchunk library.", default = false, type = "boolean"})
+    add_configs("with_libxml2",       {description = "Build with libxml2 instead of libexpat.", default = false, type = "boolean"})
```
star-hengxing: Please remove `with_`

packages/l/libsolv/xmake.lua
```diff
@@ -0,0 +1,181 @@
...
+    add_configs("conda",     {description = "Build with conda dependency support.", default = false, type = "boolean"})
+    add_configs("appdata",   {description = "Build with AppStream appdata support.", default = false, type = "boolean"})
+    add_configs("multi_semantics", {description = "Build with support for multiple distribution types.", default = false, type = "boolean"})
+
+    add_configs("lzma_compression",   {description = "Build with lzma/xz compression support.", default = false, type = "boolean"})
+    add_configs("bzip2_compression",  {description = "Build with bzip2 compression support.", default = false, type = "boolean"})
+    add_configs("zstd_compression",   {description = "Build with zstd compression support.", default = false, type = "boolean"})
+    add_configs("zchunk_compression", {description = "Build with zchunk compression support.", default = false, type = "boolean"})
+
+    add_configs("with_system_zchunk", {description = "Use system zchunk library.", default = false, type = "boolean"})
```
star-hengxing: ~We can use `add_requireconf("libsolv.libzchunk", {override = true, system = true})`~
Generally speaking, the `use system library` option should be enabled by default, because disabling it will automatically fetch dependencies. However, the logic for `with_system_zchunk` in CMake is to call pkgconfig to search, so I think this config name should be changed to whether to enable zchunk.

packages/l/libsolv/xmake.lua
```diff
@@ -0,0 +1,181 @@
...
+    add_configs("with_libxml2",       {description = "Build with libxml2 instead of libexpat.", default = false, type = "boolean"})
+    if not is_plat("windows", "mingw", "msys", "cygwin") then
+        add_configs("without_cookieopen", {description = "Disable the use of stdio cookie opens.", default = false, type = "boolean"})
```
star-hengxing: - Please remove `without_`
- `Disable` -> `Enable`

packages/l/libsolv/xmake.lua
Redbeanw44602: I checked libsolv's CMakeLists.txt again, `WITH_SYSTEM_ZCHUNK` seems to just use PkgConfig to search for packages, I will try to enable it by default and remove that option (`zchunk_compression` already exists).

PR #7856, opened 2025-08-10
openvpn3: add package.

Redbeanw44602:


packages/o/openvpn3/xmake.lua
```diff
@@ -0,0 +1,68 @@
+package("openvpn3")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://openvpn.github.io/openvpn3/")
+    set_description("OpenVPN 3 is a C++ class library that implements the functionality of an OpenVPN client.")
+
+    add_urls("https://github.com/OpenVPN/openvpn3/archive/refs/tags/release/$(version).tar.gz", {alias = "tarball"})
+    add_urls("https://github.com/OpenVPN/openvpn3.git", {alias = "git"})
+    add_versions("tarball:3.11.3", "4a5d18059d6270bd103e290ed4e3bc773e838bc3b22c861408190869fd43074e")
+    add_versions("git:3.11.3", "release/3.11.3")
```
star-hengxing, line 9: This alias probably isn’t needed...

packages/o/openvpn3/xmake.lua
```diff
@@ -0,0 +1,68 @@
...
+
+    add_configs("use_mbedtls", {description = "Use mbed TLS instead of OpenSSL.", default = false, type = "boolean"})
```
star-hengxing: Please remove `use_`

packages/o/openvpn3/xmake.lua
```diff
@@ -0,0 +1,68 @@
...
+    if is_plat("windows", "msys", "mingw", "cygwin") then
+        add_resources("*", "tap-windows6", "https://raw.githubusercontent.com/OpenVPN/tap-windows6/refs/heads/master/src/tap-windows.h", "0b9ea5e4b4dc2c2764fde3383440af9a11a1ead3f2a4bbec674be40da384d10e")
+        add_defines("_WIN32_WINNT=0x0600", "TAP_WIN_COMPONENT_ID=tap0901", "_CRT_SECURE_NO_WARNINGS", "ASIO_DISABLE_LOCAL_SOCKETS")
+        add_syslinks("fwpuclnt", "iphlpapi", "wininet", "setupapi", "rpcrt4", "wtsapi32", "advapi32", "ole32", "shell32", "ws2_32", "wsock32")
+    end
+    if is_plat("macosx") then
+        add_frameworks("CoreFoundation", "IOKit", "CoreServices", "SystemConfiguration")
+    end
+    if is_plat("linux") then
+        add_deps("libcap")
```
star-hengxing, line 30: It's better to separate the logic for `add_deps` and `add_syslinks`, that way the code will look a bit cleaner.
```lua
if is_plat("linux") then
    add_deps("libcap")
-- ...
end

if is_plat("macosx") then
    add_frameworks("CoreFoundation", "IOKit", "CoreServices", "SystemConfiguration")
-- ...
end
```

packages/o/openvpn3/xmake.lua
Redbeanw44602: 🤔 Every tag has a "release/" prefix. How can I avoid using alias?

SirLynix: as far as I know you can't in current xmake version, you should keep this as is.

PR #7863, opened 2025-08-11
librdkafka: fix & clean up

apocelipes:
Fixes #7699 
Fixes #7438
Fixes #7091
Fixes #6741
Fixes #5799 
Fixes #5528 
Fixes #5183 
Fixes #4642 
Fixes #4056 
Fixes #3377


packages/l/librdkafka/xmake.lua
```diff
@@ -40,6 +42,15 @@ package("librdkafka")
         end
     end)
 
+    if on_check then
+        on_check("wasm", function (package)
+            raise("package(librdkafka) does not support wasm!")
+        end)
+        on_check("bsd", function (package)
+            raise("package(librdkafka) dep(libcurl) does not support bsd!")
```
waruqi: why not use `on_install("!wasm and !bsd")`

apocelipes: I'd like to display different warning messages.

apocelipes: libcurl has already reported the message.

waruqi: if you use `use on_install("!wasm and !bsd")`, it will also report same errors.

Please use the same configuration method as other packages

apocelipes: @waruqi Yes, the new commit uses `on_install("!wasm and !bsd")` now. PTAL.

PR #7865, opened 2025-08-11
netcdf-c: add new package

choyy:


choyy:
Why in Linux(arm) and Linux hdf5 has no libfiles and links?
图片

choyy:
If I don't set the link order, in Linux(arm) hdf5 has wrong link order. It's strange that in Linux/Arch/Fedora the link order is correct. Do you know what's the problem? @star-hengxing 
图片

star-hengxing:
wait https://github.com/xmake-io/xmake-repo/pull/7911

PR #7877, opened 2025-08-12
mimalloc: update to v2.2.4

miredirex:


miredirex:
can you retrigger? or is the package itself broken? I see that these #7430 #6943 failed too

star-hengxing:
You can reference this [patch](https://github.com/microsoft/mimalloc/pull/400) to fix it...

miredirex:
5b07bbefb2f26bbe71b7de99064e4b66d8b661ab this is the only sane solution I found. For some reason even though arch is set to x86, CMake reads it as x64, [see here](https://github.com/xmake-io/xmake-repo/actions/runs/16921582281/job/47952734414#step:5:739). I don't really have time to investigate further

p.s. v3.1.5 now passes too

PR #7938, opened 2025-08-21
glgpuselect

ArthurVasseur:
GLGpuSelect is a drop‑in replacement for opengl32.dll (Windows) and — coming soon — libGL.so (Linux), enabling per‑application GPU routing for any OpenGL app.


ArthurVasseur:
Can you restart the jobs the ci fails with these errors:
```
The job was not acquired by Runner of type hosted even after multiple attempts
Internal server error. Correlation ID: dda13b29-a399-4423-9d1f-06f94e840c08</div>
```

PR #7950, opened 2025-08-23
Add packages: llgl and gaussianlib

zrurf:
* Add package: LLGL (version 0.04b)
* Add package: GaussianLib, which is the dependency of LLGL
* Known problem: LLGL does not have the on_test() function because when the on_test() function is executed, problems will occur when linking to the dxgi system library.


waruqi:
Only one package can be added to a PR; please split them.

zrurf:
Ok, I will close this PR and resubmit it.

PR #7953, opened 2025-08-23
Add package: LLGL

zrurf:
Add package: LLGL (github.com/LukasBanana/LLGL)
* Only the latest version v0.04b has been added.
* Known problem: LLGL does not have the on_test() function because when the on_test() function is executed, problems will occur when linking to the dxgi system library.

### DIFF
--- a/packages/l/llgl/xmake.lua
+++ b/packages/l/llgl/xmake.lua
@@ -138,8 +138,6 @@ endif()
         local opt = {}
         if package:is_plat("linux") then
             opt.packagedeps = {"wayland", "libxrandr", "libxrender"}
-        elseif package:is_plat("bsd", "cross") then
-            opt.packagedeps = {"libx11", "xorgproto"}
         end
         import("package.tools.cmake").install(package, configs, opt)
         if package:is_plat("android") then
@@ -158,3 +156,4 @@ endif()
             }
         ]]}, {configs = {languages = "c++11"}}))
     end)
+


packages/l/llgl/xmake.lua
```diff
@@ -0,0 +1,78 @@
...
+        if package:config("null") then
+            package:add("links", "LLGL_Null")
+        end
+    end)
+
+    on_install(function (package)
+        local dep_gaussian = package:dep("gaussianlib")
+
+        if dep_gaussian then
+            includedir = dep_gaussian:installdir("include")
```
waruqi: we shoud use dep_gaussian:fetch() to get includedirs

packages/l/llgl/xmake.lua
```diff
@@ -0,0 +1,78 @@
+package("llgl")
+    set_description("Low Level Graphics Library (LLGL) is a thin abstraction layer for the modern graphics APIs OpenGL, Direct3D, Vulkan, and Metal")
+    set_homepage("https://github.com/LukasBanana/LLGL")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/LukasBanana/LLGL/archive/refs/tags/Release-$(version).tar.gz",
+             "https://github.com/LukasBanana/LLGL.git")
+
+    add_versions("v0.04b", "fdeda39bd31522bced0d889655b290e06688975d58ab20756c3eda9a5f21391f")
```
waruqi: please use semver `v0.04` and `Release-$(version)b.tar.gz`

or use version function.

zrurf: This project has not yet released a stable version, so the version number carries a "b" suffix. I'm not sure if it will conflict with our current changes when the stable version is released.

PR #7963, opened 2025-08-24
itk: use external dependencies

choyy:


star-hengxing:
Why `add_deps("gtest")`, test should be disabled

choyy:
itk(shared) uses shared hdf5 as dependency, and another dependency libminc would use static hdf5. I can patch itk to use static hdf5, but is there a better solution? @star-hengxing

PR #7965, opened 2025-08-24
Add sqlgen package

Persson-dev:
I'm not really sure how cmake detect packages.
The patch for sqlite3 seems a bit hacky and cmake can't detect the mariadb package for some reason, this is why I commented it out.

### DIFF
--- a/packages/s/sqlgen/xmake.lua
+++ b/packages/s/sqlgen/xmake.lua
@@ -1,0 +1,97 @@
+package("sqlgen")
+    set_homepage("https://github.com/getml/sqlgen")
+    set_description("sqlgen is an ORM and SQL query generator for C++-20, similar to Python's SQLAlchemy/SQLModel or Rust's Diesel.")
+    set_license("MIT")
+
+    add_urls("https://github.com/getml/sqlgen/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/getml/sqlgen.git")
+
+    add_versions("v0.2.0", "c093036ebdf2aaf1003b2d1623713b97106ed43b1d39dc3d4f38e381f371799e")
+
+    add_patches("0.2.0", "patches/0.2.0/cmake.patch", "e9819b9a8a2c8f8a5b6c553eac3bb10fc65856aa9af451f83e2dbf55ca6c66c0")
+
+    add_deps("cmake", "reflect-cpp")
+
+    -- add_configs("mysql", {description = "Enable MySQL Support", default = false})
+    add_configs("postgres", {description = "Enable PostgreSQL Support", default = true})
+    add_configs("sqlite", {description = "Enable SQLite Support", default = true})
+
+    on_install(function (package)
+        local configs = {
+            "-DSQLGEN_USE_VCPKG=OFF",
+        }
+        table.insert(configs, "-DSQLGEN_MYSQL=" .. (package:config("mysql") and "ON" or "OFF"))
+        table.insert(configs, "-DSQLGEN_POSTGRES=" .. (package:config("postgres") and "ON" or "OFF"))
+        table.insert(configs, "-DSQLGEN_SQLITE3=" .. (package:config("sqlite") and "ON" or "OFF"))
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_load(function (package)
+        if package:config("mysql") then
+            package:add("deps", "mariadb-connector-c")
+        end
+        if package:config("postgres") then
+            package:add("deps", "libpq")
+        end
+        if package:config("sqlite") then
+            package:add("deps", "sqlite3")
+        end
+    end)
+
+    on_test(function (package)
+        if package:config("postgres") then
+            assert(package:check_cxxsnippets({test = [[
+                #include <sqlgen/postgres.hpp>
+
+                // PostgreSQL connection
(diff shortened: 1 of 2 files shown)


Persson-dev:
shared linking is broken on windows for some reasons...

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,98 @@
...
+
+    -- add_configs("mysql", {description = "Enable MySQL Support", default = false})
+    add_configs("postgres", {description = "Enable PostgreSQL Support", default = true})
+    add_configs("sqlite", {description = "Enable SQLite Support", default = true})
+
+    on_install("windows|!arm64 or macosx|!arm64 or linux|!arm64 or bsd|!arm64", function (package)
+        local configs = {
+            "-DSQLGEN_USE_VCPKG=OFF",
+        }
+        table.insert(configs, "-DSQLGEN_MYSQL=" .. (package:config("mysql") and "ON" or "OFF"))
```
waruqi: mysql config is not defined

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,98 @@
...
+        table.insert(configs, "-DSQLGEN_POSTGRES=" .. (package:config("postgres") and "ON" or "OFF"))
+        table.insert(configs, "-DSQLGEN_SQLITE3=" .. (package:config("sqlite") and "ON" or "OFF"))
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        -- allow only static linking on Windows
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. ((package:config("shared") and not is_plat("windows")) and "ON" or "OFF"))
```
waruqi: you should use add_configs + readonly to limit it.

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,98 @@
...
+            assert(package:check_cxxsnippets({test = [[
+                #include <sqlgen/sqlite.hpp>
+
+                struct User {
+                    std::string name;
+                    int age;
+                };
+
+                int main() {
+                    // Connect to SQLite database
```
waruqi: remove all comments

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,98 @@
...
+        table.insert(configs, "-DSQLGEN_MYSQL=" .. (package:config("mysql") and "ON" or "OFF"))
+        table.insert(configs, "-DSQLGEN_POSTGRES=" .. (package:config("postgres") and "ON" or "OFF"))
+        table.insert(configs, "-DSQLGEN_SQLITE3=" .. (package:config("sqlite") and "ON" or "OFF"))
+        table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
+        -- allow only static linking on Windows
+        table.insert(configs, "-DBUILD_SHARED_LIBS=" .. ((package:config("shared") and not is_plat("windows")) and "ON" or "OFF"))
+        import("package.tools.cmake").install(package, configs)
+    end)
+
+    on_load(function (package)
```
waruqi, line 23: move on_load before on_install

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,94 @@
...
+        end
+        if package:config("postgres") then
+            package:add("deps", "libpq")
+        end
+        if package:config("sqlite") then
+            package:add("deps", "sqlite3")
+        end
+    end)
+
+    on_install("windows|!arm64 or macosx|!arm64 or linux|!arm64 or bsd|!arm64", function (package)
```
waruqi: you can use on_check to simplify it, we can check arm64 in it.

```lua
on_check(function (package)
    assert(not package:is_arch("arm64"), "package(%s) does not support arm64", package:name())
end)

on_install("windows", "macosx", "linux", "bsd", ...)
```

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,94 @@
...
+                #include <sqlgen/sqlite.hpp>
+
+                struct User {
+                    std::string name;
+                    int age;
+                };
+
+                int main() {
+                    const auto conn = sqlgen::sqlite::connect("test.db");
+
```
waruqi: remove this empty line.

packages/s/sqlgen/xmake.lua
```diff
@@ -0,0 +1,94 @@
...
+        if package:config("sqlite") then
+            assert(package:check_cxxsnippets({test = [[
+                #include <sqlgen/sqlite.hpp>
+
+                struct User {
+                    std::string name;
+                    int age;
+                };
+
+                int main() {
```
waruqi: void test() {

packages/s/sqlgen/xmake.lua
Persson-dev: Since the restriction comes from the libpq dependency (from which I copied this line), is it possible to just tell xmake to check for libpq availability instead (without failing the builds)?

waruqi: you should check `package:config("postgres")` in on_check

Persson-dev: I'm not sure to understand, like that ?
```xmake
on_check(function (package)
    if package:config("postgres") then
        package:add("deps", "libpq")
    end
end)
 ```
 It does not seem to work

Comment (waruqi) [reply packages/s/sqlgen/xmake.lua]:
```lua
on_check(function (package)
    if package:config("postgres") then
        assert(not package:is_arch("arm64"), "package(%s) does not support arm64", package:name())
    end
end)
```

Persson-dev: Right, makes sense :sweat_smile:

PR #7975, opened 2025-08-25
libbpf: try to fix build

apocelipes:
High version of libbpf needs elfutils.

libbpf has no official support for Android, and it uses functions that are disabled on Android. So disable Android installations.

Fixes #7967 
Fixes #7915
Fixes #7199
Fixes #6057 
Fixes #5653 
Fixes #5446 
Fixes #4652 
Fixes #4374 
Fixes #3950 
Fixes #3794 
Fixes #3378

### DIFF
--- a/packages/l/libbpf/xmake.lua
+++ b/packages/l/libbpf/xmake.lua
@@ -1,31 +1,38 @@
 package("libbpf")
-
     set_homepage("https://github.com/libbpf/libbpf")
     set_description("Automated upstream mirror for libbpf stand-alone build.")
+    set_license("BSD-2-Clause")
 
     add_urls("https://github.com/libbpf/libbpf/archive/refs/tags/$(version).tar.gz",
              "https://github.com/libbpf/libbpf.git")
+    add_versions("v1.6.2", "16f31349c70764cba8e0fad3725cc9f52f6cf952554326aa0229daaa21ef4fbd")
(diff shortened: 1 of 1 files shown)


packages/l/libbpf/xmake.lua
```diff
@@ -1,31 +1,38 @@
...
+        if package:version() and package:version():lt("0.5") then
+            package:add("deps", "libelf")
+        else
+            package:add("deps", "elfutils")
+        end
+    end)
 
     add_includedirs("include", "include/uapi")
 
-    on_install("linux", "android", function (package)
```
waruqi, line 14: it will break xmake/bpf tests and https://github.com/libbpf/libbpf-bootstrap

https://github.com/xmake-io/xmake/actions/runs/17227312910/job/48874434222?pr=6324git

we need to support for android

waruqi:
I have reverted it first. https://github.com/xmake-io/xmake-repo/pull/7979

packages/l/libbpf/xmake.lua
apocelipes: OK, I'll send another PR to enable Android building. However, libbpf needs NDK >= r23 (or maybe newer).

waruqi: https://github.com/libbpf/libbpf-bootstrap/blob/52e97e567b7136cca1fb271be8bf5c4936947eff/examples/c/xmake.lua#L16-L17

PR #7980, opened 2025-08-26
libbpf: add back android support

apocelipes:
Based on #7975, but added Android support back.

Android needs NDK SDK version >= 23.

### DIFF
--- a/packages/l/libbpf/xmake.lua
+++ b/packages/l/libbpf/xmake.lua
@@ -16,7 +16,11 @@ package("libbpf")
             local ndk_sdkver = ndk:config("ndk_sdkver")
             local ndkver = ndk:config("ndkver")
             assert(ndkver and tonumber(ndkver) < 26, "package(libbpf): need ndk version < 26 for android")
-            assert(ndk_sdkver and tonumber(ndk_sdkver) == 23, "package(libbpf): need ndk api level == 23 for android")
+            if package:version() and package:version():lt("0.5") then
+                assert(ndk_sdkver and tonumber(ndk_sdkver) >= 23, "package(libbpf) dep(libelf): need ndk api level >= 23 for android")
+            else
+                assert(ndk_sdkver and tonumber(ndk_sdkver) == 23, "package(libbpf) dep(elfutils): need ndk api level == 23 for android")
+            end
         end)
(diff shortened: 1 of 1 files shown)


apocelipes:
The existing Android CI has not been able to perform effective testing because elfutils requires ndk sdk <= 23, while libbpf requires >= 23 (from v0.1), so only v23 could be used, but our CI only has sdk version 21 and 30.

apocelipes:
@waruqi PTAL.

packages/l/libbpf/xmake.lua
```diff
@@ -1,31 +1,55 @@
...
+                assert(ndk_sdkver and tonumber(ndk_sdkver) == 23, "package(libbpf) dep(elfutils): need ndk api level == 23 for android")
+            end
+        end)
+    end
+
+    on_load(function (package)
+        if package:version() and package:version():lt("0.5") then
+            package:add("deps", "libelf")
+        else
+            package:add("deps", "elfutils")
```
waruqi, line 31: elfutils only support ndk < r26 now, if use ndk r27/28, it will break.

https://github.com/xmake-io/xmake-repo/blob/972d58a30cdc1f5ecf08bd61121f2716f7cfa888/packages/e/elfutils/xmake.lua#L46 

but latest ndk is r27 in xmake-repo

apocelipes: libbpf can do nothing here. I think elfutils should be update.

apocelipes: NDK > 26 leads to this issue: libbpf-bootstrap/288
I found the version limitation was introduced by #5307.

waruqi: > libbpf can do nothing here. I think elfutils should be update.

Yes, elfutils needs to be updated to support the new ndk, but it will break some existing projects now.

apocelipes: OK, I'll try to fix elfutils firstly.

apocelipes: If this is blocking some important work progress, you can revert it. I'll re-send it after elfutils can be compiled with NDK r27+.

PR #7993, opened 2025-08-28
Libpq: add arm support

Persson-dev:


Persson-dev:
The failing windows job should succeed if it's rerun, it has nothing to do with the package

PR #8015, opened 2025-08-31
sdl3webgpu: add package

e2fo2l:
### An extension for the SDL3 library for using WebGPU native.
https://github.com/eliemichel/sdl3webgpu

Based on the xmake.lua file for glfw3webgpu.

### DIFF
--- a/packages/s/sdl3webgpu/xmake.lua
+++ b/packages/s/sdl3webgpu/xmake.lua
@@ -3,10 +3,10 @@ package("sdl3webgpu")
     set_homepage("https://github.com/eliemichel/sdl3webgpu")
     set_license("MIT")
 
-    add_urls("https://github.com/eliemichel/sdl3webgpu/archive/refs/tags/$(version).tar.gz",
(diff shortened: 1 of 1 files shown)


e2fo2l:
It seems that like for sdl2webgpu, v1.0.0 is based on an older wgpu-native version, it needs to be fetched from a more recent commit

packages/s/sdl3webgpu/xmake.lua
```diff
@@ -0,0 +1,55 @@
...
+
+    on_install("windows|x64", "windows|x86", "linux|x86_64", "macosx|x86_64", "macosx|arm64", function (package)
+        if package:is_plat("macosx", "iphoneos") then
+            os.mv("sdl3webgpu.c", "sdl3webgpu.m")
+        end
+
+        io.writefile("xmake.lua", [[
+            add_rules("mode.debug", "mode.release")
+
+            add_requires("wgpu-native", "libsdl3")
```
waruqi, line 25: you need to add `add_deps` in package configuraion.

PR #8023, opened 2025-09-01
expat: fix cmake

choyy:


star-hengxing:
`lib/cmake/expat-2.7.1/expat.cmake`
```cmake
set_target_properties(expat::expat PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "XML_STATIC"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```
After my testing, only when using `find_package(XXX CONFIG REQUIRED)` and cmake read these macros, so we only need to patch find_package.

Issue #10, opened 2019-03-29
Collect C/C++ libraries


waruqi:

waruqi:
#1 ffmpeg

wutiantong:
https://github.com/nlohmann/json

https://github.com/onqtam/doctest

waruqi:
> https://github.com/nlohmann/json

This package has been collected, you can see: https://github.com/xmake-io/xmake-repo/tree/master/packages/n/nlohmann_json

lineCode:
https://www.cairographics.org/
https://www.pango.org/

waruqi:
> https://github.com/onqtam/doctest

@wutiantong  done

https://github.com/xmake-io/xmake-repo/tree/master/packages/d/doctest

nikhedonia:
I hope this is helpful:

Here is a list of [Buckaroo](https://github.com/loopperfect/buckaroo)'s officially supported libraries:
https://github.com/buckaroo-pm

And more are on our wishlist:
https://github.com/LoopPerfect/buckaroo-wishlist/issues

waruqi:
> I hope this is helpful:
> 
> [Here](https://github.com/buckaroo-pm) is a list of [Buckaroo](https://github.com/loopperfect/buckaroo)'s officially supported libraries and then there is more on our [wishlist](https://github.com/LoopPerfect/buckaroo-wishlist/issues).

Ok, I will look at it. 

Thanks!

Berrysoft:
[Here](https://github.com/Microsoft/vcpkg/tree/master/ports) are the libraries support by [vcpkg](https://github.com/Microsoft/vcpkg), maybe it's helpful.

kassane:
[Nana Library](https://github.com/cnjinhao/nana)

waruqi:
> [Nana Library](https://github.com/cnjinhao/nana)

Ok, I will add it.

waruqi:
> https://www.cairographics.org/

@lineCode done

```lua
add_requires("cairo", "pixman")
```

Support linux, macOS and windows.

kassane:
[Cutelyst](https://github.com/cutelyst/cutelyst)

waruqi:
Add boost library. for example:

```lua
add_requires("boost")
target("test")
    set_kind("binary")
    set_languages("c++14")
    add_files("src/*.cpp")
    add_packages("boost")
```

Only add the given boost library.

```lua
[...]

PR #8031, opened 2025-09-02
gte: add new package

luadebug:
https://github.com/xmake-io/xmake-repo/issues/10#issuecomment-3205502089


luadebug:
Theoretically it's fine to merge as another Linux runners might miss GL/EGL installed?

PR #8039, opened 2025-09-03
hiredis-vip: add new package

jiaobendaye:


packages/h/hiredis-vip/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+    add_urls("https://github.com/vipshop/hiredis-vip/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/vipshop/hiredis-vip.git")
+
+    add_versions("0.3.0", "84e0f9367fa25089fc073b7a8a0725043c48cccec827acf4555a63da68f36be5")
+
+    add_configs("shared", {description = "Build shared library.", default = true, type = "boolean"})
+
+    on_install(function (package)
+        local configs = {}
+        table.insert(configs, "PREFIX="..package:installdir())
```
waruqi: add spaces. ` .. `

packages/h/hiredis-vip/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+        if not package:config("debug") or package:is_mode("release") then
+            table.insert(configs, "DEBUG=")
+        end
+        os.vrunv("make install ", configs)
```
waruqi: you should use `package.tools.make` to install it.

packages/h/hiredis-vip/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+        os.cp(path.join(os.curdir(), "net.h"), package:installdir("include/hiredis-vip"))
+    end)
+
+    on_test(function (package)
+          assert(package:has_cfuncs("redisCommand", {includes = "hiredis-vip/hiredis.h"}))
```
waruqi: incorrect indent

packages/h/hiredis-vip/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+    set_homepage("https://github.com/vipshop/hiredis-vip")
+    set_description("Support redis cluster.  Maintained and used at vipshop.")
+    set_license("BSD-3-Clause")
+
+    add_urls("https://github.com/vipshop/hiredis-vip/archive/refs/tags/$(version).tar.gz",
+             "https://github.com/vipshop/hiredis-vip.git")
+
+    add_versions("0.3.0", "84e0f9367fa25089fc073b7a8a0725043c48cccec827acf4555a63da68f36be5")
+
+    add_configs("shared", {description = "Build shared library.", default = true, type = "boolean"})
```
waruqi: if it only support shared, you should set readonly.

packages/h/hiredis-vip/xmake.lua
```diff
@@ -0,0 +1,25 @@
...
+
+    on_install(function (package)
```
waruqi: please limit platforms.

packages/h/hiredis-vip/xmake.lua
jiaobendaye: 如果用 package.tools.make的话，发现configs只会在make时生效，第二步执行 make install时就没了，所以直接用的os.vrunv

packages/h/hiredis-vip/xmake.lua
jiaobendaye: also support static

packages/h/hiredis-vip/xmake.lua
luadebug: I think at this case we have to follow existing recipe structure. That is why last lines of recipes should use tool not vrun.

packages/h/hiredis-vip/xmake.lua
luadebug: If it supports static as well we do not `add_config` for this, since both cases re supported.

packages/h/hiredis-vip/xmake.lua
luadebug: It is exact `!windows and !mingw` ?

packages/h/hiredis-vip/xmake.lua
luadebug: Please keep indent 4 spaces.

packages/h/hiredis-vip/xmake.lua
luadebug: Please add spaces between `..` as to follow existing recipe style.

PR #8053, opened 2025-09-05
pocketpy: update to v2.1.1

Willaaaaaaa:


packages/p/pocketpy/xmake.lua
```diff
@@ -5,33 +5,84 @@ package("pocketpy")
...
+                    py_newint(py_list_getitem(r0, 1), 2);
+                    py_newint(py_list_getitem(r0, 2), 3);
+                    
+                    py_finalize();
+                    return;
+                
+                __ERROR:  
+                    py_printexc();
+                    py_finalize();
+                }
```
waruqi, line 57: please simplify test snippets, remove duplicate code

Willaaaaaaa:
And I'm wondering whether it is a good practice to use `add_resources` to get the c file. Since we can't auto-update this c file's hash.

PR #8072, opened 2025-09-07
cpp-httplib: Update to v0.26.0

Willaaaaaaa:
- add checks for winver, vs_ver
- ban cygwin according to readme
- fix the patch
- remove duplicate Android ndk check

TODO: cmake error

### DIFF
--- a/packages/c/cpp-httplib/patches/v0.26.0/fix-mingw.diff
+++ b/packages/c/cpp-httplib/patches/v0.26.0/fix-mingw.diff
@@ -1,20 +1,21 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index d39958a..c14c624 100644
+index d39958a..ba21f3c 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -117,8 +117,13 @@ if(BUILD_SHARED_LIBS AND WIN32 AND HTTPLIB_COMPILE)
+@@ -117,8 +117,14 @@ if(BUILD_SHARED_LIBS AND WIN32 AND HTTPLIB_COMPILE)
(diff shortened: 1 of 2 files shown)


Willaaaaaaa:
using:
```cmake
message(STATUS "CMAKE_SYSTEM_NAME = <${CMAKE_SYSTEM_NAME}>")
message(STATUS "CMAKE_SYSTEM_VERSION = <${CMAKE_SYSTEM_VERSION}>")

if(CMAKE_SYSTEM_NAME MATCHES "Windows")
	if(${CMAKE_SYSTEM_VERSION} VERSION_LESS "10.0.0")
		message(SEND_ERROR "Windows ${CMAKE_SYSTEM_VERSION} or lower is not supported. Please use Windows 10 or later.")
	endif()
endif()
```
get: https://github.com/xmake-io/xmake-repo/actions/runs/17525052896/job/49774329556?pr=8072#step:5:1224

---

When cross-compiling, `CMAKE_SYSTEM_VERSION` should be specified along with `CMAKE_SYSTEM_NAME`. Otherwise `CMAKE_SYSTEM_VERSION` will be empty (that is `if( VERSION_LESS "10.0.0")`), thus leading to errors.

ref: https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_VERSION.html#system-version-for-cross-compiling

Willaaaaaaa:
https://github.com/yhirose/cpp-httplib/blob/89c932f313c6437c38f2982869beacc89c2f2246/README.md?plain=1#L1274
Should we support mingw? Although it passed CI, it's not offically supported.

waruqi:
If it works fine, it would be better to support it.

PR #8092, opened 2025-09-11
openssh: add package.

Redbeanw44602:


packages/o/openssh/xmake.lua
```diff
@@ -0,0 +1,170 @@
...
+        if package:config("kerberos5") then
+            package:add("deps", "krb5")
+        end
+
+        if package:config("privsep_path") == nil then
+            package:config_set("privsep_path", package:installdir("var/empty"):gsub("\\", "/"))
+        end
+    end)
+
+    on_install("linux", "bsd", "macosx", "msys", "cygwin", function (package)
```
Redbeanw44602: Question: This seems to cause the `MingW (Msys2)` tests to not run, and my intention is only allow mingw on msys to build this package. Can anyone help? 🤔

waruqi: "mingw@msys"

luadebug: > MingW (Msys2) tests to not run

Are you sure that MSYS tests exists at this Github CI workflow, as I saw there is only MinGW Github CI workflow.

> "mingw@msys"

What is difference between `"mingw@msys"` and just `"msys"`? I think they are same, but `"cygwin"` is not `mingw on msys`.

Redbeanw44602: @luadebug 
> Are you sure that MSYS tests exists at this Github CI workflow?

(This workflow) uses mingw via msys2 on windows.
>  What is difference between "mingw@msys" and just "msys"? I think they are same.

I'm confused  about this as well, I thought they should be the same. 😕

---

@waruqi 

I actually tried this pattern but it results in another error:
```
note: the following packages are unsupported on msys/x86_64:
  -> openssh 10.0p2 [host, license:BSD-2-Clause]
```
Links: https://github.com/Redbeanw44602/xmake-repo-testing/pull/42/commits/78dfb4270269c2d6c74fefb0b8ebc14477fc3490 ([actions](https://github.com/Redbeanw44602/xmake-repo-testing/actions/runs/17641682884/job/50129986547))

waruqi: - mingw: mingw is only target platform. we can build mingw program on macOS/Linux or Msys2 or Windows cmd, powershell...
- mingw@msys: Build mingw program on msys2
- mingw@windows: Build mingw program on Windows cmd,powershell,.. (non-msys2 shell)
- mingw@macosx: Build mingw program on macOS
- msys: this is a target platform, we can use the default GNU GCC (not mingw/gcc) to build program on Msys2. it is equal to `msys@msys`
- cygwin: this is target platform. we can use the default GNU GCC to build program on Cygwin. it is equal to `cygwin@cygwin`
- mingw@cygwin: Build mingw program on Cygwin.

waruqi: "msys" uses GNU gcc to build programs by default, instead of mingw/gcc. GNU gcc does not use the Windows API directly, but uses the ported POSIX API. e.g. fork()/exec() instead of CreateProcess (mingw)

"mingw@msys" will switch to mingw platform and use mingw toolchain to build program on Msys2.

Redbeanw44602: Thanks for the clear explanation, I was confusing mingw-gcc with gnu-gcc. 😂

waruqi: > @luadebug
> 
> > Are you sure that MSYS tests exists at this Github CI workflow?
> 
> (This workflow) uses mingw via msys2 on windows.
> 
> > What is difference between "mingw@msys" and just "msys"? I think they are same.
> 
> I'm confused about this as well, I thought they should be the same. 😕
> 
> @waruqi
> 
> I actually tried this pattern but it results in another error:
> 
> ```
> note: the following packages are unsupported on msys/x86_64:
>   -> openssh 10.0p2 [host, license:BSD-2-Clause]
> ```
> 
> Links: [Redbeanw44602/xmake-repo-testing@78dfb42](https://github.com/Redbeanw44602/xmake-repo-testing/commit/78dfb4270269c2d6c74fefb0b8ebc14477fc3490) ([actions](https://github.com/Redbeanw44602/xmake-repo-testing/actions/runs/17641682884/job/50129986547))

because you configured `set_kind("binary")`, this package is a binary package.

it only can run on host machine instead of target platform.

so you should use `on_install("@linux", "@bsd", "@macosx", "@msys", "@cygwin")`

we can only use GNU/gcc to build the host program on msys2 now.

so you can remove msys support.

`on_install("@linux", "@bsd", "@macosx", "@cygwin")`

Redbeanw44602: I will comment out msys support for now🙂

Redbeanw44602:
> If you remove msys2 and kept just cygwin support, would your patches be really useful here at context of cygwin, as it looks like MSYS2 backport patches that does not look like respect cygwin ?

Hmm you are right, maybe I should avoid applying patches to mingw.

packages/o/openssh/xmake.lua
```diff
@@ -0,0 +1,172 @@
...
+    add_configs("ipaddr_display",          {description = "Use ip address instead of hostname in $DISPLAY.", type = "boolean", default = nil})
+    add_configs("default_path",            {description = "Specify default $PATH environment for server.", type = "string", default = nil})
+    add_configs("superuser_path",          {description = "Specify different path for super-user.", type = "string", default = nil})
+    add_configs("ip4in6",                  {description = "Check for and convert IPv4 in IPv6 mapped addresses.", type = "boolean", default = nil})
+    add_configs("bsd_auth",                {description = "Enable BSD auth support.", type = "boolean", default = nil})
+    add_configs("pid_dir",                 {description = "Specify location of sshd.pid file.", type = "string", default = nil})
+    add_configs("lastlog_dir",             {description = "Specify lastlog location common locations.", type = "string", default = nil})
+
+    -- patches from: https://github.com/msys2/MSYS2-packages/tree/master/openssh
+    if is_plat("msys") then
```
waruqi: description/is_plat is only for target platform. but it's binary package.

you should use `target:is_plat` and `target:add("deps")`, `target:add("patches")` in on_load.

waruqi: or you should use `if os.subhost() == "msys" then`

PR #8109, opened 2025-09-14
fmt: fix clang-cl with exceptions

star-hengxing:


packages/f/fmt/xmake.lua
```diff
@@ -76,7 +77,18 @@ package("fmt")
...
+
+        local opt = {}
+        if package:config("exceptions") then
+            opt.cxflags = {
+                "-DFMT_USE_EXCEPTIONS", -- renamed in 11.1
+                "-DFMT_EXCEPTIONS",
+            }
+            if package:has_tool("cxx", "clang_cl") then
+                table.insert(opt.cxflags, "/EHsc")
+            end
```
DownerCase, line 89: Genuine question. I've seen a few PRs like these now. Why does only clang-cl need the flag and not msvc as well?

star-hengxing: msvc && clang-cl disable exceptions by default.
cmake always pass `/EHsc` for msvc by default, but clang-cl dose not.

DownerCase: Sorry for derailing the PR but...
Is this affected by how XMake does things? I just tested directly with CMake 4.1 and a minimal CMake project and both cl and clang-cl produced the same CMAKE_CXX_FLAGS with /EHsc

I.e here are relevant values from cmake cache
cl:
```
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_CXX_COMPILER:STRING=Z:/vs2022/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe
CMAKE_CXX_FLAGS:STRING=/DWIN32 /D_WINDOWS /EHsc
CMAKE_CXX_FLAGS_DEBUG:STRING=/Zi /Ob0 /Od /RTC1
```

clang-cl
```
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_CXX_COMPILER:STRING=Z:/vs2022/VC/Tools/Llvm/x64/bin/clang-cl.exe
CMAKE_CXX_FLAGS:STRING=/DWIN32 /D_WINDOWS /EHsc
CMAKE_CXX_FLAGS_DEBUG:STRING=/Zi /Ob0 /Od /RTC1
```

The CMake project:
```
cmake_minimum_required(VERSION 3.20)
project(minimal LANGUAGES CXX)
```

Configured with:
```
cmake -S . -B build-clang-cl -DCMAKE_CXX_COMPILER=clang-cl --fresh
cmake -S . -B build-cl -DCMAKE_CXX_COMPILER=cl --fresh
```

I wonder if what you're actually seeing is  a result of this bug: https://github.com/xmake-io/xmake/issues/6781 where the default flags (including `/EHsc`) in certain situations gets overwritten with xmake/user specified compile flags.

star-hengxing: Thanks your feeback, Seem like xmake bug, I will try to fix it.

PR #8136, opened 2025-09-17
Fix rustup package on Windows

SirLynix:
https://github.com/xmake-io/xmake/commit/8be13f815541675a1592f58505c047e0f27b3a8e broke rustup package because `rustup-init` relies on its filename.

```
D:\a\ShaderLang\xmake-global\.xmake\cache\packages\2509\r\rustup\1.28.1\rustup-1.28.1 --no-modify-path --no-update-default-toolchain --profile=minimal --default-toolchain=none -y
error: unknown proxy name: 'rustup-1.28'; valid proxy names are 'rustc', 'rustdoc', 'cargo', 'rust-lldb', 'rust-gdb', 'rust-gdbgui', 'rls', 'cargo-clippy', 'clippy-driver', 'cargo-miri', 'rust-analyzer', 'rustfmt', 'cargo-fmt'
error: .\.xmake-cache-W37\core\sandbox\modules\os.lua:378: execv(D:\a\ShaderLang\xmake-global\.xmake\cache\packages\2509\r\rustup\1.28.1\rustup-1.28.1 --no-modify-path --no-update-default-toolchain --profile=minimal --default-toolchain=none -y) failed(1)
```

### DIFF
--- a/packages/r/rustup/xmake.lua
+++ b/packages/r/rustup/xmake.lua
@@ -5,20 +5,20 @@ package("rustup")
 
     if is_host("windows") then
         if os.arch() == "x64" then
-            add_urls("https://static.rust-lang.org/rustup/archive/$(version)/x86_64-pc-windows-msvc/rustup-init.exe")
+            add_urls("https://static.rust-lang.org/rustup/archive/$(version)/x86_64-pc-windows-msvc/rustup-init.exe", {filename = "rustup-init.exe"})
 
             add_versions("1.28.1", "7b83039a1b9305b0c50f23b2e2f03319b8d7859b28106e49ba82c06d81289df6")
         elseif os.arch() == "x86" then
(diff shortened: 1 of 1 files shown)


star-hengxing:
I think we can only rename tarball(.tar/.zip) @waruqi

waruqi:
https://github.com/xmake-io/xmake/pull/6835

SirLynix:
> [xmake-io/xmake#6835](https://github.com/xmake-io/xmake/pull/6835)

note that in this case it's not about the extension, the executable runs but it fetches its own name because its the same binary for multiple executables (same as unzip iirc)

waruqi:
> > [xmake-io/xmake#6835](https://github.com/xmake-io/xmake/pull/6835)
> 
> note that in this case it's not about the extension, the executable runs but it fetches its own name because its the same binary for multiple executables (same as unzip iirc)

This is only the case for special packages, so I still merged the patch.

But the extension also has problems and needs to be fixed.

PR #8143, opened 2025-09-17
Upgrade `buildir` to `builddir` on `libzip`

PerikiyoXD:
Fixes warning:
> [!WARNING]
>  package:buildir() has been deprecated, please use package:builddir()


qudix:
What happens when someone using an older version of xmake before the name change tries to use the package?

SirLynix:
we could check if `package.builddir` exists first
```lua
local builddir = package.builddir and package:builddir() or package:buildir()
```

star-hengxing:
Copy pdb just workaround code, fixed https://github.com/xmake-io/xmake/issues/5566

PR #8151, opened 2025-09-18
wayland: update to 1.24.0

luadebug:
Place fossies mirror first in list, since `https://gitlab.freedesktop.org` is protected by anubis anti-DDoS, which could filter our request with higher possibility rather mirror url.


luadebug:
Workaround is to update wayland, since fossies remove 1.23...

PR #8167, opened 2025-09-20
libeconf: add package.

Redbeanw44602:


Redbeanw44602:
The package fails to build under Linux (Clang) due to:
```
bfd plugin: LLVM gold plugin has failed to create LTO module: Invalid attribute group entry (Producer: 'LLVM20.1.8' Reader: 'LLVM 18.1.3')
```

Multiple versions of LLVM are installed in the ubuntu runner, but AR appears to be selecting an older version of `LLVMgold.so`, I've temporarily removed the other versions of LLVM. 

I think this might be an xmake bug. (should choose `llvm-ar-20` instead of `ar`)

waruqi:
> The package fails to build under Linux (Clang) due to:
> 
> ```
> bfd plugin: LLVM gold plugin has failed to create LTO module: Invalid attribute group entry (Producer: 'LLVM20.1.8' Reader: 'LLVM 18.1.3')
> ```
> 
> Multiple versions of LLVM are installed in the ubuntu runner, but AR appears to be selecting an older version of `LLVMgold.so`, I've temporarily removed the other versions of LLVM.
> 
> I think this might be an xmake bug. (should choose `llvm-ar-20` instead of `ar`)

I don't think this is a problem with xmake, xmake always uses the default compiler tools

clang toolchain: clang/clang++, ar
llvm toolchain: clang/clang, llvm-ar



> * Linux (Clang) / build (ubuntu-latest, shared) (pull_request)

try it again. 

https://github.com/xmake-io/xmake/commit/f9bc475a5ba47b954d23d16c71a4b6bbd4cb6ca4

waruqi:
ci 配置不要改

PR #8168, opened 2025-09-20
libseccomp: add package.

Redbeanw44602:


packages/l/libseccomp/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("libseccomp")
+    set_description("The libseccomp library provides an easy to use, platform independent, interface to the Linux Kernel's syscall filtering mechanism.")
+
```
star-hengxing, line 5: Missing License

packages/l/libseccomp/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+    add_urls("https://github.com/seccomp/libseccomp/releases/download/v$(version)/libseccomp-$(version).tar.gz")
```
star-hengxing: Missing git url

packages/l/libseccomp/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+    add_versions("2.6.0", "83b6085232d1588c379dc9b9cae47bb37407cf262e6e74993c61ba72d2a784dc")
+
+    add_configs("tools", {description = "Build tools.", default = true, type = "boolean"})
+
+    add_deps("gperf")
+    on_install("linux", function (package)
+        local configs = {
+            "--disable-dependency-tracking",
+            "--disable-python"
+        }
```
waruqi, line 23: we need to config shared/static libraries.

packages/l/libseccomp/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("libseccomp")
+    set_description("The libseccomp library provides an easy to use, platform independent, interface to the Linux Kernel's syscall filtering mechanism.")
+
+    add_urls("https://github.com/seccomp/libseccomp/releases/download/v$(version)/libseccomp-$(version).tar.gz")
+    add_versions("2.6.0", "83b6085232d1588c379dc9b9cae47bb37407cf262e6e74993c61ba72d2a784dc")
+
+    add_configs("tools", {description = "Build tools.", default = true, type = "boolean"})
```
star-hengxing: Another package require tools? if no, please set false by default

waruqi: we can always disable tools. this package is only a library

packages/l/libseccomp/xmake.lua
Redbeanw44602: Why? xmake seems to have done this already:

https://github.com/xmake-io/xmake/blob/ff52d9da8840c689867a915e0971a8a03678e8ad/xmake/modules/package/tools/autoconf.lua#L199

waruqi: ok

PR #8191, opened 2025-09-22
Update README and CONTRIBUTING

black-desk:
correct links


black-desk:
Perhaps we should link to https://xmake.io/guide/package-management/package-distribution.html#package-distribution?

PR #8165, opened 2025-09-20
libkmod: fix pkg-config and upgrade to v33.

Redbeanw44602:
## about pkg-config fix
The original package incorrectly uses `v31` as the version number (it should be `31`), which will cause other projects that use this package through pkg-config to be unable to correctly compare the version number.

PR #8113, opened 2025-09-15
libxcrypt: add package.

Redbeanw44602:

PR #8238, opened 2025-09-26
libsystemd: add package.

Redbeanw44602:
### About pruning meson.build
Since other pakcages are unlikely to have dependencies on systemd tools, so we only need to build libsystemd/libudev. `meson.build` is stripped to maintain a clean installdir and improve build speed.

### Dependency Tree
 - systemd
   - #8113 
   - #8114 
   - #8165 
   - #8166 
   - #8169 
   - #8168

### DIFF
--- a/packages/l/libsystemd/xmake.lua
+++ b/packages/l/libsystemd/xmake.lua
@@ -1,0 +1,146 @@
+package("libsystemd")
+    set_homepage("https://systemd.io")
+    set_description("The systemd System and Service Manager.")
+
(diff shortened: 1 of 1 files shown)


waruqi:
please remove this. https://github.com/xmake-io/xmake-repo/blob/a2ff81e34b75e239b50e905bcd19e75c86fc52f3/scripts/test.lua#L294

Redbeanw44602:
Oh, it's been merged. I'll remove it in #8230

PR #8257, opened 2025-09-28
fix(doctest): Copy headers to the include directory

24bit-xjkp:
1. Some packages like `FakeIt` use <doctest.h>, so we need a copy of header files in the include dir

Hear is an example in [FakeIt](https://github.com/eranpeer/FakeIt/blob/1d172eb87929b8bd5a8dd9a981fc0e43b9245475/single_header/doctest/fakeit.hpp#L1164) package. It use `<doctest.h>` directly, but the doctest package install `<doctest.h>` in the `include/doctest` directory, so a copy of `<doctest.h>` in `include` directory is nedded.

### DIFF
--- a/packages/d/doctest/xmake.lua
+++ b/packages/d/doctest/xmake.lua
@@ -1,5 +1,4 @@
 package("doctest")
-
     set_kind("library", {headeronly = true})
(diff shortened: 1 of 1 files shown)


star-hengxing:
We can add `cmake install` for config file installation.

PR #8258, opened 2025-09-28
boost: fix 1.85.0 when using vs2026

Willaaaaaaa:
Fix patch to support vs2026

VCToolsVersion = "14.50.35503"

### DIFF
--- a/packages/b/boost/b2/load.lua
+++ b/packages/b/boost/b2/load.lua
@@ -90,7 +90,7 @@ function main(package)
             local vs_toolset_ver = import("core.base.semver").new(vs_toolset)
             local minor = vs_toolset_ver:minor()
             if minor and minor >= 44 then
-                package:add("patches", "<=1.85.0", "patches/1.85.0/fix-msvc.patch", "6457f110fc844c9ee019c66419ba6697a8864ce79d3ff5c413d36fb734fb51a9")
+                package:add("patches", "<=1.85.0", "patches/1.85.0/fix-msvc.patch", "e3c9fcfa424581c385e0346bd747f38d83fe8dcda9a5a9b87fb11796209c52ca")
             end
         end
     end


packages/b/boost/b2/load.lua
```diff
@@ -89,8 +89,8 @@ function main(package)
         if vs_toolset then
             local vs_toolset_ver = import("core.base.semver").new(vs_toolset)
             local minor = vs_toolset_ver:minor()
-            if minor and minor >= 40 then
-                package:add("patches", "<=1.85.0", "patches/1.85.0/fix-v144.patch", "1ba99cb2e2f03a4ba489a32596c62e1310b6c73ba4d19afa8796bcf180c84422")
+            if minor and minor >= 44 then
```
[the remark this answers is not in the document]
Willaaaaaaa: `if [ MATCH "(14.3)" : $(version) ]`
I think 44 would be a better choice. ?

Willaaaaaaa:
I believe the current CI checkers don't support VS 2026 insiders whatsoever. So I can only trouble you guys to review it manually.❤

packages/b/boost/b2/load.lua
```diff
@@ -89,8 +89,8 @@ function main(package)
...
+                package:add("patches", "<=1.85.0", "patches/1.85.0/fix-msvc.patch", "6457f110fc844c9ee019c66419ba6697a8864ce79d3ff5c413d36fb734fb51a9")
```
star-hengxing: This is the `crlf` file hash, we need the `lf` file hash

Willaaaaaaa: omg, I thought it would be the same. Thanks for pointing it out!

Willaaaaaaa: Use `lf` file hash only. Ref: https://github.com/xmake-io/xmake/blob/2f4d778bcbde53dd6a854f84b9070aee77b7853b/xmake/modules/private/action/require/impl/actions/patch_sources.lua#L29-L49

PR #8277, opened 2025-09-30
omath: update to 3.8.1

luadebug:
Drop old version. Keep only latest version for now.

### DIFF
--- a/packages/o/omath/xmake.lua
+++ b/packages/o/omath/xmake.lua
@@ -10,7 +10,7 @@ package("omath")
 
     add_patches("v3.8.1", "patches/v3.8.1/fix-build.patch", "c1554cf0cdd027d6386544871d6248c868f8f95add343660334888da52119ae9")
 
-    if is_arch("x86", "i386", "x64", "x86_64") then
+    if is_arch("x86_64", "x64", "x86", "i386", "i686") then
         add_configs("avx2",  {description = "Enable AVX2", default = true, type = "boolean"})
     end
     add_configs("imgui", {description = "Define method to convert omath types to imgui types", default = true, type = "boolean"})


luadebug:
BTW https://github.com/xmake-io/xmake-repo/pull/8280 has similiar issue but not for `imgui` but for `minizip`.
I am pretty sure workaround for that case similiar to imgui for this PR:
```
error: @programdir/core/sandbox/modules/os.lua:273: CMake Error at CMakeLists.txt:316 (find_package):
  Could not find a configuration file for package "minizip" that is
  compatible with requested version "".

  The following configuration files were considered but not accepted:

    /home/runner/.xmake/packages/m/minizip/v1.2.13/f067aec52be74dbca44545ce4382d6ca/lib/cmake/minizip/minizipConfig.cmake, version: 1.0.0 (64bit)



ompiler identification is GNU 13.3.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/runner/work/xmake-repo/xmake-repo/arm-gnu-toolchain-13.3.rel1-x86_64-arm-none-linux-gnueabihf/bin/arm-none-linux-gnueabihf-g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found ZLIB: /home/runner/.xmake/packages/z/zlib/v1.3.1/f19d02983c2441c481b4cbe5802d62c1/lib/libz.a (found suitable version "1.3.1", minimum required is "1.2.8")
-- Configuring incomplete, errors occurred!

stack traceback:
    [C]: in function 'error'
    [@programdir/core/base/os.lua:1085]:
    [@programdir/core/sandbox/modules/os.lua:273]:
    [@programdir/core/sandbox/modules/os.lua:291]: in function 'vrunv'
    [@programdir/modules/package/tools/cmake.lua:1310]: in function 'configure'
    [@programdir/modules/package/tools/cmake.lua:1352]: in function 'install'
    [...xmake-repo/xmake-repo/packages/l/libxlsxwriter/xmake.lua:84]: in function 'script'
    [...dir/modules/private/action/require/impl/utils/filter.lua:114]: in function 'call'
    [.../modules/private/action/require/impl/actions/install.lua:472]:

  => install libxlsxwriter v1.2.3 .. failed
  ```

File: packages/i/imgui/xmake.lua [hunk 1]
```diff
@@ -214,6 +214,12 @@ package("imgui")
 
         os.cp(path.join(package:scriptdir(), "port", "xmake.lua"), "xmake.lua")
         import("package.tools.xmake").install(package, configs)
+        local config_version_file = path.join(package:installdir("lib"), "cmake", "imgui", "imguiConfigVersion.cmake")
```
star-hengxing, line 217: I don't think modifying the cmake config file in the package description is the right choice. If modifications are needed, then a PR should be submitted to xmake. If the cmake config is unstable, pkgconfig should be prioritized instead.

luadebug: imgui cmake config was supposed to be stable, the problem is that XMake repository itself need fix for auto-generated CMakeConfigs during cross build as I show that so expected fix here is to improve auto-generation mechanism itself, we would just keep getting similiar issue over and over as #8280 minizip auto-generate cmake config for cross?

```add_rules("utils.install.cmake_importfiles")``` is what generate 64 bit instead of 32 bit for cross builds.

star-hengxing: So you need to improve here.
https://github.com/xmake-io/xmake/blob/3c3f90e4407257bacdeb6939aabdeb7fc732f355/xmake/scripts/cmake_importfiles/xxxConfigVersion.cmake#L39
https://github.com/xmake-io/xmake/blob/5670477e415e24962c57eac21194c63946ee5e1e/xmake/modules/target/action/install/cmake_importfiles.lua#L50

waruqi: I merged it first, and you can remove this patch after the new version of xmake is released.

PR #8278, opened 2025-09-30
Fix/libgit2 openssl3 linux

MasterLaplace:
## What's the issue?

I ran into this annoying linking error when trying to build my [Flakkari game server project](https://github.com/MasterLaplace/Flakkari) on Ubuntu 24.04:

```sh
error: /usr/bin/ld: /.xmake/packages/l/libgit2/v1.9.1/.../lib/libgit2.a(openssl.c.o): in function `openssl_certificate':
openssl.c:(.text+0x5cb): undefined reference to `SSL_get1_peer_certificate'
/usr/bin/ld: /.xmake/packages/l/libgit2/v1.9.1/.../lib/libgit2.a(openssl.c.o): in function `openssl_connect':
openssl.c:(.text+0x826): undefined reference to `SSL_get1_peer_certificate'
collect2: error: ld returned 1 exit status
```

Turns out this happens on any recent Linux distro (Ubuntu 24.04+, Fedora 36+) because OpenSSL 3.0+ deprecated some APIs that libgit2 uses, but libgit2 isn't compiled with the right compatibility flags.

There's also a bonus issue: if you use both libgit2 and libcurl in the same project, you can end up with version conflicts where one uses OpenSSL 1.x and the other uses 3.x.

## How I fixed it

**For libgit2:** Added the `OPENSSL_API_COMPAT=0x10100000L` flags when building with OpenSSL 3.0+ on Linux. This tells OpenSSL to keep the old APIs available for backward compatibility.

**For libcurl:** Made it prefer OpenSSL 3.0+ by default on Linux so both libraries use the same version and don't fight each other.

Both changes only kick in on Linux with OpenSSL 3.0+, so nothing breaks on other platforms.

## Tested on

- Ubuntu 24.04 with OpenSSL 3.0.13 ✓
- [Flakkari project](https://github.com/MasterLaplace/Flakkari) builds successfully with both libgit2 and libcurl ✓
- Didn't break anything on other setups ✓

## What this fix allows

Before this fix, my Flakkari project had a simple approach that would just break on modern Linux:

```lua
add_requires("nlohmann_json", "singleton", "libcurl", "libgit2")
```

Now with this fix, I can use proper cross-platform configuration that actually works:

```lua
add_requires("libcurl", {configs = {openssl3 = is_plat("linux", "macosx")}})
add_requires("libgit2", {configs = {https = is_plat("windows") and "winhttp" or "openssl3", tools = false}})
```

You can see the exact changes here: [Flakkari project dependencies](https://github.com/MasterLaplace/Flakkari/commit/d034937#diff-1e65723e47b81d67c25a98980a10861a012068422521d2331e559e547a4e79cc).

This should fix the build issues for anyone running into this on modern Linux distros.

### DIFF
--- a/packages/l/libgit2/xmake.lua
+++ b/packages/l/libgit2/xmake.lua
@@ -106,7 +106,7 @@ package("libgit2")
                 if package:is_plat("windows", "mingw", "msys") then
                     table.join2(links, {"ws2_32", "advapi32", "bcrypt"})
                 end
-    
+
                 io.replace("cmake/FindmbedTLS.cmake",
                     [["-L${MBEDTLS_LIBRARY_DIR} -l${MBEDTLS_LIBRARY_FILE} -l${MBEDX509_LIBRARY_FILE} -l${MBEDCRYPTO_LIBRARY_FILE}"]],
                     table.concat(links, " "), {plain = true})
@@ -122,6 +122,13 @@ package("libgit2")
             "-DUSE_HTTP_PARSER=llhttp",
             "-DUSE_GSSAPI=OFF"
         }
+
+        -- Fix OpenSSL 3.0+ compatibility on Linux
+        if package:is_plat("linux") and https == "openssl3" then
+            table.insert(configs, "-DCMAKE_C_FLAGS=-DOPENSSL_API_COMPAT=0x10100000L")
+            table.insert(configs, "-DCMAKE_CXX_FLAGS=-DOPENSSL_API_COMPAT=0x10100000L")
+        end
+
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
         table.insert(configs, "-DBUILD_SHARED_LIBS=" .. (package:config("shared") and "ON" or "OFF"))
         table.insert(configs, "-DUSE_SSH=" .. (package:config("ssh") and "ON" or "OFF"))
@@ -135,7 +142,7 @@ package("libgit2")
         end
         import("package.tools.cmake").install(package, configs, opt)
         if package:is_plat("linux") and linuxos.name() == "fedora" then
-            io.replace(path.join(package:installdir("lib/pkgconfig"), "libgit2.pc"), 
+            io.replace(path.join(package:installdir("lib/pkgconfig"), "libgit2.pc"),
                 "Requires.private: openssl ", "Requires.private: openssl3 ", {plain = true})
         end
     end)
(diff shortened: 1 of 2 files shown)


packages/l/libgit2/xmake.lua
```diff
@@ -61,6 +61,11 @@ package("libgit2")
             package:add("deps", https)
         end
 
+        -- Fix OpenSSL 3.0+ compatibility on Linux
+        if package:is_plat("linux") and https == "openssl3" then
+            package:add("defines", "OPENSSL_API_COMPAT=0x10100000L")
```
SirLynix: this won't build libgit2 with these defines, you can use opt for this
```lua
opt.cxflags = {"-DOPENSSL_API_COMPAT=0x10100000L"}
```

MasterLaplace: Done, thanks for the suggestion

MasterLaplace:
I think the OpenSSL fix for libgit2 is good to go, but the CI is failing on two unrelated configurations (WASM and cross ARM):

1. WASM (ubuntu-latest, static)
```
error: incompatible pointer types passing 'size_t *' (aka 'unsigned long *')
to parameter of type 'unsigned int *' [-Wincompatible-pointer-types]
/source/src/util/array.h:55:35
```
There's an old bug in libgit2 related to Emscripten with `size_t*` vs `unsigned int*` types.
Also, the OpenSSL define appears correctly: `-DOPENSSL_API_COMPAT=0x10100000L`

2. Cross ARM (ubuntu-latest, arm-none-linux-gnueabihf)
```
error: package(llhttp >=9.3.0) unsupported cross pltform
packages/l/llhttp/xmake.lua:16
```
This is linked to `llhttp` v9.3.0 which had cross-compilation issues. The restriction was [added in xmake-repo](https://github.com/xmake-io/xmake-repo/commit/919b64f478daa247a49174775087848f3d1968a5) on August 5, 2025
```lua
on_check("cross", function (package)
    if package:version():ge("9.3.0") then
        raise("package(llhttp >=9.3.0) unsupported cross pltform")
    end
end)
```

By the way, it would be cool to change the error message to "... cross platform" instead of "... cross pltform"

**Potential fixes for the 2 failing platforms:**

Fix WASM Emscripten size_t/unsigned int pointer type conflicts by adding warning suppression
```lua
if package:is_plat("wasm") then
    opt.cxflags = opt.cxflags or {}
    table.insert(opt.cxflags, "-Wno-incompatible-pointer-types")
end
```

Use compatible llhttp version for `Cross` platform
```lua
if is_plat("cross") then
    add_deps("pcre2", "llhttp <=9.2.1")
else
    add_deps("pcre2", "llhttp")
end
```

PR #8285, opened 2025-10-01
llhttp: support cross plat

luadebug:
Since it uses two times install, first time from CMake, second time from XMake, I have to add cflags to both buildsystems.
Fix llhttp for https://github.com/xmake-io/xmake-repo/pull/8278


luadebug:
MacOS container behaves unpredirectable, always meet 500 error, just as shown for https://github.com/xmake-io/xmake-repo/pull/8284 perhaps github macos containers have issues now.

The job was not acquired by Runner of type hosted even after multiple attempts. #8278 #8286 well macos containers still have that issue.

PR #8311, opened 2025-10-04
Allow linux-arm64 builds for libavif

BlueGlassBlock:
Mitigations have been deployed on libyuv's xmake file in #7677.


BlueGlassBlock:
Closed in favor of #8310

PR #8331, opened 2025-10-06
nzsl: fix compilation when asan is enabled

SirLynix:
asan policy is not passed to xmake tool, so we need to pass it ourselves


waruqi:
no any ci tests? you should push to dev branch.

SirLynix:
I don't get why master was used, rebasing

PR #8337, opened 2025-10-07
joltphysics: Guard JPH_PROFILE_ENABLED behind option

zachrammell:
This previously always had JPH_PROFILE_ENABLED defined despite the option existing to conditionally enable it, which can cause subtle failures with differing definitions between library and usage code.


waruqi:
Does it ready now?

PR #8338, opened 2025-10-07
minizip: Change bzip2 config default to false

star-hengxing:


SirLynix:
Shouldn't we also use packagedeps on assimp since it can break if we enable bzip2

waruqi:
https://github.com/xmake-io/xmake-repo/pull/8339

PR #8352, opened 2025-10-10
directx12-agility: add new package

luadebug:
I am confused how to create this package, feel free to suggest.
As normally there is not the best maybe way to copy paste *dx12.lib* out of system to package lib folder, but I do not know another solution.

### DIFF
--- a/packages/d/directx12-agility/port/cmakelists.txt
+++ b/packages/d/directx12-agility/port/cmakelists.txt
@@ -36,7 +36,6 @@ configure_file(
     "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake" @ONLY
 )
 
-
 write_basic_package_version_file(
     "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config-version.cmake"
     VERSION ${PROJECT_VERSION}


luadebug:
Looks good to me, let me know if something should be changed.

packages/d/directx12-agility/xmake.lua
```diff
@@ -0,0 +1,50 @@
...
+            REDIST_ARCH = "x64"
+        elseif not package:is_arch("arm.*") and not package:is_arch64() then
+            REDIST_ARCH = "win32"
+        end
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "D3D12Core.dll"), package:installdir("bin"))
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "D3D12Core.pdb"), package:installdir("bin"))
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "d3d12SDKLayers.dll"), package:installdir("bin"))
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "d3d12SDKLayers.pdb"), package:installdir("bin"))
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "d3dconfig.exe"), package:installdir("bin"))
+        os.cp(path.join("build/native/bin", REDIST_ARCH, "d3dconfig.pdb"), package:installdir("bin"))
```
waruqi, line 30: why not use ?

```lua
os.cp(path.join("build/native/bin", REDIST_ARCH, "*.exe"), package:installdir("bin"))
os.cp(path.join("build/native/bin", REDIST_ARCH, "*.dll"), package:installdir("bin"))
os.cp(path.join("build/native/bin", REDIST_ARCH, "*.pdb"), package:installdir("bin"))
```

luadebug: I meant to follow vcpkg implementation of this.
To come thinking that dx12.lib is actually IMPLIB for two .DLLs from this NuGET package, that is insane.
https://github.com/microsoft/vcpkg/blob/master/ports/directx12-agility/portfile.cmake#L26-L37
That shows if we build Debug two DLLs are being copied, but if we build Release only one DLL is being copied.
image

PR #8370, opened 2025-10-14
minizip: Improve includedirs

star-hengxing:


luadebug:
Will it help to libxlsx... PR?
Did you test minizip cmakeconfig?
Also remove cmakeconfig workaround please.

star-hengxing:
> Will it help to libxlsx... PR?

I don't know~

> Did you test minizip cmakeconfig?

I tested assimp.

> Also remove cmakeconfig workaround please.

Why?

luadebug:
>Why?

See 3.0.4 changelog refering to my PR, I mention to remove 5 lines of code (81-86) next to install that are workaround for 32 bit cross compilation auto generated cmakeconfig by xmake port.

packages/m/minizip/xmake.lua
```diff
@@ -79,7 +79,7 @@ package("minizip")
             import("package.tools.xmake").install(package, configs)
 
             local config_version_file = path.join(package:installdir("lib"), "cmake", "minizip", "minizipConfigVersion.cmake")
-            if package:is_plat("cross") and package:check_sizeof("void*") == "4" and os.exists(config_version_file) then
+            if xmake:version():lt("3.0.4") and package:is_plat("cross") and package:check_sizeof("void*") == "4" and os.exists(config_version_file) then
```
luadebug, line 82: If you're willing to provide backward compatibility with older versions of XMake could you do that for imgui package at different PR as well?

Hence I am unsure making some function like ```xmake_autocmake_config_fixup()``` which stores those repeating 5 lines of code is better.

CC: @waruqi

waruqi: We can wait for a few versions until it is removed after 3.1.x.

PR #8371, opened 2025-10-14
elfio: upgrade to 3.12, fix stdint include in elf_types.hpp.

Redbeanw44602:


luadebug:
Will it be better to create patchfile and include `<cstdint>`?

Redbeanw44602:
> Will it be better to create patchfile and include `<cstdint>`?

https://github.com/serge1/ELFIO/blob/7d30a22fc5aac06adfe7887ae57f3701b6b5f913/examples/c_wrapper/c_example.c#L26

I thought about it, but it seems that there is an intention to use `elf_types.hpp` as a C header🤨.

Hmm... is a patchfile necessary? This is just a simple fix and the issue is already resolved in HEAD.

PR #8403, opened 2025-10-16
minizip: xmake version check error

dulingzhi:
Within the on_install environment, xmake.version does not exists.

  => install minizip v1.2.13 .. failed

attempt to call a nil value (method 'version')
if you want to get more verbose errors, please see:
  -> C:\Users\XXXX\AppData\Local\.xmake\cache\packages\2510\m\minizip\v1.2.13\installdir.failed\logs\install.txt

### DIFF
--- a/packages/m/minizip/xmake.lua
+++ b/packages/m/minizip/xmake.lua
@@ -41,7 +41,6 @@ package("minizip")
         end
     end)
 
-    local is_old_xmake = xmake:version():lt("3.0.4")
     on_install(function (package)
         os.cd(path.join("contrib/minizip"))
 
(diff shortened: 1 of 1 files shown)


packages/m/minizip/xmake.lua
```diff
@@ -79,7 +80,7 @@ package("minizip")
             import("package.tools.xmake").install(package, configs)
 
             local config_version_file = path.join(package:installdir("lib"), "cmake", "minizip", "minizipConfigVersion.cmake")
-            if xmake:version():lt("3.0.4") and package:is_plat("cross") and package:check_sizeof("void*") == "4" and os.exists(config_version_file) then
+            if is_old_xmake and package:is_plat("cross") and package:check_sizeof("void*") == "4" and os.exists(config_version_file) then
```
waruqi: Conflict between built-in xmake module and tools.xmake module.

you can try

```lua
import("package.tools.xmake", {anonymous = true}).install(package, configs)

if xmake:version():lt("3.0.4") then
end
```

Issue #8458, opened 2025-10-22
raylib包编译出错


buqiaomuyudeheshang:
### Xmake 版本

v2.9.9+20250409

### 操作系统版本和架构

debian 13

### 描述问题

在进行安装raylib库的时候总是最后安装失败

### 期待的结果

能够安装raylib库

### 工程配置
```lua
set_project("webui")
add_rules("mode.debug", "mode.release")
add_requires("webui","raylib")


target("webui")
    set_kind("binary")
    add_files("src/*.c")
    add_packages("webui","raylib")
```

### 附加信息和错误日志
```console
checking for ping ... /usr/bin/ping
pinging the host(github.com) ... 0 ms
/usr/bin/tar -xf /home/yuxiaogang/.xmake/cache/packages/2510/r/raylib/5.5/5.5.tar.gz -C source.tmp
checking for gcc ... /usr/bin/gcc
checking for the c++ compiler (cxx) ... gcc
checking for flags (-fPIC) ... ok
> gcc "-fPIC" "-m64"
checking for cmake ... /usr/bin/cmake
/usr/bin/cmake -DBUILD_EXAMPLES=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/home/yuxiaogang/.xmake/packages/r/raylib/5.5/efaf75c5cc7042868d9f9fc4196a080a -DCMAKE_INSTALL_LIBDIR:PATH=lib -G "Unix Makefiles" -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LINKER_FLAGS_RELEASE= /home/yuxiaogang/.xmake/cache/packages/2510/r/raylib/5.5/source
CMake Deprecation Warning at CMakeLists.txt:10 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
[...]

PR #8459, opened 2025-10-22
raylib: improve

Willaaaaaaa:
some minor style improvement

### DIFF
--- a/packages/r/raylib/xmake.lua
+++ b/packages/r/raylib/xmake.lua
@@ -1,7 +1,7 @@
 package("raylib")
-
     set_homepage("http://www.raylib.com")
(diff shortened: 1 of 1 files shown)


luadebug (approved):
https://github.com/xmake-io/xmake-repo/issues/8458#issuecomment-3433141765

If we use

```xmake
if is_plat("windows", "linux", "macosx") then
    add_deps("opengl")
end
```

And keep `add_deps("opengl", {optional = true})` would it resolve that issue?

Willaaaaaaa:
i don't think so. But we can add checks (?): 
```xmake
if package:is_plat("windows", "linux", "macosx") then
    if not package:dep("opengl") then
        -- raise an error: plz install opengl
    end
end
```

PR #8501, opened 2025-10-27
criterion: add version 2.4.3

luadebug:
Resolves: https://github.com/xmake-io/xmake-repo/pull/8415


luadebug:
Maybe we need to use packagedeps here?

star-hengxing:
idk, you can try.

luadebug:
> idk, you can try.

Obviously it's better to wait for libgit2 package improvements, maybe if we remove cmake/Find-dependency.cmake files we would get rid of those errors

PR #8545, opened 2025-11-02
add filc toolchain

paul-reilly:
As per issue xmake-io/xmake#6981 here is the Fil-C toolchain as a toolchain package.

### DIFF
--- a/packages/f/filc/xmake.lua
+++ b/packages/f/filc/xmake.lua
@@ -18,7 +18,7 @@ package("filc")
         os.cp(path.join(os.curdir(), "*"), installdir)
 
         -- now we replicate the commands in setup.sh...
-        local pathx = path.joinenv(path.join(installdir, "pizfix/lib64"), path.join(installdir, "pizfix/lib"))
+        local pathx = path.join(installdir, "pizfix/lib64") .. ":" .. path.join(installdir, "pizfix/lib")
         os.runv("patchelf", {"--set-rpath", pathx, "pizfix/lib/libc.so"})
         os.runv("patchelf", {"--set-rpath", pathx, "pizfix/lib/libpizlo.so"})
         os.runv("patchelf", {"--set-rpath", pathx, "pizfix/lib/libc++.so.1.0"})


packages/f/filc/xmake.lua
```diff
@@ -0,0 +1,19 @@
+package("filc")
+    set_kind("toolchain")
+    set_homepage("https://fil-c.org/")
+    set_description("A memory safe implementation of the C and C++ programming languages.")
+    set_license("LLVM")
+
+    set_urls("https://github.com/pizlonator/fil-c/releases/download/v$(version)/filc-$(version)-linux-x86_64.tar.xz")
+    add_versions("0.673", "bc4bc7ec9fcd8abe559b8ce85b4c39b81d64ca6e3b23f89a6656f04233fc3395")
+
+    on_install("@linux|x86_64", function (package)
```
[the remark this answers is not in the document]
paul-reilly: .

packages/f/filc/xmake.lua
```diff
@@ -0,0 +1,26 @@
...
+        end
+    end
+
+    add_deps("patchelf") -- needed for setup.sh
+
+    on_install("@linux|x86_64", function (package)
+        local installdir = package:installdir()
+        os.cp(path.join(os.curdir(), "*"), installdir)
+        os.runv(path.join(installdir, "setup.sh"), {}, {curdir = installdir})
+        package:addenv("PATH", "build/bin")
```
waruqi, line 40: installdir/build/bin?

packages/f/filc/xmake.lua
```diff
@@ -0,0 +1,26 @@
...
+            add_versions("0.674", "a8ec349f383a49dacc09a9540643164c67081245e591c1e6823609653f2c9740")
+        end
+    end
+
+    add_deps("patchelf") -- needed for setup.sh
+
+    on_install("@linux|x86_64", function (package)
+        local installdir = package:installdir()
+        os.cp(path.join(os.curdir(), "*"), installdir)
+        os.runv(path.join(installdir, "setup.sh"), {}, {curdir = installdir})
```
waruqi: {shell = true}

packages/f/filc/xmake.lua
paul-reilly: Yes, that's the directory structure of the project. There's another populated directory at root level and I didn't want to change relative locations. I can move the installdir/build/bin|include|lib to installdir/bin|include|lib if you want.

packages/f/filc/xmake.lua
paul-reilly: Done, thanks!

packages/f/filc/xmake.lua
waruqi: pizfix? if it is unused, we should remove it. It will take up additional disk space. and only copy filc to `installdir/bin`, `installdir/include`, `installdir/lib`

if it is used, 

and we should not run `setup.sh`, it will do some system ops.

```sh
ln -s /usr/include/linux .
if test -d /usr/include/x86_64-linux-gnu/asm
then
    ln -s /usr/include/x86_64-linux-gnu/asm .
else
    ln -s /usr/include/asm .
fi
ln -s /usr/include/asm-generic .
cd ../..
set +x
```

waruqi: If some of its operations are necessary, you can use os.cp/os.ln to implement them, or modify the setup.sh script.

paul-reilly: Thanks, yes the pizfix directory is required. All the operations in the setup.h script are required too. I'll copy the directories of system files to where those local symlinks to the system dirs are created.

paul-reilly:
Just to confirm, it does NOT work when copying `packagedir/build/bin|include|lib` to `packagedir/bin|include|lib` so I've kept `packagedir/build/bin` added to the path.

packages/f/filc/xmake.lua
```diff
@@ -0,0 +1,47 @@
...
+
+    add_deps("patchelf") -- needed for setup.sh
+
+    on_install("@linux|x86_64", function (package)
+        local installdir = package:installdir()
+        os.cp(path.join(os.curdir(), "*"), installdir)
+
+        -- now we replicate the commands in setup.sh...
+        local pathx = path.join(installdir, "/pizfix/lib64") .. ":" .. path.join(installdir, "/pizfix/lib")
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib/libc.so"}, {shell=true})
```
waruqi: patchelf is not shell script file, we need not to set `{shell = true}`

waruqi:
```lua
os.runv("patchelf", {"--set-rpath", pathx, "pizfix/lib/libc.so"})
```

packages/f/filc/xmake.lua
```diff
@@ -0,0 +1,47 @@
...
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib/libc++.so.1.0"}, {shell=true})
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib/libc++abi.so.1.0"}, {shell=true})
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib_test/libpizlo.so"}, {shell=true})
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib_test_gcverify/libpizlo.so"}, {shell=true})
+        os.runv("patchelf --set-rpath", {pathx, "pizfix/lib_gcverify/libpizlo.so"}, {shell=true})
+
+        -- ... but copy dirs instead of symlinking to system dirs
+        local dest = path.join(installdir, "pizfix", "os-include")
+        os.cp("/usr/include/linux", path.join(dest, "linux"))
+        if os.exists("/usr/include/x86_64-linux-gnu/asm") then
```
star-hengxing, line 32: `os.trycp`

packages/f/filc/xmake.lua
paul-reilly: Thanks, will do.

packages/f/filc/xmake.lua
paul-reilly: Should we `assert` any `os.trycp` failures?

PR #8564, opened 2025-11-05
gfortran: compatible with non-English locales

choyy:


choyy:
@star-hengxing how to handle failed ci? limit them in on_fetch?

star-hengxing:
I think we can directly merge.

PR #8568, opened 2025-11-05
subhook: Add mirror for this no-longer-public repository

Willaaaaaaa:
- ~`PucklaJ/dynareadout`: No re-uploaded mirror found.~
- `Zeex/subhook`: Found `Dasharo/subhook` and `xmake-mirror/subhook`.


Willaaaaaaa:
seems that `on_load`, unlike `on_check`, doesn't have a `try-catch` block. So `raise` will directly cause an error. Any good idea about this? @star-hengxing 
I don't want to use `raise` inside `on_check`, because it's called after a bunch of other funtions.

packages/s/subhook/xmake.lua
```diff
@@ -6,6 +6,10 @@ package("subhook")
     add_urls("https://github.com/Zeex/subhook.git")
     add_versions("2023.02.10", "e935959d2f9cc642bcbb5e7759b2b1e7196b0947")
 
+    on_load(function (package)
+        raise("package(subhook) is no longer public (Updated on 2025/11/05).")
+    end)
+
```
luadebug: https://github.com/xmake-mirror/subhook
I made mirror for this, feel free to add_version 0.8.2?

star-hengxing:
> seems that `on_load`, unlike `on_check`, doesn't have a `try-catch` block. So `raise` will directly cause an error. Any good idea about this? @star-hengxing I don't want to use `raise` inside `on_check`, because it's called after a bunch of other funtions.

We can directly merge this pr without ci passed, because user already see the error message.
```
error: @programdir/core/main.lua:272: /__w/xmake-repo/xmake-repo/packages/d/dynareadout/xmake.lua:32: package(dynareadout) is no longer public (Updated on 2025/11/05).
```

packages/s/subhook/xmake.lua
Willaaaaaaa: Thanks a lot! Added this repo url.

Willaaaaaaa:
> We can directly merge this pr without ci passed, because user already see the error message.
> 
> ```
> error: @programdir/core/main.lua:272: /__w/xmake-repo/xmake-repo/packages/d/dynareadout/xmake.lua:32: package(dynareadout) is no longer public (Updated on 2025/11/05).
> ```

OK. I temporarily use `on_check`. After `subhook` passes CI, we can revert https://github.com/xmake-io/xmake-repo/pull/8568/commits/0f9d4efd8db75c039662ea750c4540d5695575ff if necessary.

Willaaaaaaa:
@luadebug Why closed #8585 ? Your solution seems much better than this one.

luadebug:
> @luadebug Why closed #8585 ? Your solution seems much better than this one.

That would create conflict for your branch, so I think that is better to place those changes in your branch to resolve conflict.

Willaaaaaaa:
> I think that is better to place those changes in your branch to resolve conflict.

Like this? cherry-pick

luadebug:
Yup 👍😊 so we would resolve branch conflicts beforehand.

packages/d/dynareadout/xmake.lua
```diff
@@ -1,58 +1,40 @@
...
-    add_versions("23.06",   "515f0b0d20c46e00f393fb9bb0f2baf303244d39e35a080741276681eb454926")
-    add_versions("23.07",   "6db52a13771cdb7613b97cf0d2bcffdb87ce0cce4cba7e6d80330977b2ac6210")
-    add_versions("23.08",   "5be972426f67ce63a7ac3beaf3866b824abbc9c15af2d47d1fea21687417b493")
-    add_versions("23.09",   "a62e0575521aacf3f567dfd578d6edc51edc07d4b744e5b5ae5d30f662be424b")
-    add_versions("23.10",   "834d356e80019a7288001c2796c9ce14c2a8e5494c1051fae402f4503b10c1e5")
-    add_versions("23.10.1", "6ebb6e9d9b3db637476cc9bd5342e4779be175f87225261da35c9270790e77d7")
-    add_versions("23.11",   "2222bea87af9289dbc1a52adc5f09058863c503003e94193ca9388eff9e4ff04")
-    add_versions("23.12",   "6afc9c0857b6227e39466aae00db606d3376e61d518bb73544d8240fe3a66a90")
-    add_versions("24.01",   "f39094567272816c9a7dd84d3eaa0ef5c26309eeeadba814cac12f82e93ae0e1")
-    add_versions("24.02",   "acd483bd737a7769087befa1eb2010426c1328bb84ab0481ea11cdeb7655c64e")
```
waruqi, line 26: why remove old versions? it will break some user projects.

packages/d/dynareadout/port/xmake.lua
```diff
@@ -0,0 +1,106 @@
+set_project("dynareadout")
+
+option("build_test")
+    set_default(false)
+    set_showmenu(true)
```
waruqi:
```lua
option("build_test", {default = false})
```

packages/d/dynareadout/port/xmake.lua
```diff
@@ -0,0 +1,106 @@
...
+    if not get_config("profiling") then
+        remove_files("src/profiling.c")
+    end
+    add_headerfiles("src/*.h")
+    if is_kind("shared") then
+        add_rules("utils.symbols.export_all")
+    end
+target_end()
+
+if get_config("build_cpp") or get_config("build_python") then
```
waruqi: use has_config instead of get_config

packages/d/dynareadout/port/xmake.lua
```diff
@@ -0,0 +1,106 @@
...
+        add_packages("pybind11")
+        add_options("profiling")
+        add_files("src/python/*.cpp")
+        add_headerfiles("src/python/*.hpp")
+        add_includedirs("src", "src/cpp")
+        add_rpathdirs("@executable_path")
+
+        on_load(function (target)
+            local ext_file = os.tmpfile()
+            os.execv("python3-config", {"--extension-suffix"}, {stdout=ext_file})
```
waruqi: use os.iorunv to get stdout.

and python3 is used, but package has no python deps. `add_deps("python 3.x")`

and we need also add python requires to bind python envs in port/xmake.lua

```lua
add_requires("python 3.x")

target()
    add_packages("python")
```

waruqi: and we should use `python.module` rule instead of it. it will modify extension.

https://github.com/xmake-io/xmake/blob/72f441c0a6f0ac864e7b4b0c06e8523434e4b599/xmake/rules/python/module/xmake.lua#L35

waruqi:
These are two unrelated packages; we should split them into two separate pr files.

Willaaaaaaa:
> These are two unrelated packages; we should split them into two separate pr files.

My initial thought was just showing up an error. I didn't think of adding a mirror, so I put them together. my bad.

So in this way, @luadebug can reopen #8585 and make modifications according to ruki's advice, and I can revert all the `dynareadout` part in the same time. Is this a reasonable solution?

packages/d/dynareadout/xmake.lua
Willaaaaaaa: > why remove old versions? it will break some user projects.

Our mirror have no other versions. Even the currently existing version is not a repo-author released one.
https://dashboard.stablebuild.com/pypi-deleted-packages/pkg/dynareadout have some source code. But it neither has a `.git` folder nor is well-structed. Even if we added them one by one manually, we can't guarantee they are the same as the old ones.

waruqi:
> > These are two unrelated packages; we should split them into two separate pr files.
> 
> My initial thought was just showing up an error. I didn't think of adding a mirror, so I put them together. my bad.
> 
> So in this way, @luadebug can reopen #8585 and make modifications according to ruki's advice, and I can revert all the `dynareadout` part in the same time. Is this a reasonable solution?

sure, you can try.

packages/d/dynareadout/xmake.lua
```diff
@@ -1,58 +1,40 @@
...
         end
     end)
 
     on_install("windows", "linux", "macosx", "mingw", function (package)
+        os.cd("lib/dynareadout")
+        os.cp(path.join(package:scriptdir(), "port", "xmake.lua"), "xmake.lua")
         local configs = {}
         configs.build_test = "n"
-        configs.build_cpp = package:config("cpp") and "y" or "n"
+        configs.build_cpp = package:config("cpp_bind") and "y" or "n"
```
star-hengxing:
```diff
- configs.build_cpp = package:config("cpp_bind") and "y" or "n"
+ configs.build_cpp = package:config("cpp_bind")
```

star-hengxing:
Can we merge now?

Willaaaaaaa:
> Can we merge now?

yes, this one is ready. and if luadebug is busy these days, i can submit dynareadout this sunday.🥺

luadebug:
Well indeed I am busy currently, but that's fine feel free to submit PR for dyna whenever you want and wish to do so. 👍😅

PR #8580, opened 2025-11-07
opencv: fix build & cmake

choyy:
cmake error: https://github.com/xmake-io/xmake-repo/pull/8484
```
CMake Error at /github/home/.xmake/packages/o/opencv/4.12.0/f64cc25c8ea945fd9fed9b60af1ed663/lib/cmake/opencv4/OpenCVModules.cmake:363 (set_target_properties):
  The link interface of target "opencv_wechat_qrcode" contains:

    Iconv::Iconv

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /github/home/.xmake/packages/o/opencv/4.12.0/f64cc25c8ea945fd9fed9b60af1ed663/lib/cmake/opencv4/OpenCVConfig.cmake:128 (include)
  CMakeLists.txt:221 (FIND_PACKAGE)
```

### DIFF
--- a/packages/o/opencv/xmake.lua
+++ b/packages/o/opencv/xmake.lua
@@ -1,5 +1,4 @@
 package("opencv")
-
     set_homepage("https://opencv.org/")
     set_description("A open source computer vision library.")
     set_license("Apache-2.0")


packages/o/opencv/xmake.lua
```diff
@@ -237,7 +238,13 @@ package("opencv")
                 shflags = {"-Wl,-Bsymbolic"}
             end
         end
-        import("package.tools.cmake").install(package, configs, {buildir = "bd", shflags = shflags, ldflags = ldflags})
+        import("package.tools.cmake").install(package, configs, {builddir = "bd", shflags = shflags, ldflags = ldflags})
+
+        if package:is_plat("linux", "macosx") then
```
luadebug: https://github.com/opencv/opencv/issues/20404#issuecomment-1207259071
Maybe related, as for MinGW.

If zxing.hpp is like this (according to contrib resource https://codeload.github.com/opencv/opencv_contrib/tar.gz/refs/tags/4.12.0): That would be `if not package:is_plat("windows", "android")` ?

```cpp
#if defined(__ANDROID_API__) || defined(_MSC_VER)

#ifndef NO_ICONV
#define NO_ICONV
#endif

#endif
```

If we apply https://github.com/lebarsfa/Packages/commit/4b271d2b5c2ed6eb36a5e140c0e9e9fffb7977a7#diff-e13272f82adfb96ae33f958418128ddbc9bd47bc4819f0ddb82ca6358f365bc6 `choco/libopencv-dev/tools/no_iconv_mingw.patch` that would be  `if not package:is_plat("windows", "android", "mingw")` ?

choyy: https://github.com/opencv/opencv_contrib/pull/2916 disabled MinGW, https://github.com/opencv/opencv_contrib/pull/2931 enabled MinGW later. Then it would be `not package:is_plat("windows", "android")`. But I'm not sure if the cmake file is like `lib/cmake/opencv4/OpenCVModules.cmake` on MinGW.

luadebug: Everything looks good, if we could do check by package version for MinGW filter maybe it would be better...?

choyy: It seems no version was released during MinGW was disabled?

PR #8617, opened 2025-11-14
oneshot: add package

gibbz00:


packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,24 @@
+package("oneshot")
+    set_kind("library", {headeronly = true})
+    -- IMPROVEMENT: configurable standalone or boost asio
+    add_deps("boost", {configs = {asio = true}})
```
Willaaaaaaa: You can use `add_config()` to choose "configurable standalone or boost asio". And then in `on_load`, you can decide which dep to add according to this config.

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+
+    set_homepage("https://github.com/ashtum/oneshot")
+    set_description("A single-header scheduler aware C++ oneshot channel.")
+    set_license("BSL-1.0")
```
Willaaaaaaa, line 5: Plz move these lines to L3 like other pkg descriptions.

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+
+    add_urls("https://github.com/ashtum/oneshot.git")
+
```
Willaaaaaaa, line 8:
```suggestion
    add_versions("2024.12.29", "57ab00d924d6436e3ee649be1a14a349fb52a8d0")

```

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+    on_install(function (package)
+        os.cp("include", package:installdir())
+    end)
+
+    on_test(function (package)
+        local cxflags = {}
```
Willaaaaaaa: You can remove this line and `, cxflags = cxflags` below since it's empty.

packages/o/oneshot/xmake.lua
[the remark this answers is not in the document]
Willaaaaaaa: Yes, you can read this as an example:
https://github.com/xmake-io/xmake-repo/blob/569aab2db4d2c9bd189e761f4ac5077965ddcd69/packages/w/websocketpp/xmake.lua#L10
https://github.com/xmake-io/xmake-repo/blob/569aab2db4d2c9bd189e761f4ac5077965ddcd69/packages/w/websocketpp/xmake.lua#L14-L17

packages/o/oneshot/xmake.lua
gibbz00: Thanks, I was unsure what to use since the package does not seem to have a tagged realease 😊

packages/o/oneshot/xmake.lua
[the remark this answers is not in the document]
gibbz00: Thanks!

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,24 @@
...
+        assert(package:check_cxxsnippets({test = [[
+            #include <oneshot.hpp>
+            void test() {
+                auto [sender, receiver] = oneshot::create<std::string>();
+            }
+        ]]}, {configs = {languages = "cxx17", cxflags = cxflags}}))
+    end)
```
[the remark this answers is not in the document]
gibbz00: Surely, this must be the AI hallucinating?

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,35 @@
+package("oneshot")
+
```
Willaaaaaaa: Remove this line

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+    on_load(function (package)
+        if package:config("asio_standalone") then
+            package:add("defines", "ONESHOT_ASIO_STANDALONE")
+            package:add("deps", "asio")
+        else
+            package:add("deps", "boost", {configs = {asio = true}})
+        end
+    end)
+
```
Willaaaaaaa, line 21: And this, plz. It will be neater~

packages/o/oneshot/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+
+    on_install(function (package)
```
Willaaaaaaa: I guess we need "!wasm", because the previous ci showed asio don't support wasm.
https://github.com/xmake-io/xmake-repo/actions/runs/19358166128/job/55384420696?pr=8617#step:5:15756
```
/home/runner/.xmake/packages/b/boost/1.88.0/1440e794752d45de9b6be951d5b3b27f/include/boost/asio/detail/signal_blocker.hpp:27:3: error: Only Windows and POSIX are supported!
   27 | # error Only Windows and POSIX are supported!
      |   ^
```

packages/o/oneshot/xmake.lua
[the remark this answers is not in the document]
luadebug: We could sorta try to do add that maybe? As we have both of them `asio` and `boost`? And indeed that would require to alter on_load handler that if package:config ig... rather following generated solution...
Nvm that is done already...

gibbz00:
Thanks @luadebug for doing the last bits  🙏🙏

PR #8644, opened 2025-11-19
cppli: add package

TheBearodactyl:
Adds cppli, which is a C++ CLI application library that I wrote for fun a while back

### DIFF
--- a/packages/c/cppli/xmake.lua
+++ b/packages/c/cppli/xmake.lua
@@ -20,7 +20,7 @@ package("cppli")
         ]]}, {configs = {languages = "c++20"}}), "package(cppli): need std::source_location from <source_location> header.")
     end)
 
-    on_install("!bsd", function (package)
+    on_install(function (package)
         local configs = {}
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
         table.insert(configs, "-DCPPLI_BUILD_TESTS=OFF")


packages/c/cppli/xmake.lua
```diff
@@ -0,0 +1,25 @@
+package("cppli")
+set_homepage("https://cppli.bearodactyl.dev")
```
waruqi: incorrect indent

luadebug:
CMake 4.1.1 or higher is required.  You are running version 4.0.3
Well yet CMake package was not updated further than 4.0.3

packages/c/cppli/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+    add_urls("https://github.com/TheBearodactyl/cppli.git")
+    add_versions("2025.10.22", "98c8c2e8ee65d7a5a6b160cf0b85ba1be39ffb05")
+
+    add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
+
+    add_deps("cmake")
+    
+    on_install(function(package)
+        local configs = {}
+        table.insert(configs, "-DCMAKE_POLICY_VERSION_MINIMUM=3.5")
```
luadebug: Normally XMake maybe was supposed to pass that...?

packages/c/cppli/xmake.lua
```diff
@@ -0,0 +1,27 @@
...
+        import("package.tools.cmake").install(package, configs)
+    end)
+    
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            void test() {
+                cli::Parser parser("myapp", "A test application");
+                std::string help = parser.generate_help();
+            }
+        ]]}, {configs = {languages = "c++20"}, includes = "cppli.hpp"}))
```
luadebug, line 36: Library comes as c++23 target, though c++20 should be fine.

packages/c/cppli/xmake.lua
```diff
@@ -0,0 +1,27 @@
+package("cppli")
+    set_homepage("https://cppli.bearodactyl.dev")
+    set_description("an intuitive CLI framework for C++")
+    
+    add_urls("https://github.com/TheBearodactyl/cppli.git")
+    add_versions("2025.10.22", "98c8c2e8ee65d7a5a6b160cf0b85ba1be39ffb05")
+
+    add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
```
luadebug, line 11: That is static only target.

luadebug:
Maybe FreeBSD runner is broken... %_%

PR #8663, opened 2025-11-22
emhash: update to v1.0.1 & clean up

apocelipes:


apocelipes:
@star-hengxing The CMakelists.txt is for its unit tests and benchmarks.

star-hengxing:
Using `cmake install` will provide a config file, allowing other CMake libraries to use `find_package`.

PR #8685, opened 2025-11-25
update openssl3 to 3.6.0 and 3.5.4

charlesseizilles:


star-hengxing:
You can add the patch https://github.com/openssl/openssl/issues/28745

charlesseizilles:
I had to add the .gitattributes because I'm on windows and git pulled the patches with CRLF line endings, and the SHA256 were wrong.
Shouldn't this be done at the repository's scale?

SirLynix:
> I had to add the .gitattributes because I'm on windows and git pulled the patches with CRLF line endings, and the SHA256 were wrong.
> Shouldn't this be done at the repository's scale?

Don't do that, compute the SHA256 for the file with LF lines instead, xmake will handle the case where git pulls CRLF

packages/o/openssl3/patches/.gitattributes
```diff
@@ -0,0 +1 @@
+*.patch text eol=lf
```
waruqi: ?

PR #8710, opened 2025-11-28
colmap: add package

choyy:


star-hengxing:
Seems like we need to unbundle freeimage deps first.

choyy:
trying to fix it

PR #8719, opened 2025-11-29
Update TiltedCore add version 0.2.8

rfortier:
Update just adds the latest TiltedCore version 0.2.8

### DIFF
--- a/packages/t/tiltedcore/xmake.lua
+++ b/packages/t/tiltedcore/xmake.lua
@@ -33,7 +33,7 @@ package("tiltedcore")
         end
     end)
 
-    on_install(function (package)
+    on_install("windows", "mingw", "linux", "cross", "android", "macosx|!arm*", function (package)
         import("package.tools.xmake").install(package)
     end)
 


rfortier:
@luadebug, can I ask what you are trying to accomplish with these commits? All I was doing was added a new version to TiltedCore.

I did not find a need to include <mutex.h> for things to compile properly, and it appears you backed that part out. 
Also, we never designed a shared-library version of TiltedCore; it is meant to be statically linked. There's a comment to "keep it static," but I don't see that change backed out.

This package has no application on platforms other than Windows, and possibly Linux I'd have to check if we use it there. Why are you trying to add others? We've never done a minute of testing on them, and it is misleading to hint that they are supported.

SirLynix:
> I did not find a need to include <mutex.h> for things to compile properly, and it appears you backed that part out.

It does fail on Linux ARM64 without that mutex include:
https://github.com/xmake-io/xmake-repo/actions/runs/19789732900/job/56716625366

> Also, we never designed a shared-library version of TiltedCore; it is meant to be statically linked. There's a comment to "keep it static," but I don't see that change backed out.

That's what luadebug did, it prevented the library to be used as a shared library (notice how they added the shared config and forced its value to false using readonly, otherwise xmake would have added it as a mutable config).

> This package has no application on platforms other than Windows, and possibly Linux I'd have to check if we use it there. Why are you trying to add others? We've never done a minute of testing on them, and it is misleading to hint that they are supported.

It's something we do with a lot of packages but it's true that can be misleading as it could be understood as a support for those platforms.

PR #8731, opened 2025-12-01
muda: add package

star-hengxing:


star-hengxing:
Current ci can't test cuda releated library.

PR #8787, opened 2025-12-08
ffmpeg: add av1 deps support

star-hengxing:


jfulem:
@star-hengxing after this change the installation of ffmpeg on windows is failing.
`checking for make ... no
error: make not found!
  => install ffmpeg 7.1 .. failed
error: install failed!`

star-hengxing:
This PR theoretically should not break existing projects, as it only adds new options. This issue is likely related to the MSYS2 environment. If MSYS2 was installed via xrepo, please uninstall it and try again. If the problem persists, please open an issue.

PR #8791, opened 2025-12-08
Add ImGui versions 1.92.2, 1.92.3, 1.92.4 and 1.92.5

Seedking:


SirLynix:
Thanks, I added the missing versions as well

PR #8832, opened 2025-12-12
glslang-nihui: test for upstream

Willaaaaaaa:


nihui:
@Willaaaaaaa 
newly rebased branch https://github.com/nihui/glslang/tree/cpp14-3

Issue #8839, opened 2025-12-14
MoltenVK: Ask for improvement


Willaaaaaaa:
### Operating System Version and Architecture

MacOS

### Describe Bug

MoltenVK now is not well packaged currently and may cause issues including duplicate symbol and duplicate lib files.
1. Please check the relation between `libMoltenVK` and `libMoltenVKShaderConverter`. They seem to have the same symbols within.
2. Do not use add env if `1.` is confirmed as an issue.
3. Check the `os.mv` logic. Current logic may lead to `.a` and `.dylib` files exist at the same time.

### Additional Information and Error Logs

https://github.com/xmake-io/xmake-repo/pull/8822#issuecomment-3649390706

std-microblock:
<img width="1323" height="727" alt="Image" src="https://github.com/user-attachments/assets/391ffb54-d1b7-497a-9516-daa3a1394fd2" />

<img width="1611" height="1363" alt="Image" src="https://github.com/user-attachments/assets/3cd4082b-2fa3-440a-9d98-99c8418fd608" />

<img width="1111" height="394" alt="Image" src="https://github.com/user-attachments/assets/58b513df-bd87-44db-9a4d-d85da91c80ee" />

<img width="1166" height="235" alt="Image" src="https://github.com/user-attachments/assets/30cf94a0-1e43-4b9d-b1b7-e62f2c3cf63f" />

**动态库** MoltenVK 已经链接到 MoltenVKShaderConverter，不应该再链接到 converter，否则会造成符号重复

PR #8846, opened 2025-12-15
fix(moltenvk): Remove link to MoltenVKShaderConverter

std-microblock:
https://github.com/xmake-io/xmake-repo/issues/8839 https://github.com/xmake-io/xmake-repo/issues/8839#issuecomment-3653073531

### DIFF
--- a/packages/m/moltenvk/xmake.lua
+++ b/packages/m/moltenvk/xmake.lua
@@ -47,23 +47,36 @@ package("moltenvk")
     on_install("macosx", "iphoneos", function (package)
         local plat = package:is_plat("iphoneos") and "iOS" or "macOS"
         local configs = {"--" .. plat:lower()}
-        if package:debug() then
+        if package:is_debug() then
             table.insert(configs, "--debug")
         end
         os.vrunv("./fetchDependencies", configs)
-        local conf = package:debug() and "Debug" or "Release"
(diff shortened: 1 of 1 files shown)


packages/m/moltenvk/xmake.lua
Willaaaaaaa: is_debug()

Willaaaaaaa: is_debug()

Willaaaaaaa:
```diff
-         os.vrun("xcodebuild build -quiet -project MoltenVKPackaging.xcodeproj -scheme \"MoltenVK Package (" ..plat .. " only)\" -configuration \"" .. conf)
+         os.vrun("xcodebuild build -quiet -project MoltenVKPackaging.xcodeproj -scheme \"MoltenVK Package (" .. plat .. " only)\" -configuration \"" .. conf .. "\"")
```

waruqi: we should use os.vrunv instead of os.vrun

waruqi: duplicate root dir. please use variable to simplify it.

```lua
local moltenvk_rootdir = path.join("Package", conf, "MoltenVK")
```

PR #8936, opened 2025-12-29
Auto-update openal-soft to 1.25.0

SirLynix:
New version of openal-soft detected (package version: 1.24.3, last github version: 1.25.0)


SirLynix:
OpenAL-soft now builds using C++20, it seems it broke build on systems where C++20 is poorly supported

PR #8941, opened 2025-12-30
Auto-update ncnn to 20250916

waruqi:
New version of ncnn detected (package version: 2025.09.16, last github version: 20250916)


packages/n/ncnn/xmake.lua
apocelipes, line 1: Image

apocelipes: Duplicated.

PR #8948, opened 2025-12-31
alpaka: add package

star-hengxing:


packages/a/alpaka/xmake.lua
```diff
@@ -0,0 +1,24 @@
+package("alpaka")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://alpaka.readthedocs.io")
+    set_description("Abstraction Library for Parallel Kernel Acceleration 🦙")
```
luadebug: Maybe we should not use emojis here?

PR #8955, opened 2026-01-01
slikenet: add new package

luadebug:
Maybe need for `floatengine`


star-hengxing:
Maybe we can add the missing `dllimport`
https://github.com/SLikeSoft/SLikeNet/blob/d5f775d789563a2d505e2afbf99a550d990bb49e/Source/include/slikenet/Export.h#L21

star-hengxing (approved):

luadebug:
> Maybe we can add the missing `dllimport` https://github.com/SLikeSoft/SLikeNet/blob/d5f775d789563a2d505e2afbf99a550d990bb49e/Source/include/slikenet/Export.h#L21

That looks quite complicated as there are a lot of functions that do not allow `dllimport` as they are defined **remove the RAK_DLL_EXPORT macro from emplate class with all its code defined in the header file.**.

PR #8959, opened 2026-01-01
boost: update boost version to 1.90.0.

levihuayuzhang:
The new version (1.90.0+) of `boost` seems not providing `tar.gz`, use `7z` instead.


levihuayuzhang:
Test runners may need `7z`, `7za` installed to use `boost 1.90.0`'s `7z` archive.

PR #8964, opened 2026-01-01
Update libsdl3 to 3.4.0

SirLynix:
New version of libsdl3 detected (package version: 3.2.28, last github version: 3.4.0)

### DIFF
--- a/packages/l/libxss/xmake.lua
+++ b/packages/l/libxss/xmake.lua
@@ -35,5 +35,5 @@ package("libxss")
     end)
 
     on_test(function (package)
-        assert(package:has_ctypes("XScreenSaverInfo", {includes = "X11/extensions/scrnsaver.h"}))
+        assert(package:has_cfuncs("XScreenSaverQueryInfo", {includes = "X11/extensions/scrnsaver.h"}))
     end)


SirLynix:
iOS fails to build because SDL uses `GCEventInteraction` (introduced with iOS 18) which isn't supported by macos-14.

I think the code should be patched to fix compilation on lower platforms, anyone familiar with macos/iOS to help?

luadebug:
Does anything change if you replace macos-14 to macos-15 for iOS runner?

SirLynix:
> Does anything change if you replace macos-14 to macos-15 for iOS runner?

Probably, macos-14 includes XCode 16 which is the minimal required version for libsdl but it seems to default to XCode 15.4. I guess macos-15 would have the right default version.

However thanks to someone in the SDL repository (see the linked issue) it was easy to fix compilation on XCode < 16

luadebug:
> > Does anything change if you replace macos-14 to macos-15 for iOS runner?
> 
> Probably, macos-14 includes XCode 16 which is the minimal required version for libsdl but it seems to default to XCode 15.4. I guess macos-15 would have the right default version.
> 
> However thanks to someone in the SDL repository (see the linked issue) it was easy to fix compilation on XCode < 16

Maybe worth of having two iOS runners like macos-14 and macos-15 if that is the case. Although would be nice to upstream **fix compilation on XCode < 16** I guess.

packages/l/libxss/xmake.lua
```diff
@@ -0,0 +1,39 @@
...
+            table.insert(configs, "--with-pic")
+        end
+        if package:is_cross() then
+            table.insert(configs, "--disable-malloc0returnsnull")
+        end
+        import("package.tools.autoconf").install(package, configs)
+    end)
+
+    on_test(function (package)
+        assert(package:has_ctypes("XScreenSaverInfo", {includes = "X11/extensions/scrnsaver.h"}))
```
waruqi: we should check cfunc, it will check links.

SirLynix: Done

PR #8965, opened 2026-01-01
ncnn: clear versions

apocelipes:
Delete the duplicated version and correct version names:

image

### DIFF
--- a/packages/n/ncnn/xmake.lua
+++ b/packages/n/ncnn/xmake.lua
@@ -4,18 +4,10 @@ package("ncnn")
     set_license("BSD-3-Clause")
 
     add_urls("https://github.com/Tencent/ncnn/archive/refs/tags/$(version).tar.gz",
-             "https://github.com/Tencent/ncnn.git", {
-        submodules = false,
-        version = function(version)
-            if version then
-                return version:gsub("%.", "")
-            end
-        end
-    })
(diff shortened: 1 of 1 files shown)


packages/n/ncnn/xmake.lua
```diff
@@ -4,18 +4,10 @@ package("ncnn")
...
-                return version:gsub("%.", "")
-            end
-        end
-    })
+            "https://github.com/Tencent/ncnn.git", { submodules = false })
 
     add_versions("20250916", "7d463f1e5061facd02b8af5e792e059088695cdcfcc152c8f4892f6ffe5eab1a")
-    add_versions("2025.09.16", "7d463f1e5061facd02b8af5e792e059088695cdcfcc152c8f4892f6ffe5eab1a")
-    add_versions("2025.05.03", "3afea4cf092ce97d06305b72c6affbcfb3530f536ae8e81a4f22007d82b729e9")
+    add_versions("20250503", "3afea4cf092ce97d06305b72c6affbcfb3530f536ae8e81a4f22007d82b729e9")
```
waruqi, line 10: we should remove 20250916 and 20250503

apocelipes: That would cause conflicts with autoupdate.lua.

waruqi: ok

PR #8971, opened 2026-01-02
gpm: add new package

luadebug:
Awww, well I am not lucky when it comes to work with Linux libraries.
Fedora looks like completely empty Linux OS without anything... so it took a little while.

### DIFF
--- a/packages/g/glm/xmake.lua
+++ b/packages/g/glm/xmake.lua
@@ -7,6 +7,7 @@ package("glm")
              {version = function(version) return version:gsub("%+", ".") end})
     add_urls("https://github.com/g-truc/glm.git")
 
+    add_versions("1.0.3", "6775e47231a446fd086d660ecc18bcd076531cfedd912fbd66e576b118607001")
     add_versions("1.0.2", "19edf2e860297efab1c74950e6076bf4dad9de483826bc95e2e0f2c758a43f65")
     add_versions("1.0.1", "9f3174561fd26904b23f0db5e560971cbf9b3cbda0b280f04d5c379d03bf234c")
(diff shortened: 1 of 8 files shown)


packages/g/gpm/xmake.lua
```diff
@@ -0,0 +1,63 @@
...
+            envs.ACLOCAL_PATH = path.join(libtool:installdir("share"), "aclocal")
+        end
+        local configs = {}
+        table.insert(configs, "--prefix=" .. path.unix(package:installdir()))
+        table.insert(configs, 'CFLAGS=-std=c17')
+        if package:is_debug() then
+            table.insert(configs, "--enable-debug")
+        end
+        os.vrunv("./autogen.sh", {shell = true}, {envs = envs})
+        os.vrunv("./configure", configs, {shell = true})
```
waruqi: why do not use autoconf.install? it will call autogen.sh too.

luadebug: OK, I removed autogen.sh and configure parts.

packages/g/gpm/xmake.lua
```diff
@@ -0,0 +1,68 @@
...
+        os.cd(libdir)
+        if package:config("shared") then
+            local files = os.files("libgpm.so.*")
+            if #files > 0 then
+                os.ln(path.filename(files[1]), "libgpm.so")
+            end
+            os.tryrm("libgpm.a")
+        else
+            os.tryrm("libgpm.so*")
+        end
```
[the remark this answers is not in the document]
luadebug: @gemini-code-assist There is no os.scope

PR #8979, opened 2026-01-03
vk-bootstrap: Added missing header file, "VkBootstrapFeatureChain.h".

bulganonur:


luadebug (approved):
This should resolve on_test failure.

luadebug:
Maybe that would require update all vulkan related packages to succeed?

PR #9027, opened 2026-01-09
shared-mime-info: add 2.4 version

star-hengxing:


luadebug:
Probably should work on Linux machine but something off when it encounters CI?

luadebug:
Does libxml2 package provides binary xmllint? If so we should add bin to PATH, but as I remember it does not install xmllint into bin folder...

PR #9044, opened 2026-01-10
improve cmake to support for schemes

waruqi:
wait for https://github.com/xmake-io/xmake/pull/7189


luadebug:
Oh congrats :DDD

SirLynix:
It's a very nice feature but cumbersome to write, especially when support for older xmake is required.

Maybe splitting schemes in multiple files would be possible while keeping retrocompatibility?

waruqi:
> It's a very nice feature but cumbersome to write, especially when support for older xmake is required.
> 
> Maybe splitting schemes in multiple files would be possible while keeping retrocompatibility?

like this?

SirLynix:
> like this?

yes I find it more readable to not mix the scheme code

PR #9045, opened 2026-01-10
ftgl: fix windows shared build

star-hengxing:


luadebug:
There are probably two more names to check via pkg_search_module like `FREETYPE` and `FREETYPE2`.

packages/f/ftgl/xmake.lua
```diff
@@ -3,19 +3,22 @@ package("ftgl")
...
-    end
-
     add_deps("cmake")
-    add_deps("freetype", "zlib")
+    if is_subhost("windows") then
+        add_deps("pkgconf")
+    else
+        add_deps("pkg-config")
+    end
+    add_deps("freetype", {configs = {zlib = true}})
```
luadebug, line 21: zlib is on by default though?

luadebug:
Thanks for the fix.

PR #9056, opened 2026-01-11
glaze: update to v7.0.0

qudix:
- there are several versions in between


qudix:
freebsd ci failure is unrelated

PR #9089, opened 2026-01-13
criterion: update to 2.4.3

luadebug:
Very old one 8 months ig.

https://github.com/xmake-io/xmake-repo/pull/8415

### DIFF
--- a/packages/c/criterion/xmake.lua
+++ b/packages/c/criterion/xmake.lua
@@ -35,12 +35,11 @@ package("criterion")
     end)
 
     on_install("windows|!arm*", "linux", "macosx", "cross", "mingw@windows,msys", "bsd", "msys", function (package)
-        import("lib.detect.find_tool")
         os.rm("subprojects")
         import("patch")(package)
         local opt = {}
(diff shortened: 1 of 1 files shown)


packages/c/criterion/xmake.lua
```diff
@@ -33,23 +35,26 @@ package("criterion")
...
 
     on_install("windows|!arm*", "linux", "macosx", "cross", "mingw@windows,msys", "bsd", "msys", function (package)
+        import("lib.detect.find_tool")
         os.rm("subprojects")
         import("patch")(package)
         local opt = {}
-        --    Gather protoc-gen-nanopb from python3 pip
-        local python = package:is_plat("windows") and "python" or "python3"
-        os.vrun(python .. " -m pip install protobuf==5.29.3 nanopb==0.4.9.1")
+        local python = find_tool("python")
```
waruqi: we need not to find it, python has been added $PATH, we can run it directly.

luadebug: Done, let me know if there is need further changes.

PR #9122, opened 2026-01-16
Add boost math standalone

irenier:
Separate boost math from boost

### DIFF
--- a/packages/b/boost_math/xmake.lua
+++ b/packages/b/boost_math/xmake.lua
@@ -1,6 +1,6 @@
 package("boost_math")
     set_kind("library", {headeronly = true})
-    set_homepage("http://boost.org/libs/math")
+    set_homepage("https://www.boost.org/libs/math")
     set_description("Boost Math Library")
     set_license("BSL-1.0")
 


packages/b/boost_math/xmake.lua
```diff
@@ -0,0 +1,40 @@
+package("boost_math")
+    set_kind("library", {headeronly = true})
+    set_homepage("http://boost.org/libs/math")
+    set_description("Boost Math Library")
+    set_license("BSL-1.0")
+
+    add_defines("BOOST_MATH_STANDALONE")
```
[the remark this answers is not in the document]
irenier: https://github.com/boostorg/math?tab=readme-ov-file#standalone-mode

packages/b/boost_math/xmake.lua
```diff
@@ -0,0 +1,40 @@
+package("boost_math")
+    set_kind("library", {headeronly = true})
+    set_homepage("https://www.boost.org/libs/math")
+    set_description("Boost Math Library")
+    set_license("BSL-1.0")
+
+    add_defines("BOOST_MATH_STANDALONE")
```
luadebug, line 7: I am unsure that would work.
Maybe better to place it like:
```lua
    on_load(function (package)
        package:add("defines", "BOOST_MATH_STANDALONE")
    end)
```

SirLynix: It's the same

waruqi:
Why not use `add_requires("boost", {configs = {math = true}})` ? Is it not in boost?

We only maintain a single boost package and enable them via configs. Too many separate boost packages would increase maintenance workload and complexity.

irenier:
> Why not use `add_requires("boost", {configs = {math = true}})` ? Is it not in boost?
> 
> We only maintain a single boost package and enable them via configs. Too many separate boost packages would increase maintenance workload and complexity.

This is a standalone version of Boost Math, independent of the monolithic Boost project; it is analogous to [boost_pfr](https://github.com/xmake-io/xmake-repo/blob/dev/packages/b/boost_pfr/xmake.lua).

PR #9134, opened 2026-01-17
Update Vulkan packages to 1.4.335+0

SirLynix:
#9062

PR #9144, opened 2026-01-19
Adjust scheme handling for BSD platform

luadebug:
Just test. #9134
It is known fact that freebsd uses source but never uses binary for cmake package.

As we see it builds from source and works now. As FreeBSD doesn't try to use binary.

```
cmake --version
  cmake version 4.2.1
  
  CMake suite maintained and supported by Kitware (kitware.com/cmake).
    => install cmake 4.2.1 (source) .. ok
```

### DIFF
--- a/packages/c/cmake/xmake.lua
+++ b/packages/c/cmake/xmake.lua
@@ -12,17 +12,25 @@ package("cmake")
     end
 
     if add_schemes then
-        add_schemes("binary", "source")
+        if is_plat("bsd") then
+            add_schemes("source")
+        else
+            add_schemes("binary", "source")
+        end
     end
 
     on_source(function (package)
-        import("schemes.binary")
+        if not package:is_plat("bsd") then
+            import("schemes.binary")
+        end
         import("schemes.source")
         if package.current_scheme then
-            binary.add_urls(package, "binary")
+            if not package:is_plat("bsd") then
+                binary.add_urls(package, "binary")
+            end
             source.add_urls(package, "source")
         else
-            if binary.add_urls(package) then
+            if not package:is_plat("bsd") and binary.add_urls(package) then
                 package:data_set("scheme", "binary")
             else
                 source.add_urls(package)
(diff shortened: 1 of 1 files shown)


packages/c/cmake/xmake.lua
```diff
@@ -12,17 +12,25 @@ package("cmake")
     end
 
     if add_schemes then
-        add_schemes("binary", "source")
```
waruqi, line 15: We don't need to modify it; xmake will automatically ignore empty schemes if no URLs and versions are added to the binary scheme.

I did this specifically to simplify package configuration.

packages/c/cmake/xmake.lua
```diff
@@ -12,17 +12,25 @@ package("cmake")
...
+            add_schemes("source")
+        else
+            add_schemes("binary", "source")
+        end
     end
 
     on_source(function (package)
-        import("schemes.binary")
+        if not package:is_plat("bsd") then
+            import("schemes.binary")
```
waruqi, line 24: We can always import it, even if it's not used.

packages/c/cmake/xmake.lua
```diff
@@ -12,17 +12,25 @@ package("cmake")
...
+        end
         import("schemes.source")
         if package.current_scheme then
-            binary.add_urls(package, "binary")
+            if not package:is_plat("bsd") then
+                binary.add_urls(package, "binary")
+            end
```
waruqi, line 30: If it's a BSD, `binary.add_urls` won't add anything even if it's called, so there's no need to distinguish between them.

`add_urls` already handles platform detection internally, so we should process the BSD within it.

packages/c/cmake/xmake.lua
```diff
@@ -12,17 +12,25 @@ package("cmake")
...
             source.add_urls(package, "source")
         else
-            if binary.add_urls(package) then
+            if not package:is_plat("bsd") and binary.add_urls(package) then
```
waruqi, line 33: if bsd, binary.add_urls will return nil, we need not modify it.

packages/c/cmake/xmake.lua
```diff
@@ -42,10 +50,12 @@ package("cmake")
...
 
     on_install("@macosx", "@linux", "@windows", "@msys", "@cygwin", "@bsd", function (package)
-        import("schemes.binary")
+        if not package:is_plat("bsd") then
+            import("schemes.binary")
+        end
         import("schemes.source")
         local scheme_name = package.current_scheme and package:current_scheme():name() or package:data("scheme")
-        if scheme_name == "binary" then
+        if not package:is_plat("bsd") and scheme_name == "binary" then
```
waruqi, line 58: For the BSD platform, the binary scheme is an empty scheme; it has been ignored and will not be executed until on_install.

waruqi:
I have fixed in xmake.

PR #9153, opened 2026-01-21
fix: do not disable multi-threaded support defaultly for openssl3

zjyhjqs:
see: https://github.com/openssl/openssl/blob/master/INSTALL.md#notes-on-multi-threading

Note that `OPENSSL_NO_HEARTBEATS` no longer exists in OpenSSL 3.x

### DIFF
--- a/packages/o/openssl3/xmake.lua
+++ b/packages/o/openssl3/xmake.lua
@@ -34,6 +34,8 @@ package("openssl3")
 
     -- https://security.stackexchange.com/questions/173425/how-do-i-calculate-md2-hash-with-openssl
     add_configs("md2", {description = "Enable MD2 on OpenSSl3 or not", default = false, type = "boolean"})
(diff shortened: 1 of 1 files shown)


packages/o/openssl3/xmake.lua
```diff
@@ -223,9 +223,7 @@ package("openssl3")
 
         local target = target_plat .. "-" .. target_arch
         local configs = {target,
-                         "-DOPENSSL_NO_HEARTBEATS",
```
apocelipes, line 226: It's better to add an option that controls whether openssl3 building with multi-thread or not.

We can add an option like this:

https://github.com/xmake-io/xmake-repo/blob/a99acc216c4bd0b3cd12e994d9c01e9ab289b629/packages/o/openssl3/xmake.lua#L36

Then we need to decide whether to remove those compilation flags based on the values of the option.

zjyhjqs: done

apocelipes:
LGTM.

PR #9173, opened 2026-01-23
pinocchio: add package

star-hengxing:


star-hengxing:
```
Error: Process completed with exit code 143.
```
linux gcc out of memory xD.

PR #9186, opened 2026-01-25
notcurses: add new package

Willaaaaaaa:
https://github.com/dankamongmen/notcurses

for https://github.com/xmake-io/xmake-repo/issues/10#issuecomment-3795872018


WangZhongDian:
The `libunistring` you rely on does not support Windows

PR #9251, opened 2026-02-02
zimg: add more platform support

star-hengxing:


packages/z/zimg/port/xmake.lua
```diff
@@ -0,0 +1,37 @@
...
+
+set_languages("c++17")
+
+local dirs = os.dirs("src/zimg/*")
+
+target("zimg")
+    set_kind("$(kind)")
+    add_includedirs("src/zimg")
+    for _, dir in ipairs(dirs) do
+        add_files(path.join(dir, "*.cpp"))
```
waruqi: 为啥不直接用 `add_files("src/zimg/*/*.cpp")`

PR #9273, opened 2026-02-05
node_crunch2: add package

willi-kappler:


willi-kappler:
Thanks for improve and fix, what else is needed ?
The built action is failing since it is not meant to be used on mobile devices.

luadebug:
> Thanks for improve and fix, what else is needed ?
> The built action is failing since it is not meant to be used on mobile devices.

Well maybe either limit arm64 macos/ios or either wait some support patch as it seems work for x86_64 macos.
For Windows OS there seems some symbols missing inside on test so maybe its better to test another symbols at case windows os.
Well certainly we would limit wasm as dependencies are not supporting wasm.
For android we can place expected header check in on_check handler. So it would filter ndk that does not provide expected headerfile...

willi-kappler:
Thanks for your work @luadebug!
I'll patch my build configuration accordingly for future versions and will also add a GitHub action for various platforms.

willi-kappler:
"expected" is available since C++23

PR #9367, opened 2026-02-19
fix(cpr)!: ssl setup

gibbz00:
The previous xmake.lua conflates SSH and SSL setup. Adding both ssh dependencies (libbssh2 and zlib) and CPR_ENABLE_SSL when setting the `ssl` config option to true.

`cpr` is just a wrapper for crl, and does not make use of curls subdependencies. As such, this commit removes the ssl config option and the libssh2 and zlib dependencies that came with it. Users whishing to extend the curl config, should instead use `add_requireconfs("cpr.libcurl", {configs = {...}})`.

The exception to this is openssl. The `libcurl` xmake.lua will always enable some form of SSL, it can't be disabled [libcurl_ssl]. This commit also makes sure that an openssl depedendency is always included. It doesn't make much sense to set CPR_ENABLE_SSL to false if it is always supported by the `libcurl` configuration.

`openssl3` is set by default, but it can be downgraded to `openssl` (1.1.1) with the `downgrade_openssl` configuration option. This option is then also propagated to `libcurl`.

[libcurl_ssl]: https://github.com/xmake-io/xmake-repo/blob/f06c50cde91716d6e4068bc7f295890f616d4f35/packages/l/libcurl/xmake.lua#L43-L51


packages/c/cpr/xmake.lua
```diff
@@ -38,12 +38,9 @@ package("cpr")
...
 
     on_load(function (package)
-        if package:config("ssl") then
-            package:add("deps", "libcurl", {configs = {libssh2 = true, zlib = true}})
-            package:add("deps", "libssh2")
-        else
-            package:add("deps", "libcurl")
-        end
+        local downgrade_ssl = package:config("downgrade_openssl")
+        package:add("deps", "libcurl", {configs = {openssl = downgrade_ssl})
```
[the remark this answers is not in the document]
gibbz00: This is just the AI hallucinating by not understanding that `openssl = false` is used to select openssl3 over openssl on linux.

https://github.com/xmake-io/xmake-repo/blob/f06c50cde91716d6e4068bc7f295890f616d4f35/packages/l/libcurl/xmake.lua#L43-L51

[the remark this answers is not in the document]
gibbz00: My bad, missed that `nil != false`.

packages/c/cpr/xmake.lua
```diff
@@ -18,7 +18,7 @@ package("cpr")
     add_versions("1.7.2", "aa38a414fe2ffc49af13a08b6ab34df825fdd2e7a1213d032d835a779e14176f")
     add_versions("1.6.2", "c45f9c55797380c6ba44060f0c73713fbd7989eeb1147aedb8723aa14f3afaa3")
 
-    add_configs("ssl", {description = "Enable SSL.", default = false, type = "boolean"})
+    add_configs("downgrade_openssl", {description = "openssl package instead of openssl3", default = false, type = "boolean"})
```
waruqi, line 23: Please do not arbitrarily modify existing package configuration names, as this will disrupt existing users' projects.

You can add new configuration names and mark the SSL configuration description as deprecated, then output a warning message.

gibbz00: @waruqi  Hey again,

Sorry for the late reply (and for the ping if it was inappropriate). 

I've now made sure to not make any breaking changes to the previous 'ssl' config option, but instead mark it as deprecated.

packages/c/cpr/xmake.lua
```diff
@@ -40,8 +41,23 @@ package("cpr")
 
     on_load(function (package)
         if package:config("ssl") then
+            wprint([[
+                wprint("package(cpr): config 'ssl' is deprecated.
```
waruqi: ??

gibbz00: Woopsie, thanks.

PR #9411, opened 2026-02-26
Added renderdoc package

ghost:
Added renderdoc package (only version 1.42 for now) to allow use of the In-Application API to do custom captures of graphics or headless applications based on graphics API like Vulkan, OpenGL or Direct3D.


ghost:
FYI, the unverified commits are there because I forgot to update my GPG key. Next ones will be signed.

packages/r/renderdoc/xmake.lua
```diff
@@ -0,0 +1,60 @@
+package("renderdoc")
+    set_homepage("https://renderdoc.org/")
+    set_description("RenderDoc In-Application API")
```
Willaaaaaaa, line 3: plz use its official github description.

packages/r/renderdoc/xmake.lua
```diff
@@ -0,0 +1,60 @@
...
+    if is_plat("windows") and is_arch("x64") then
+        add_urls("https://renderdoc.org/stable/$(version)/RenderDoc_$(version)_64.zip")
+        add_versions("1.42", "1448d896d43904f50c1496df9a8d015553e0449ee107bac9b97359f2dffe373d")
+    elseif is_plat("windows") and is_arch("x86") then
+        add_urls("https://renderdoc.org/stable/$(version)/RenderDoc_$(version)_32.zip")
+        add_versions("1.42", "dcdd159c53cd8ec1d31d562150e49d927734a01297375e0876949347e1c9a09e")
+    else
+        add_urls("https://renderdoc.org/stable/$(version)/renderdoc_$(version).tar.gz")
+        add_versions("1.42", "25e66e1d35bff7a31a139f2c0b47f510947e782859dbfbbfd0450eca60bc16d2")
+    end
```
Willaaaaaaa, line 18:
```suggestion
    add_urls("https://github.com/baldurk/renderdoc/archive/refs/tags/$(version).tar.gz")
    add_versions("xxx", "sha256sum")
```
We should stick to using its source code to build instead of packaged files.

packages/r/renderdoc/xmake.lua
```diff
@@ -0,0 +1,60 @@
...
+        os.cp("renderdoc_app.h", package:installdir("include"))
+        os.cp("lib/librenderdoc.dll", package:installdir("lib"))
+    end)
+
+    on_install("linux", function (package)
+        os.cp("include/renderdoc_app.h", package:installdir("include"))
+        os.cp("lib/librenderdoc.so", package:installdir("lib"))
+    end)
+
+    on_test(function (package)
```
Willaaaaaaa, line 33: Simplify test code plz.

packages/r/renderdoc/xmake.lua
SirLynix: Note that now xmake supports scheme to support both

Willaaaaaaa: > Note that now xmake supports scheme to support both

oh, yes! perfectly true. forgot about that.

PR #9216, opened 2026-01-29
openal-soft: update to v1.25.1

apocelipes:
Fixes #8950 
Fixes #8936 
Fixes #6262 
Fixes #6019

PR #9712, opened 2026-04-11
openal-soft: update to 1.25.1

rempas:
Added openal-soft versions 1.25.0 and 1.25.1. I also added conditional code to check for minimal c++ compiler support which didn't exist before.

### DIFF
--- a/packages/o/openal-soft/xmake.lua
+++ b/packages/o/openal-soft/xmake.lua
@@ -67,16 +67,5 @@ package("openal-soft")
     end)
 
     on_test(function (package)
-        if package:version():ge("1.25.0") then
-            local required_cpp_ver = "c++20"
-        elseif  package:version():ge("1.24.0") then
-            local required_cpp_ver = "c++17"
-        else
-            local required_cpp_ver = "c++11"
-        end
-
-        assert(package:has_cfuncs("alGetProcAddress",
-            {configs = { languages = required_cpp_ver}},
-            {includes = "AL/al.h"}
-        ))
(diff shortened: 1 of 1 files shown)


SirLynix:
OpenAL soft doesn't require C++ to use it, it's a C API, it only requires C++ to build it

rempas:
> OpenAL soft doesn't require C++ to use it, it's a C API, it only requires C++ to build it

In the docs, it isn't clear if the "languages" option, is only used for building (tho, that is what I assume). Deepwiki says that it's only for building the package.

SirLynix:
> > OpenAL soft doesn't require C++ to use it, it's a C API, it only requires C++ to build it
> 
> In the docs, it isn't clear if the "languages" option, is only used for building (tho, that is what I assume). Deepwiki says that it's only for building the package.

In the on_test callback the library is already built, it's only testing an example code, so it can be built with C

rempas:
> In the on_test callback the library is already built, it's only testing an example code, so it can be built with C

I see! I'm an idiot, I'm sorry for wasting your time. Thank you for letting me know, I fixed it! It still fails but it seems to be errors in the source code of openal-soft. I don't know at this point...

apocelipes:
There are too many compiler errors with Clang in this version. Maybe we should wait for the next version. See also #9216.

rempas:
Yeah, I understand. I will open an issue in the openal-soft repo to let the maintainer(s) know about that and hopefully make it work properly with clang.

PR #9730, opened 2026-04-12
github actions: limit build actions to package directory

burdockcascade:
this limits the build actions to only run when changes are made to the packages directory


waruqi:
？What problem is this trying to solve?

PR #9731, opened 2026-04-12
github actions: minor cron fix

burdockcascade:
this is a minor fix to the cron schedule in the github actions


waruqi:
This seems to make no difference.

PR #9747, opened 2026-04-14
Add support for LLVM version 21.1.0

ArthurVasseur:


packages/l/libllvm/xmake.lua
```diff
@@ -39,6 +43,9 @@ package("libllvm")
         add_versions("tarball:19.1.7", "82401fea7b79d0078043f7598b835284d6650a75b93e64b6f761ea7b63097501")
         add_versions("git:19.1.7", "llvmorg-19.1.7")
 
+        add_versions("tarball:21.1.0", "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855")
+        add_versions("git:21.1.0", "llvmorg-21.1.0")
```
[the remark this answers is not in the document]
ArthurVasseur: ?? no

ArthurVasseur:
The ubuntu build is passing if I am using [act](https://github.com/nektos/act)
image

The CI is not powerful enough

.github/workflows/iphoneos.yml
```diff
@@ -25,4 +25,4 @@ jobs:
 
       - name: Tests
         run: |
-          xmake l ./scripts/test.lua -D -p iphoneos
+          xmake l ./scripts/test.lua -j1 -D -p iphoneos
```
waruqi: We don't need to modify `ci`; we can determine if the current job is on `ci` and limit the number of jobs. 

[https://github.com/xmake-io/xmake-repo/blob/971d99d38d2efc6c18a2494b9541847b91d5e799/packages/p/pinocchio/xmake.lua#L138-L139](https://github.com/xmake-io/xmake-repo/blob/971d99d38d2efc6c18a2494b9541847b91d5e799/packages/p/pinocchio/xmake.lua#L138-L139)

waruqi:
> The ubuntu build is passing if I am using [act](https://github.com/nektos/act) image
> 
> The CI is not powerful enough

Due to CI's resource limitations, I will manually merge it once your local testing and verification are successful.

.github/workflows/iphoneos.yml
ArthurVasseur: okay thanks

ArthurVasseur:
FreeBSD static build is working
image

ArthurVasseur:
Could not test on Apple-based systems or ARM-based operating systems

luadebug:
Maybe we can try `macos-15` for ios builder but I am not sure it would be faster. As it says M1/M2 it would probably be same M1 as `macos-14` and 7GB ram.
Placing -j1 directly inside of YML workflow for iOS to avoid OOM did not help, it still occur OOM. :(

ArthurVasseur:
What should we do here?

luadebug:
Maybe we can improve existing runners but this would be very hard to achieve as we can try increase SWAP size  https://github.com/marketplace/actions/maximize-build-disk-space

PR #9761, opened 2026-04-16
openssl3: add versions 3.6.2, 3.5.6, 3.4.5, 3.3.7, 3.0.20

zjyhjqs:
Note: Where to place OpenSSL 4.x? (4.0.0 is released)


SirLynix:
> Note: Where to place OpenSSL 4.x? (4.0.0 is released)

I see two options:
- making an openssl4 package
- renaming the openssl3 package to opensslnext or something and add versions 3/4/...

I advise we follow what Linux package managers do

PR #9795, opened 2026-04-19
aui: add android and ios support

luadebug:
Add android and ios support for https://github.com/aui-framework/aui/issues/522


luadebug:
During gtk3 library build two cairo libraries present. Linux OS host and Android NDK so it cause symbols duplication. I lack of idea how to workaround upon this situation as well as I have not issues on my Linux OS  PC.

luadebug:
It looks like removing it from packagedeps has helped to advance further.

PR #9869, opened 2026-04-26
fix: find OpenGL by correct name on Linux

xq114:


xq114:
fetch-only package, tested on local Debian 13

PR #9878, opened 2026-04-27
Add reflect package

BobWorts:


packages/r/reflect/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+        local tc = package:is_plat("windows") and package:toolchain("msvc")
+        if not tc then return end
+
+        local toolset = tc:config("vs_toolset")
+        if not toolset then return end
+
+        local major, minor = toolset:match("^(%d+)%.(%d+)")
+        if not major or not minor then return end
+
+        if tonumber(major) == 14 and tonumber(minor) >= 50 then
```
waruqi: I think we should use `#if defined(_MSC_VER) && _MSC_VER >= 1950` in patch code.

luadebug: I am just confused of two first lines of patch does it apply?

BobWorts: @waruqi Yes, that would be the better solution

BobWorts: @luadebug If you're confused about the missing file extension, it simply doesn't have one.

Shiffted:
The package already exists as [`boost_reflect`](https://github.com/xmake-io/xmake-repo/blob/dev/packages/b/boost_reflect/xmake.lua), however they moved the repo to qlibs/reflect. Maybe we can update the `boost_reflect` package similar to the old [`libsdl`](https://github.com/xmake-io/xmake-repo/blob/dev/packages/l/libsdl/xmake.lua) package and then use this one instead.

waruqi:
> The package already exists as [`boost_reflect`](https://github.com/xmake-io/xmake-repo/blob/dev/packages/b/boost_reflect/xmake.lua), however they moved the repo to qlibs/reflect. Maybe we can update the `boost_reflect` package similar to the old [`libsdl`](https://github.com/xmake-io/xmake-repo/blob/dev/packages/l/libsdl/xmake.lua) package and then use this one instead.

right

PR #9883, opened 2026-04-28
Add vamp-plugin-sdk to xrepo

### LINKED ISSUE #9831: [New Package]: vamp-sdk
### Action Type

New Package

### Package Name

vamp-sdk

### Version(s)

v2.10.2

### Project Repository

https://github.com/vamp-plugins/vamp-plugin-sdk

### Project Homepage

https://vamp-plugins.org

### License

X11 AND BSD-3-Clause (I'm not sure)

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Android
- [ ] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD

Upstream uses CMake.
Dependencies: libsndfile (optional)

### Additional Context

_No response_

EZ4Stephen:
-For version, uses the latest commit instead of a tagged version. Last tag was in 2020, CMake support was added in 2021, to support most platforms. So to make it buildable on more platforms, I've used the latest commit's hash instead.

-No options used
vamp-plugin-sdk uses 3 CMake options and all 3 are off by default, and for most cases, no need to define any of them.

-One string replacement
Removes -ldl for windows in https://github.com/vamp-plugins/vamp-plugin-sdk/blob/master/pkgconfig/vamp-hostsdk.pc.in. Not sure which other build platform options need it!

Initially thought it should be called vamp-sdk, but seems like in most places it's named vamp-plugin-sdk, hence name mismatch compared to the initial request.

This closes #9831 .

### DIFF
--- a/packages/v/vamp-plugin-sdk/xmake.lua
+++ b/packages/v/vamp-plugin-sdk/xmake.lua
@@ -1,0 +1,23 @@
+package("vamp-plugin-sdk")
+    set_homepage("https://www.vamp-plugins.org")
+    set_description("The SDK for Vamp plugins, an API for audio analysis and feature extraction plugins.")
+    set_license("BSD-3-Clause AND MIT")
+
+    add_urls("https://github.com/vamp-plugins/vamp-plugin-sdk/archive/refs/tags/vamp-plugin-sdk-v$(version).tar.gz",
+             "https://github.com/vamp-plugins/vamp-plugin-sdk.git")
+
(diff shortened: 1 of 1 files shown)


EZ4Stephen:
Ah, I have to fix it to build as a shared library on Windows.

EZ4Stephen:
Are the FreeBSD tests supposed to take that long?

luadebug:
> Are the FreeBSD tests supposed to take that long?

I think building cmake from source takes very long time, but it looks like required option.

PR #9910, opened 2026-05-04
MINGW32 support

luadebug:
MSYS drops MINGW32 packages such as pkgconfig and now as we see fortran has failed to fetch as it got removed recently for 16th gcc... Downgrade to workaround this. Or either just drop MINGW32


waruqi:
Is MSYS no longer maintaining MINGW32?

luadebug:
https://github.com/msys2/MINGW-packages/pull/28751

This pr removes i686 mingw32 support for fortran? Or what exactly happen?

waruqi:
> msys2/MINGW-packages#28751

but it seems be reverted? https://github.com/msys2/MINGW-packages/commit/2ac1118c890bff61a951455e72b75b817bd4b6ed

waruqi:
Regardless, we can first use the current workaround and observe it for a period of time.

PR #9942, opened 2026-05-09
Add geo-utils-cpp v1.0.1

gistrec:
Adds [geo-utils-cpp](https://github.com/gistrec/geo-utils-cpp) — a header-only C++17 library for spherical (lat/lng) geometry on Earth coordinates: distance, bearing, polygon area, point-in-polygon, and path proximity.

- License: Apache-2.0
- Header-only, no runtime dependencies
- Headers under `include/geo/` (entry point: `<geo/geo.hpp>`)
- Already published in [microsoft/vcpkg](https://github.com/microsoft/vcpkg/tree/master/ports/geo-utils-cpp); Conan recipe in flight ([conan-io/conan-center-index#30152](https://github.com/conan-io/conan-center-index/pull/30152))

Install path is a straight `os.cp("include/geo", ...)` since the library is fully header-only — no CMake involvement needed for the consumer side.

Test snippet exercises the public API surface:
```cpp
#include <geo/geo.hpp>
void test() {
    geo::LatLng a{0.0, 0.0};
    geo::LatLng b{1.0, 1.0};
    volatile double d = geo::distance_between(a, b);
    (void)d;
}
```
Compiled with `c++17`.

### DIFF
--- a/packages/g/geo-utils-cpp/xmake.lua
+++ b/packages/g/geo-utils-cpp/xmake.lua
@@ -16,9 +16,10 @@ package("geo-utils-cpp")
 
     on_test(function (package)
         assert(package:check_cxxsnippets({test = [[
-            geo::LatLng a{0.0, 0.0};
-            geo::LatLng b{1.0, 1.0};
-            volatile double d = geo::distance_between(a, b);
-            (void)d;
+            void test() {
+                geo::LatLng a{0.0, 0.0};
+                geo::LatLng b{1.0, 1.0};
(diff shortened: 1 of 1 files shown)


gistrec:
All four review suggestions applied (force-pushed):

- `add_urls` now uses `v$(version)` and `add_versions` stores plain `1.0.1` without the `v` prefix
- `LICENSE` installed to `share/$(package:name())`
- Test snippet uses `includes = "geo/geo.hpp"` instead of inline `#include`/wrapper

PR #9955, opened 2026-05-11
Add use_cxx, override, asan configs to mimalloc

Akylzhan:


packages/m/mimalloc/xmake.lua
```diff
@@ -34,6 +34,9 @@ package("mimalloc")
     if is_plat("windows") then
         add_configs("etw", {description = "Enable Event tracing for Windows", default = false, type = "boolean"})
     end
+    add_configs("override", {description = "Override the standard malloc interface", default = false, type = "boolean"})
+    add_configs("use_cxx", {description = "Use the C++ compiler to compile the library", default = false, type = "boolean"})
+    add_configs("asan", {description = "Compile with address sanitizer support (adds a small overhead)", default = false, type = "boolean"})
```
waruqi: asan is a builtin config. we need not to define it.

packages/m/mimalloc/xmake.lua
```diff
@@ -34,6 +34,9 @@ package("mimalloc")
     if is_plat("windows") then
         add_configs("etw", {description = "Enable Event tracing for Windows", default = false, type = "boolean"})
     end
+    add_configs("override", {description = "Override the standard malloc interface", default = false, type = "boolean"})
+    add_configs("use_cxx", {description = "Use the C++ compiler to compile the library", default = false, type = "boolean"})
```
waruqi: remove `use_`, `with_` .. prefix

PR #9957, opened 2026-05-11
Fix JoltPhysics build for windows.

exdal:
Compile definitions used for this package must be defined in both target and the package itself. Currently, this package doesn't build on windows platform. See https://github.com/jrouwe/JoltPhysics/blob/f458a60722d106c0a6566f008a6e25ddd8002dd9/Jolt/RegisterTypes.cpp#L76


SirLynix:
that's weird, I'm using Jolt physics on Windows on a daily basis and have not seen this issue

Issue #9966, opened 2026-05-12
Generate a JSON package manifest to enable repology.org indexing


gistrec:
## Context

[repology.org](https://repology.org) aggregates package versions across many C/C++ package managers (vcpkg, Conan, Homebrew, distro repos) so library authors and users can see where a library is published and at which versions.

xmake-repo is currently **not** indexed by repology. A community contributor (@luadebug) has an open PR adding xrepo fetch support: **[repology/repology-updater#1585](https://github.com/repology/repology-updater/pull/1585)**. The PR is **blocked**, not on principle, but on three specific data gaps. The repology maintainer @AMDmi3 made it clear that parsing `xmake.lua` source directly is a non-starter (["We will not parse source code in any way"](https://github.com/repology/repology-updater/pull/1585#issuecomment-4276934489)) but [accepts indexing from a structured external source](https://github.com/repology/repology-updater/pull/1585#issuecomment-4280836489) if it exposes the following fields:

1. **`homepage`** must be exposed per package (currently dropped)
2. **Repository URL must be distinguishable from download URL** — no heuristics like `.git` suffix matching
3. **URLs must be concrete** — no `$(version)` interpolation, resolved against the latest version

## Proposal

Have xmake-repo CI publish a small `packages.json` manifest at a stable URL (e.g. a `gh-pages` branch, or a Release asset). Repology can then ingest directly.

Suggested minimal schema (one entry per package, latest version only — repology doesn't need history):

```json
{
  "generated_at": "2026-05-12T12:00:00Z",
  "packages": [
    {
      "name": "geo-utils-cpp",
      "version": "1.0.1",
      "description": "A tiny header-only C++17 latitude/longitude geometry library",
      "license": "Apache-2.0",
      "homepage": "https://github.com/gistrec/geo-utils-cpp",
      "repository_url": "https://github.com/gistrec/geo-utils-cpp",
      "download_url": "https://github.com/gistrec/geo-utils-cpp/archive/refs/tags/v1.0.1.tar.gz"
    }
  ]
}
```

Each field maps directly to an xmake.lua call (so generation is just sandboxed evaluation, which xmake already does on install):

| Field | Source in `xmake.lua` |
| --- | --- |
| `name` | package directory name |
| `version` | latest `add_versions(...)` entry |
| `description` | `set_description(...)` |
| `license` | `set_license(...)` |
| `homepage` | `set_homepage(...)` |
[...]

PR #9967, opened 2026-05-12
Add packages.json manifest generator (draft for #9966)

gistrec:
Companion draft PR for #9966 — adds a generator + CI workflow that emit a flat JSON index of all packages, intended for external consumers like [repology.org](https://repology.org/) (see [repology/repology-updater#1585](https://github.com/repology/repology-updater/pull/1585)).

Opening as **Draft** — the goal is to make the proposal in #9966 concrete enough to review, not to land as-is. Schema, output location, and CI trigger frequency are all open questions.

## What's in this PR

**`scripts/build_index.lua`** (~85 lines)

Iterates `packages/*/*` and loads each `xmake.lua` via the official `package.load_from_repository(...)` API — the same one used by `scripts/packages.lua` and `scripts/autoupdate.lua`. **No source parsing.** Reads metadata via `instance:get(...)`, resolves `$(version)` against the latest version, and writes `dist/repology_packages_index.json`.

Schema (one entry per package, latest version only):

```json
{
  "generated_at": "2026-05-12T14:17:00Z",
  "count": 1952,
  "packages": [
    {
      "name":           "geo-utils-cpp",
      "version":        "1.0.1",
      "description":    "Header-only C++17 library for spherical (lat/lng) geometry ...",
      "license":        "Apache-2.0",
      "homepage":       "https://github.com/gistrec/geo-utils-cpp",
      "repository_url": "https://github.com/gistrec/geo-utils-cpp",
      "download_url":   "https://github.com/gistrec/geo-utils-cpp/archive/refs/tags/v1.0.1.tar.gz"
    }
  ]
}
```

**`.github/workflows/sync.yml`** (+25 lines)

Per maintainer feedback, the index build is embedded into the existing hourly `sync.yml` workflow rather than living in a separate CI file. After the dev→master sync step, the new step generates `dist/repology_packages_index.json`, runs a sanity-check on the entry count (>1000) so a broken generator run never publishes an empty/truncated index, then pushes the file to the `gh-pages` branch via SSH (key already installed earlier in the same job, mirroring the `deploy.yml` pattern). Repology (or any other consumer) ingests from `https://xmake-io.github.io/xmake-repo/repology_packages_index.json`.

## Verification

Ran locally on the current `dev` tip:

```
$ xmake l scripts/build_index.lua
wrote dist/repology_packages_index.json (1952 packages)

$ python3 -c 'import json; d=json.load(open("dist/repology_packages_index.json")); ...'
total: 1952
unresolved $(version) in download_url: 0
missing download_url: 64        # mostly toolchain/system packages with no add_urls
missing homepage:     31        # real authoring gaps in individual packages
missing version:      56        # mostly toolchain/system packages
```

## Resolved during review

- **Publish target:** `gh-pages` branch (waruqi). Final URL: `https://xmake-io.github.io/xmake-repo/repology_packages_index.json`.
- **Trigger frequency:** consolidated into `sync.yml` (hourly cron), so the index updates at the same cadence as `dev`→`master` sync — no separate CI workflow.
- **`repository_url` derivation:** now derived from `add_urls` with forge-layout normalization (github / gitlab / bitbucket / codeberg / gitea / gitee, plus `/-/archive/` and `.git` suffixes accepted on any host); falls back to `homepage` only if the homepage is itself a recognized forge URL. No source parsing.

## Still open

- **Schema additions.** Tags, keywords, maintainer? Easy to add — happy to extend if there's interest.

## Related

- Coordination issue: #9966
- Upstream blocker on repology side: [repology/repology-updater#1585](https://github.com/repology/repology-updater/pull/1585)
- Specific repology requirements (AMDmi3): [comment](https://github.com/repology/repology-updater/pull/1585#issuecomment-4280836489)


scripts/build_index.lua
```diff
@@ -0,0 +1,126 @@
...
+            return versions[#versions]
+        end
+        for v, _ in pairs(versions) do
+            table.insert(list, tostring(v))
+        end
+    end
+    if #list == 0 then
+        return nil
+    end
+    table.sort(list)
```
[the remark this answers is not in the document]
gistrec: Fixed in 58ea4c6 — took the spirit of the suggestion but wrapped `semver.compare` in an explicit less-than function, since passing it directly errors out with `invalid order function for sorting` (Lua's `table.sort` requires a strict bool-returning comparator, but `semver.compare` returns -1/0/1). Also added a string-comparison fallback for the small number of packages whose version strings aren't valid semver (e.g. git refs), matching the pattern in `scripts/build_artifacts.lua`.

Sanity-checked on `dev`: 122 packages now report a corrected latest version — e.g. `aws-c-common` v0.9.3 → v0.12.6, `aom` 3.9.1 → 3.13.1, `botan` 3.8.1 → 3.12.0. The previous string-sort was actually wrong on a non-trivial slice of the repo.

scripts/build_index.lua
```diff
@@ -0,0 +1,126 @@
...
+    end
+    table.sort(entries, function(a, b) return a.name < b.name end)
+
+    local manifest = {
+        generated_at = os.date("!%Y-%m-%dT%H:%M:%SZ"),
+        count        = #entries,
+        packages     = entries,
+    }
+
+    os.mkdir("dist")
```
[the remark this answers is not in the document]
gistrec: Not applying — verified empirically on xmake 3.0.8 that `os.mkdir` is already idempotent (calling it on an existing directory is a no-op, not an error), and `os.mkpath` is `nil` in xmake's `os` module, so the suggested replacement wouldn't run.

```
$ xmake l - <<'LUA'
os.mkdir("/tmp/x"); print(os.exists("/tmp/x"))   --> true
os.mkdir("/tmp/x"); print(os.exists("/tmp/x"))   --> true  (no error on re-run)
print(os.mkpath)                                  --> nil
LUA
```

Leaving `os.mkdir("dist")` as-is.

waruqi:
> repository_url derivation. Currently mirrors homepage — works perfectly for ~95% of packages where the homepage is a github.com URL, less well for packages whose homepage is a project site (e.g. abseil's https://abseil.io). If repology needs a stricter repo URL, the long-term fix is a new xmake.lua field (set_source_url(...) or similar); the short-term workaround is acceptable for repology since it satisfies AMDmi3's "no .git heuristic" rule.

from add_urls 

https://github.com/xmake-io/xmake-repo/blob/02c06049ed114366d133c952da0afa62cc8e157e/packages/x/x264/xmake.lua#L6

waruqi:
> Trigger frequency. Currently on every push to master. Could downgrade to daily cron if push-rate is high enough to be noisy.

The dev to master is synchronized only once per hour.

gistrec:
@waruqi, stepping back a bit: what do you think about the overall direction of this PR?

The idea is to add a machine-readable JSON export for Repology, generated from `xmake-repo` package metadata. I proposed the current schema as a starting point, but I’m happy to adjust it if you prefer a different format or publishing mechanism

Before polishing the implementation further, I’d like to confirm:
- Is this kind of export acceptable for the repo?
- Is the proposed JSON schema roughly what you would expect?
- Should it be published via `gh-pages`, release assets, or something else?
- Should extra fields like `tags`, `maintainer`, and `repository_url` be part of the first version or left for later?

Happy to follow your preferred direction

waruqi:
> @waruqi, stepping back a bit: what do you think about the overall direction of this PR?
> 
> The idea is to add a machine-readable JSON export for Repology, generated from `xmake-repo` package metadata. I proposed the current schema as a starting point, but I’m happy to adjust it if you prefer a different format or publishing mechanism
> 
> Before polishing the implementation further, I’d like to confirm:
> 
> * Is this kind of export acceptable for the repo?
> * Is the proposed JSON schema roughly what you would expect?
> * Should it be published via `gh-pages`, release assets, or something else?
> * Should extra fields like `tags`, `maintainer`, and `repository_url` be part of the first version or left for later?
> 
> Happy to follow your preferred direction

You can publish to the gh-pages branch. If the index file is only used for Repology, it's best to rename it, for example: repology_packages_index.json

gistrec:
@waruqi done — renamed the output to `repology_packages_index.json` and updated the workflow accordingly (commit 6529bdb). Publish target stays on the `gh-pages` branch as you suggested, so once merged it'll be reachable at `https://xmake-io.github.io/xmake-repo/repology_packages_index.json`.

PR description is updated to match. Anything else you'd like adjusted before this can move out of draft?

.github/workflows/build-index.yml
```diff
@@ -0,0 +1,38 @@
+name: Build index
+
+# Generates a flat JSON index of all packages (dist/repology_packages_index.json) and
+# publishes it to the `gh-pages` branch. Consumed by external trackers
+# (e.g. repology.org) that need a structured view of the repository
+# without parsing xmake.lua source files. See xmake-io/xmake-repo#9966.
+
+on:
+  push:
```
gistrec: Maybe it's better to use CRON - run this workflow each N hours

.github/workflows/build-index.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: xmake-io/github-action-setup-xmake@v1
+        with:
+          xmake-version: latest
+
+      - name: Build repology_packages_index.json
+        run: xmake l scripts/build_index.lua
```
gistrec: Nice to have some check's, that new `repology_packages_index.json` isn't empty

.github/workflows/build-index.yml
waruqi: or you can add it in sync.yml

https://github.com/xmake-io/xmake-repo/blob/e5307faaf9cbd7b32bf76f25be1c6f478bd7964b/.github/workflows/sync.yml#L5

waruqi: I don't think it's necessary to do it in two ci.

gistrec:
Consolidated as you suggested — dropped `build-index.yml` entirely and appended the build+publish step to `sync.yml` after `automerge.lua` (commit 38eaa3e). One CI workflow now does dev→master sync **and** the repology index update at the same hourly cadence, so there's no second CI.

The new step also includes a small sanity-check (`assert d['count'] > 1000`) so a broken generator run never publishes an empty or truncated `repology_packages_index.json`. Publish is via SSH using the key already installed earlier in the same job (same pattern as `deploy.yml`), so no `permissions:` block changes were needed.

PR description is updated to reflect the consolidation. From my side this looks ready to leave draft — anything else you'd like adjusted?

PR #9968, opened 2026-05-12
libyuv: choose jpeg or jpeg-turbo

Akylzhan:


packages/l/libyuv/xmake.lua
```diff
@@ -14,7 +14,7 @@ package("libyuv")
     add_patches("1913", "patches/1913/cmake.patch", "9b61c6a5c26e727d164f06e83a3bf19863f840cd57fcee365429561e640930bf")
     add_patches("1891", "patches/1891/cmake.patch", "87086566b2180f65ff3d5ef9db7c59a6e51e2592aeeb787e45305beb4cf9d30d")
 
-    add_configs("jpeg", {description = "Build with JPEG.", default = false, type = "boolean"})
+    add_configs("jpeg", {description = "Build with JPEG.", default = "none", type = "string", values = {"none", "libjpeg", "libjpeg-turbo"}})
```
waruqi: We should not modify existing configurations, as this will break some user projects.

you can add a new config to select jpeg lib.

PR #9974, opened 2026-05-13
joltphysics: match upstream cmake for FPE definition

zachrammell:
### Problem  
The `joltphysics` package defines `JPH_FLOATING_POINT_EXCEPTIONS_ENABLED` for consumers when the platform is Windows, but Jolt's CMake only defines it when the compiler is MSVC. Under clang-cl, the library is built without the macro while consumers see it on, producing an ABI mismatch at `PhysicsSystem` init.
### Fix
Gate the package consumer define (and the matching `-D` passed to CMake) on `package:has_tool("cxx", "cl")` instead of platform, mirroring Jolt's check. MSVC unchanged; clang-cl on Windows now matches the library.

### DIFF
--- a/packages/j/joltphysics/xmake.lua
+++ b/packages/j/joltphysics/xmake.lua
@@ -70,7 +70,7 @@ package("joltphysics")
             package:add("syslinks", "advapi32")
         end
         package:add("defines", "JPH_PROFILE_ENABLED")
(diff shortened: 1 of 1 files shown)


zachrammell:
Looks like this was caused by #9957

PR #9982, opened 2026-05-14
Add cairomm (1.14.5, 1.19.0)

### LINKED ISSUE #9853: [New Package]: cairomm
### Action Type

New Package

### Package Name

cairomm

### Version(s)

v1.14.5, and the latest

### Project Repository

https://gitlab.freedesktop.org/cairo/cairomm

### Project Homepage

https://www.cairographics.org/cairomm/

### License

LGPL-2.1

### Supported Platforms

- [x] Windows
- [x] Linux
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD

Upstream: uses Meson
Dependencies: cairo, libsigcplusplus

### Additional Context

_No response_

EZ4Stephen:
1.14.x is the last version for the 1.0 ABI.

Seems to build but fails the test (at least for windows, or windows-static).
Example error:
`cairomm-vc143-1.16.lib(device.cc.obj) : error LNK2019: unresolved external symbol __imp_cairo_device_release referenced in function "public: __cdecl Cairo::Device::Lock::~Lock(void)" (??1Lock@Device@Cairo@@QEAA@XZ)`.
Would appreciate if someone figures out how to fix the errors arising from the on_test.

Also need to figure out which platforms don't have the required dependencies and exclude them.

This would close #9853 .

### DIFF
--- a/packages/c/cairomm/xmake.lua
+++ b/packages/c/cairomm/xmake.lua
@@ -13,6 +13,11 @@ package("cairomm")
     add_configs("exceptions_api", {description = "Build exceptions API and include it in the library", default = true, type = "boolean"})
     add_configs("shared", {description = "Build shared library.", default = false, type = "boolean"})
 
+    if is_subhost("windows") then
+        add_deps("pkgconf")
+    else
+        add_deps("pkg-config")
+    end
     add_deps("meson", "ninja")
 
     on_load(function (package)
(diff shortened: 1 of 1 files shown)


EZ4Stephen:
Looks like all Windows/MinGW shared builds need a fix now.

EZ4Stephen:
I just realized that pkgconf/pkg-config would get installed anyways because of cairo's xmake.lua file. So I assume all test will pass again, to which this can be merged.

PR #9984, opened 2026-05-14
Add pangomm (2.46.4, 2.56.1)

### LINKED ISSUE #9854: [New Package]: pangomm
### Action Type

New Package

### Package Name

pangomm

### Version(s)

v2.42.2, and the latest.

### Project Repository

https://gitlab.gnome.org/GNOME/pangomm

### Project Homepage

_No response_

### License

LGPL-2.1

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Android
- [ ] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD

Upstream uses CMake.
Dependencies: v2.42.2: glibmm-2.4, cairomm-1.0, pangocairo
       Latest: giomm-2.68, cairomm-1.16, pangocairo

### Additional Context

_No response_

EZ4Stephen:
2.46.4 seemed to be the latest version for the old 1.4 ABI.

This will close #9854 .

### DIFF
--- a/packages/g/glibmm/xmake.lua
+++ b/packages/g/glibmm/xmake.lua
@@ -45,7 +45,9 @@ package("glibmm")
         -- Prevent building generate_defs_glib, generate_defs_gio. Otherwise when being built,
         -- errors arise on mingw builds. Doesn't hurt to filter for all platforms.
         io.replace("tools/extra_defs_gen/meson.build", "%s*executable%b()", "", {pattern = true, multiline = true})
-        local configs = {"-Dbuild-documentation=false", "-Dbuild-examples=false"}
+        local configs = {"-Dbuild-documentation=false",
+                         "-Dbuild-examples=false",
+                         -Dmsvc14x-parallel-installable=false"}
         table.insert(configs, "-Dbuild-deprecated-api=" .. (package:config("deprecated_api") and "true" or "false"))
         table.insert(configs, "-Ddefault_library=" .. (package:config("shared") and "shared" or "static"))
         import("package.tools.meson").install(package, configs)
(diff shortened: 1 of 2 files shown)


EZ4Stephen:
Looks like I'll have to filter out windows+arm due to pango.

waruqi:
On Windows, meson couldn't find the dependent pango, so it pulled the source code remotely to compile. You could try adding `pkgconf` or checking the pango's configuration file to see why meson couldn't find it. Alternatively, you could force-disable bundled pango, making meson always use the pango dependency provided by xmake. However, this might require some patches.

EZ4Stephen:
Ah, I hadn't looked into it building pango from source. I assumed it just failed because the right version of pango wasn't found. I also need to fix the platform exclusion logic.

But regarding pkgconf, isn't pkgconf added indirectly already, through cairomm due to cairo, at least?

waruqi:
> Ah, I hadn't looked into it building pango from source. I assumed it just failed because the right version of pango wasn't found. I also need to fix the platform exclusion logic.
> 
> But regarding pkgconf, isn't pkgconf added indirectly already, through cairomm due to cairo, at least?

Yes, theoretically pkgconf should already exist, but whether meson found the pango dependency through pkgconf needs to be confirmed through debugging. If it can find it, it will not fallback to building it from source code.

EZ4Stephen:
Hmm, why does glibmm_generate_extra_defs-2.68.lib not get found despite being built, I wonder.

EZ4Stephen:
I've done some tests locally after some edits, but am still failing. I note that for glibmm-2.68, what is built is named `glibmm_generate_extra_defs-vc145-2.68` , but the failure for the build is that it can't find `glibmm_generate_extra_defs-vc143-2.68`. (Note the vc143/vc145 difference.) I don't know if that file naming difference error is an xmake, package-specific or meson/msvc problem. Further investigation needed.

For the 2.4 ABI, the .lib file name made was `glibmm_generate_extra_defs-vc143-2.4`, for reference, and I don't the error ever being about that; I've only seen the error logs be about the 2.68 one.

---

Maybe my analysis is wrong/flawed, but I'll remove 2.56.1 temporarily and see if builds pass.

EZ4Stephen:
Awaiting review.

PR #9989, opened 2026-05-15
Update pango to 1.57.1, minor changes to cairo's xmake.lua

EZ4Stephen:


EZ4Stephen:
Looks like my attempt to fix the fontconfig problem didn't work. Maybe cairo's .pc files are behind the problem? Wish I knew what to do, as I need the newer version of pango for the pangomm PR.

EZ4Stephen:
I see at least one Windows, non-arm build passed, so I feel like my fix was correct. But I suspect 2 or more MinGW builds might've frozen.
Edit: I'll have to look into MinGW again, it wasn't stuck.

EZ4Stephen:
I wonder if there's some subtle problem between using one of pkgconf or pkg-config, instead of the other...

EZ4Stephen:
I see one MinGW shared build passed (finally), but a static one failed. Need to see if that pattern of static builds failing persists.

(Edit: Wasn't a test fail.)

EZ4Stephen:
Awaiting review/merge on this one.

packages/c/cairo/xmake.lua
```diff
@@ -98,6 +98,18 @@ package("cairo")
...
+
+        -- Fix windows/mingw static builds, for when cairo is used as a dependency.
+        for _, pc in ipairs(os.files(path.join(package:installdir("lib", "pkgconfig"), "*.pc"))) do
+            if package:is_plat("windows") and not package:config("shared") then
+                io.replace(pc, "Cflags:", "Cflags: -DCAIRO_WIN32_STATIC_BUILD=1")
+            end
+            --cairo 1.18.0+: DWrite backend needs C++ runtime
+            if package:is_plat("mingw") and not package:config("shared") then
+                io.replace(pc, "Cflags:", "Cflags: -DCAIRO_WIN32_STATIC_BUILD=1")
+                io.replace(pc, "Libs:", "Libs: -lstdc++", {plain = true})
```
SirLynix, line 110: Why is static linking tied to static runtime here?

EZ4Stephen: Sorry I just learned to distinguish linking and runtime after your review. Is the correct approach `if package:is_plat("mingw") then`? I assumed the problem was something to do with cairo and static linking only and needing C++ compiler, when cairo was used as a dependency.

SirLynix: It shouldn't be necessary, xmake will inject runtime flags automatically when the runtime is set as static.

EZ4Stephen: So your doubt is regarding the existence of `io.replace(pc, "Libs:", "Libs: -lstdc++", {plain = true})` in the script?

SirLynix: Yes exactly

EZ4Stephen: See https://github.com/xmake-io/xmake-repo/actions/runs/26027548096/job/76504725109#step:7:29118

Around line 29115 or a few10's of lines later onwards.

EZ4Stephen:
Any more reviews?

packages/c/cairo/xmake.lua
waruqi: Currently, this seems to be the only way. This should only affect meson/cmake's search for dependency packages, and not xmake's integration of packages.

PR #9990, opened 2026-05-15
Add HarmonyOS target support for openssl/openssl3

hiyms:
* Patch openssl/openssl3 `Configurations/10-main.conf` to inject ohos-aarch64/ohos-arm/ohos-x86_64 target definitions

* Add "harmony" to the platform list in the package test script

Patch files from:  https://github.com/ohos-rs/ohos-openssl/blob/main/patchs/openssl.patch & https://github.com/ohos-rs/ohos-openssl/blob/111/patchs/openssl.patch

### DIFF
--- a/packages/o/openssl/configure/patch.lua
+++ b/packages/o/openssl/configure/patch.lua
@@ -168,8 +168,30 @@ function _replace_NUL_with_null(package)
     io.replace("Configurations/10-main.conf", "NUL", "null", {plain = true})
 end
 
+function _add_ohos_targets(package)
+    if package:is_plat("harmony") then
+        io.gsub("Configurations/10-main.conf",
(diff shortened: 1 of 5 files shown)


packages/o/openssl/configure/patch.lua
```diff
@@ -172,4 +172,22 @@ function main(package)
     _remove_unused_pod_usage(package)
     _replace_NUL_with_null(package)
     _fix_overlong_make_recipe(package)
+    if package:is_plat("harmony") then
```
waruqi, line 172: use a function to wrap it

packages/o/openssl3/configure/patch.lua
```diff
@@ -0,0 +1,23 @@
+function main(package)
+    if package:is_plat("harmony") then
```
waruqi, line 2: and here

PR #10012, opened 2026-05-18
Add stv 1.0.0

AkarinATCP:
[New package]: stv
A Lightweight C String-View Library.

This is a header-only library, only single header file, no build system and no deps.

### DIFF
--- a/packages/s/stv/xmake.lua
+++ b/packages/s/stv/xmake.lua
@@ -1,0 +1,23 @@
+package("stv")
+    set_kind("library", { headeronly = true })
+    set_description("A Lightweight C String-View Library")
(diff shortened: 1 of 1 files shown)


luadebug:
Maybe it is better to get more commits before submitting, though amount of commits required was not specified.
I reference to https://github.com/xmake-io/xmake-repo/pull/9690#issuecomment-4210382404

AkarinATCP:
> Maybe it is better to get more commits before submitting, though amount of commits required was not specified. I reference to [#9690 (comment)](https://github.com/xmake-io/xmake-repo/pull/9690#issuecomment-4210382404)

Thanks for the feedback.   
I will close this PR and work on improving the project first.

PR #10072, opened 2026-05-27
3 minor changes for libjpeg-turbo

EZ4Stephen:
- Made nasm as a dependency only for certain architectures (see https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/simd/CMakeLists.txt).
 - Added WITH_SIMD as an option.
 - Reformatted share/static defines to take 2 lines.

### DIFF
--- a/packages/l/libjpeg-turbo/xmake.lua
+++ b/packages/l/libjpeg-turbo/xmake.lua
@@ -8,7 +8,6 @@ package("libjpeg-turbo")
     end})
     add_urls("https://github.com/libjpeg-turbo/libjpeg-turbo.git", {alias = "git"})
 
-    add_versions("3.1.90", "076ef1431f2803a91f07e0f92433d4dcf39bc9113226c4f46ba3d3d54f514c9d")
     add_versions("3.1.4+1", "a7da42b640377c2a9a9665e2c4b0ea60cd5599afb48c2521e6df0c9dc9d15a25")
     add_versions("3.1.2", "560f6338b547544c4f9721b18d8b87685d433ec78b3c644c70d77adad22c55e6")
     add_versions("3.1.1", "304165ae11e64ab752e9cfc07c37bfdc87abd0bfe4bc699e59f34036d9c84f72")
--- a/packages/l/libjpeg-turbo/xmake.lua
+++ b/packages/l/libjpeg-turbo/xmake.lua
@@ -28,6 +28,7 @@ package("libjpeg-turbo")
     add_versions("git:3.1.4+1", "3.1.4.1")
 
     add_configs("jpeg", {description = "libjpeg API/ABI emulation target version.", default = "6", type = "string", values = {"6", "7", "8"}})
+    add_configs("simd", {description = "Include SIMD extensions, if available for this platform", default = true, type = "boolean"})
 
     if is_plat("android") then
         add_deps("make")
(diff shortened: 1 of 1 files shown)


EZ4Stephen:
I was going through the commit history of libjpeg-turbo, and I see,"[Update libjpeg-turbo to 3.1.90](https://github.com/xmake-io/xmake-repo/commit/adf3ca9e892fcd8e1bf24a1d748cf62b5d92741c)"

But it added the 3.1.4.1 tag, not 3.1.90, which is confusing me.

Tags page: https://github.com/libjpeg-turbo/libjpeg-turbo/tags
Difference between 3.1.4.1 / 3.1.90: https://github.com/libjpeg-turbo/libjpeg-turbo/compare/3.1.4.1...3.1.90

packages/l/libjpeg-turbo/xmake.lua
```diff
@@ -8,6 +8,7 @@ package("libjpeg-turbo")
     end})
     add_urls("https://github.com/libjpeg-turbo/libjpeg-turbo.git", {alias = "git"})
 
+    add_versions("3.1.90", "076ef1431f2803a91f07e0f92433d4dcf39bc9113226c4f46ba3d3d54f514c9d")
```
apocelipes: Dont add an alpha version. That's why I skip this version.

apocelipes (changes requested):

apocelipes:
> libjpeg-turbo/libjpeg-turbo@3.1.4.1...3.1.90

That's a beta release. See the Release page. 
image

[3.1.90 (3.2 beta1)](https://github.com/libjpeg-turbo/libjpeg-turbo/releases/tag/3.1.90) Pre-release

packages/l/libjpeg-turbo/xmake.lua
```diff
@@ -27,14 +28,18 @@ package("libjpeg-turbo")
...
 
     if is_plat("android") then
         add_deps("make")
     end
 
     on_load(function (package)
         if package:is_built() then
-            package:add("deps", "cmake", "nasm")
+            package:add("deps", "cmake")
+            if package:config("simd") and package:is_arch("x86_64", "i386", "x64", "x86") then
```
apocelipes, line 39: What about aarch64?

apocelipes (changes requested):

packages/l/libjpeg-turbo/xmake.lua
EZ4Stephen: Per my research, aarch doesn't fall under `if(CPU_TYPE STREQUAL "x86_64" OR CPU_TYPE STREQUAL "i386")` and there were no mentions of NASM in or beyond the `elseif(CPU_TYPE STREQUAL "arm64" OR CPU_TYPE STREQUAL "arm" OR WITH_SIMDE)` part of the CMakeLists.txt in the simd folder.

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/simd/CMakeLists.txt

You can also confirm that no significant change was done since/around the 3.1.4.1 release or so's timeframe which would negate this edit.

EZ4Stephen:
Should I add the WITH_SIMDE option as well?

apocelipes:
> WITH_SIMDE

> "Use SIMD Everywhere or Emscripten to translate Arm/Neon intrinsics into SIMD intrinsics for other platforms (Experimental)"

So we may need not to add this option.

EZ4Stephen:
Should License be saying IJG AND BSD-3-Clause, or should it remain as BSD-3-Clause?

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/LICENSE.md

apocelipes:
> https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/LICENSE.md

We should keep it as is.

apocelipes (approved):

PR #10115, opened 2026-06-02
Add stv v2.2.0

AkarinATCP:
A Lightweight C String-View Library.

Provides common string operations.

This is a header-only library, only single header file, no build system and no deps.

### DIFF
--- a/packages/s/stv/xmake.lua
+++ b/packages/s/stv/xmake.lua
@@ -3,9 +3,11 @@ package("stv")
     set_description("A Lightweight C String-View Library")
     set_homepage("https://github.com/AkarinATCP/stv")
     set_license("MIT")
-    
+
     set_urls("https://github.com/AkarinATCP/stv/archive/refs/tags/v$(version).tar.gz")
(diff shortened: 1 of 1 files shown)


AkarinATCP:
But actually, 'stv.h' is right in 'include/' .

packages/s/stv/xmake.lua
```diff
@@ -0,0 +1,17 @@
...
+    set_urls("https://github.com/AkarinATCP/stv/archive/refs/tags/v$(version).tar.gz")
+    add_versions("2.2.0", "5beed1ca3fa388266ad26130175b31811c8e4eb6ee7e1613dc7e761291e77da5")
+
+    on_install(function (package)
+        os.cp("include/stv.h", package:installdir("include"))
+    end)
+
+    on_test(function (package)
+        package:set("languages", "c99")
+        assert(package:has_cincludes("stv.h"), "stv.h not found")
```
waruqi: please use has_cfuncs to check links

PR #10161, opened 2026-06-09
chore: add licenses

apocelipes:
Only add licenses no functionality changes.


apocelipes:
Since only one package fails on wasm, I'd like to merge it.

PR #10217, opened 2026-06-16
fllama.cpp: add versions from 4000 to 9647

IncroyablePix:
Added the following llama.cpp versions:
- 9647
- 9500
- 9000
- 8500
- 6000
- 4000

### DIFF
--- a/packages/l/llama.cpp/xmake.lua
+++ b/packages/l/llama.cpp/xmake.lua
@@ -28,9 +28,12 @@ package("llama.cpp")
         add_syslinks("advapi32")
     end
 
-    add_links("llama", "ggml", "ggml-base", "ggml-cpu")
+    add_links("llama", "ggml")
 
-    add_deps("cmake", "ninja")
+    add_deps("cmake")
+    if is_plat("windows") then
+        add_deps("ninja")
+    end
 
     if on_check then
         on_check("android", function (package)
@@ -54,6 +57,11 @@ package("llama.cpp")
     end
 
     on_load(function (package)
+        local version = tonumber(package:version_str())
+        if version and version >= 3800 then
+            package:add("links", "ggml-base", "ggml-cpu")
+        end
+
         if package:config("shared") then
             package:add("defines", "GGML_SHARED", "LLAMA_SHARED")
         end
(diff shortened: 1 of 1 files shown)


packages/l/llama.cpp/xmake.lua
```diff
@@ -96,7 +108,40 @@ package("llama.cpp")
...
+            end
+        end
+
+        local opt = {builddir = "b"}
+        if package:is_plat("windows") then
+            local ninja = package:dep("ninja")
+            if ninja then
+                local ninja_exe = path.join(ninja:installdir("bin"), "ninja.exe")
+                if os.isfile(ninja_exe) then
+                    table.insert(configs, "-DCMAKE_MAKE_PROGRAM=" .. ninja_exe)
```
waruqi, line 146: why? The cmake package now uses ninja by default, and xmake will configure it automatically, so you don't need to do anything.

packages/l/llama.cpp/xmake.lua
```diff
@@ -18,11 +24,13 @@ package("llama.cpp")
...
         add_syslinks("pthread")
+    elseif is_plat("windows") then
+        add_syslinks("advapi32")
     end
 
-    add_links("llama", "ggml")
+    add_links("llama", "ggml", "ggml-base", "ggml-cpu")
 
-    add_deps("cmake")
+    add_deps("cmake", "ninja")
```
waruqi: cmake has added ninja deps now.

packages/l/llama.cpp/xmake.lua
```diff
@@ -96,7 +108,40 @@ package("llama.cpp")
...
+                if os.isfile(ninja_exe) then
+                    table.insert(configs, "-DCMAKE_MAKE_PROGRAM=" .. ninja_exe)
+                end
+            end
+        end
+
+        import("package.tools.cmake").install(package, configs, opt)
+
+        if package:config("vulkan") then
+            local vulkan_libs = os.files(path.join(package:installdir("lib"), "*ggml-vulkan*"))
```
waruqi, line 154: you should check it in on_test, and use lib.detect.find_library

packages/l/llama.cpp/xmake.lua
```diff
@@ -96,7 +108,40 @@ package("llama.cpp")
...
+            if not (fetched and fetched.program) and not os.isfile(glslc) then
+                glslc = nil
+            end
+            assert(glslc, "package(llama.cpp): Vulkan requires glslc from shaderc")
+            if glslc then
+                table.insert(configs, "-DVulkan_GLSLC_EXECUTABLE=" .. glslc)
+            end
+        end
+
+        local opt = {builddir = "b"}
```
waruqi, line 140: ？why change builddir?

Issue #10219, opened 2026-06-17
[New Package]: RGFW


LiYulin-s:
### Action Type

New Package

### Package Name

RGFW

### Version(s)

1.8.1

### Project Repository

https://github.com/ColleagueRiley/RGFW.git

### Project Homepage

https://github.com/ColleagueRiley/RGFW.git

### License

zlib

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Android
- [ ] iOS
- [x] WebAssembly (Wasm)
- [x] FreeBSD
- [ ] Other (specify in Additional Context)

### Build System & Dependencies

Upstream use Makefile or headeronly
Requires libx11, libxrandr, libxcursor, libxi, libxext for x11
wayland, wayland-protocols, libxkbcommon for wayland

optionally depends on opengl

PR #10220, opened 2026-06-17
Add RGFW package

LiYulin-s:
#10219

### DIFF
--- a/packages/r/rgfw/xmake.lua
+++ b/packages/r/rgfw/xmake.lua
@@ -1,5 +1,4 @@
 package("rgfw")
-
     set_homepage("https://github.com/ColleagueRiley/RGFW")
     set_description("A single-header windowing framework for creating windows, graphics contexts and handling input.")
     set_license("zlib")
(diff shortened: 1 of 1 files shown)


packages/r/rgfw/xmake.lua
```diff
@@ -0,0 +1,114 @@
...
+    add_configs("headeronly", { description = "Use RGFW as a single-header library.", default = true, type = "boolean" })
+    add_configs("x11", { description = "Build support for X11.", default = is_plat("linux", "bsd"), type = "boolean" })
+    add_configs("wayland", { description = "Build support for Wayland.", default = false, type = "boolean" })
+
+    add_deps("opengl", { optional = true })
+
+    if is_plat("windows", "mingw", "msys") then
+        add_syslinks("gdi32", "shell32", "user32")
+    elseif is_plat("macosx") then
+        add_frameworks("Cocoa", "CoreVideo", "OpenGL", "IOKit")
```
ColleagueRiley: Linking OpenGL here would break the optional opengl linking from ealier. 

In RGFW 2.0 is released you won't need to link RGFW at all because it's linked dynamically, as that is better for bindings/using RGFW as a shared library.

PR #10278, opened 2026-06-26
Add aubio

EZ4Stephen:


EZ4Stephen:
I've just realized I could pass more defines without necessarily doing more io.replace. Will be doing a lot of edits (in one commit) once the BSD runs pass.

EZ4Stephen:
Might be worthwhile to add HAVE_MEMCPY_HACKS, and the duo of HAVE_SOURCE_APPLE_AUDIO and HAVE_SINK_APPLE_AUDIO unconditionally. The last 2 might need more changes though. I don't mind if the package is accepted as-is.

PR #10291, opened 2026-06-27
libarchive: Add fallback link

EZ4Stephen:


EZ4Stephen:
My build was failing at the download step; curl would give an error 30 seconds into trying to download libarchive (and then saying `error: unmatched checksum, current hash(ac1618bd) != original hash(038918ea)`). Investigated https://libarchive.org/downloads/ and saw it was missing the recently released 3.8.8. Checked that the sha256 of the github release matched the one already in the file before making the edit.

apocelipes (approved):

PR #10327, opened 2026-07-02
Update geo-utils-cpp to v1.2.0

gistrec:
Adds v1.1.0 of [geo-utils-cpp](https://github.com/gistrec/geo-utils-cpp) (header-only C++17 spherical lat/lng geometry).

Upstream release notes: https://github.com/gistrec/geo-utils-cpp/releases/tag/v1.1.0 — adds Encoded Polyline `encode`/`decode`, Douglas-Peucker `simplify`, `is_closed_polygon`, `LatLng::is_valid()`, and antimeridian/pole correctness fixes. Install logic and the test snippet are unchanged.

Verified locally on macOS arm64 with `xmake l scripts/test.lua -D -k geo-utils-cpp` — installs 1.1.0 and passes the snippet test.

### DIFF
--- a/packages/g/geo-utils-cpp/xmake.lua
+++ b/packages/g/geo-utils-cpp/xmake.lua
@@ -7,6 +7,7 @@ package("geo-utils-cpp")
     add_urls("https://github.com/gistrec/geo-utils-cpp/archive/refs/tags/v$(version).tar.gz",
              "https://github.com/gistrec/geo-utils-cpp.git")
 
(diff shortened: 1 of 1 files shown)


gistrec:
Added the just-released [v1.2.0](https://github.com/gistrec/geo-utils-cpp/releases/tag/v1.2.0) on top (both 1.1.0 and 1.2.0 tags are live upstream and installable). Verified locally on macOS arm64 with `xmake l scripts/test.lua -D -k geo-utils-cpp` — installs 1.2.0 and passes the snippet test.

PR #10330, opened 2026-07-02
Fix syslinks for wgsl-validator on Windows and Linux platforms

luadebug:
Fixes #8424

### DIFF
--- a/packages/w/wgsl-validator/xmake.lua
+++ b/packages/w/wgsl-validator/xmake.lua
@@ -13,9 +13,9 @@ package("wgsl-validator")
     add_configs("shared", {description = "Build shared library.", default = false, type = "boolean", readonly = true})
 
     if is_plat("windows", "mingw") then
-        add_syslinks("Advapi32", "User32", "Userenv", "WS2_32", "RuntimeObject", "NtDll")
+        add_syslinks("kernel32", "ntdll", "userenv", "ws2_32", "dbghelp")
     elseif is_plat("linux", "bsd") then
-        add_syslinks("pthread")
+        add_syslinks("dl", "m", "pthread", "rt", "util")
     end
 
     on_install(function (package)
--- a/packages/w/wgsl-validator/xmake.lua
+++ b/packages/w/wgsl-validator/xmake.lua
@@ -18,7 +18,7 @@ package("wgsl-validator")
         add_syslinks("dl", "m", "pthread", "rt", "util")
     end
 
-    on_install(function (package)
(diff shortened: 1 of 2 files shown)


packages/w/wgsl-validator/xmake.lua
```diff
@@ -33,6 +33,13 @@ package("wgsl-validator")
                 set_values("rust.cratetype", "staticlib")
                 add_packages("cargo::naga")
         ]])
+        if is_host("windows") and package:is_plat("mingw", "msys") then
+            local mingw_root = os.programdir():match("^(.-[/\\]mingw%d+)[/\\]")
```
waruqi, line 37: why?

luadebug: To capture MinGW root, D:\a\_temp\msys64\mingw64. Then path.join(mingw_root, "bin") => …\msys64\mingw64\bin — the directory holding the runtime DLLs xmake.exe needs.

So we resolve ```-1073741511 = 0xC0000139 = STATUS_ENTRYPOINT_NOT_FOUND ("the procedure entry point X could not be located in DLL Y")```

waruqi: Why can the xmake parent process run normally?

In addition, mingw xmake should be statically linked and not rely on dll. 

```
elif is_plat "mingw"; then 
add_ldflags "-static-libgcc" "-lwintrust" "-lcrypt32"
```

If it truly is a missing DLL issue, the changes should be made in tools.xmake, not just in this single package.

Also, other packages' xmake.install processes never seem to have this problem.

luadebug: If it depends on zero non-system DLLs there probably wont be such problem. We can edit xmake's core/xmake.sh. Instead of ncursesw / readline (dynamic link) we should go for static link, so it would avoid symbol clock_gettime64 lack. Here is workaround https://github.com/xmake-io/xmake/pull/7637

https://medium.com/@chandanarajiv01/solved-c-code-not-compiling-in-vs-code-clock-gettime64-entry-point-not-found-error-gcc-15-2-0-b98ad83f42a2 - this website states out gcc 15.2.0 has similiar problem because that version was released with toolchain bug I guess.

But I dont know reason, here is mention of that https://github.com/msys2/MINGW-packages/issues/24355

waruqi: I accidentally merged it just now, but we can open a new PR to continue making improvements once that xmake patch gets merged.

PR #10340, opened 2026-07-03
Update libcurl

EZ4Stephen:


EZ4Stephen:
I want to know if the first change (removing _imp prefix for the .lib) should be reverted and that prefix goes in the .pc file fix.

EZ4Stephen:
Also I wonder if libcurl was dodging those auto-updates. I don't quite know how those work.

EZ4Stephen:
I tried committing and got the warning about a commit that happened while I was editing. Thanks for the fixes.

PR #10362, opened 2026-07-06
libusb: update to 1.0.30

luadebug:
Workaround for https://github.com/xmake-io/xmake-repo/pull/10361

### DIFF
--- a/packages/l/libusb/xmake.lua
+++ b/packages/l/libusb/xmake.lua
@@ -6,12 +6,13 @@ package("libusb")
     add_urls("https://github.com/libusb/libusb/archive/refs/tags/$(version).tar.gz",
              "https://github.com/libusb/libusb.git")
 
+    add_versions("v1.0.30", "2ae28adb0bb9558c86135c4e1c11b320b0805461e207a64a6e520a114094bf07")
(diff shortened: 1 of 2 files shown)


packages/l/libusb/xmake.lua
```diff
@@ -55,6 +60,8 @@ package("libusb")
             [[get_filename_component(LIBUSB_ROOT "libusb/libusb" ABSOLUTE)]],
             [[get_filename_component(LIBUSB_ROOT "libusb" ABSOLUTE)]], {plain = true})
 
+        import("devel.git").apply(path.join(package:scriptdir(), "patches", "libusb-cmake-pkgconfig.patch"))
```
waruqi, line 63: why not use add_patches?

luadebug: ```os.cp(path.join(dir, "CMakeLists.txt"), os.curdir())``` line 57 appears inside of ```on_install("!iphoneos and !bsd", function (package)``` so CMakeLists.txt being copied from resource folder to source folder, as I assume add_patches meant to patch only once on start directly inside of source folder.

waruqi: well

PR #10370, opened 2026-07-06
Add liblsl and brainflow

MasterLaplace:
Two packages that aren't in the repo yet, used together for EEG/biosignal acquisition.

- **liblsl** (sccn/liblsl): the Lab Streaming Layer, for time-synchronised
  streaming of multi-channel data over a local network. Boost and pugixml are
  built bundled, so it needs no extra packages.
- **brainflow** (brainflow-dev/brainflow): a uniform API to pull and parse data
  from EEG/EMG/ECG boards. The optional heavy backends (Bluetooth, BLE, ONNX,
  OpenMP, libftdi) are off by default to keep the build self-contained.

Both carry an `on_test` and pass `scripts/test.lua` locally on Linux
(liblsl v1.17.7, brainflow 5.22.2).

### DIFF
--- a/packages/b/brainflow/xmake.lua
+++ b/packages/b/brainflow/xmake.lua
@@ -41,7 +41,8 @@ package("brainflow")
             "-DBUILD_SYNCHRONI_SDK=OFF",
             "-DUSE_LIBFTDI=OFF",
             "-DUSE_OPENMP=OFF",
-            "-DWARNINGS_AS_ERRORS=OFF"
+            "-DWARNINGS_AS_ERRORS=OFF",
+            "-DBUILD_SHARED_LIBS=OFF"
         }
 
         table.insert(configs, "-DCMAKE_BUILD_TYPE=" .. (package:is_debug() and "Debug" or "Release"))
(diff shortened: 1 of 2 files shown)


packages/b/brainflow/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+
+    add_deps("cmake")
+
+    on_load(function (package)
+        package:add("includedirs", "inc")
+        package:add("links", "Brainflow", "BoardController", "DataHandler", "MLModule")
+        if package:is_plat("linux") then
+            package:add("syslinks", "pthread", "dl")
+        end
+    end)
```
waruqi: For simple configuration logic, there is no need to configure it within `on_load`.

```lua
add_includedirs("inc")
add_links("Brainflow", "BoardController", "DataHandler", "MLModule")
if is_plat("linux") then
    add_syslinks("pthread", "dl")
end
```

MasterLaplace: Done, and made the same change to liblsl

PR #10408, opened 2026-07-12
onnxruntime: fix cmake

choyy:


packages/o/onnxruntime/xmake.lua
```diff
@@ -129,6 +129,14 @@ package("onnxruntime")
             os.mv("lib/*.dll", package:installdir("bin"))
         end
         os.cp("*", package:installdir())
+        local cmake_file = package:installdir("lib/cmake/onnxruntime/onnxruntimeTargets.cmake")
```
waruqi: Why not replace it in the current source code directory before installation?

PR #10415, opened 2026-07-12
Auto-update joltphysics to v5.6.0

SirLynix:
New version of joltphysics detected (package version: v5.5.0, last github version: v5.6.0)


SirLynix:
I disabled GPU support by default due to required packages (dxc, glslang) having poor support outside of x86 64bits but it would be a nice addition in the future

PR #10429, opened 2026-07-14
Add structural_interface package

Maltose0118:
Add the structural_interface package for the C++26 header-only structural interface and type-erasure library.\n\nThe package was tested locally with xmake v3.0.9 and GCC 16.0.1 using the official package test script. The test requires GCC static reflection support and -freflection.

### DIFF
--- a/packages/s/structural_interface/xmake.lua
+++ b/packages/s/structural_interface/xmake.lua
@@ -1,0 +1,39 @@
+package("structural_interface")
+    set_policy("check.auto_ignore_flags", false)
+    set_kind("library", {headeronly = true})
+    set_homepage("https://github.com/Maltose0118/structural_interface")
+    set_description("Header-only C++26 library for structural interfaces and type erasure")
+    set_license("MIT")
+
+    add_urls("https://github.com/Maltose0118/structural_interface/archive/refs/tags/v$(version).tar.gz")
+    add_versions("0.1.0", "a71b07193d9fdfc1bf74b5898a057331ead982ee8896567ea51761af4ba184e1")
+
+    on_load(function(package)
+        package:add("cxxflags", "-std=c++26", "-freflection", {
+            tools = "gcc",
+            force = true
+        })
+    end)
+
+    on_install(function(package)
+        os.cp("include", package:installdir())
+    end)
+
+    on_test(function(package)
+        assert(package:check_cxxsnippets({test = [[
+            #include <structural_interface.hpp>
+
+            void test() {
+                struct Interface {
+                    void draw() const;
+                };
+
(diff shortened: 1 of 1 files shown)


Maltose0118:
Updated the package recipe to disable automatic ignoring of `-freflection` and keep the required `-std=c++26`/`-freflection` flags for GCC consumers. The suggested package-level `set_languages(c++26)` is not available in xrepo package scripts (it raises an undefined global error), so the standard is retained as a forced package C++ flag. Re-tested locally with xmake v3.0.9 and GCC 16.0.1; the official package test passes.

packages/s/structural_interface/xmake.lua
```diff
@@ -0,0 +1,39 @@
+package("structural_interface")
+    set_policy("check.auto_ignore_flags", false)
```
waruqi: remove it

packages/s/structural_interface/xmake.lua
```diff
@@ -0,0 +1,39 @@
...
+    set_description("Header-only C++26 library for structural interfaces and type erasure")
+    set_license("MIT")
+
+    add_urls("https://github.com/Maltose0118/structural_interface/archive/refs/tags/v$(version).tar.gz")
+    add_versions("0.1.0", "a71b07193d9fdfc1bf74b5898a057331ead982ee8896567ea51761af4ba184e1")
+
+    on_load(function(package)
+        package:add("cxxflags", "-std=c++26", "-freflection", {
+            tools = "gcc",
+            force = true
```
waruqi: ? 

```lua
add_cxxflags("...")
```

Maltose0118:
Updated the package recipe based on the review feedback.

- Removed `set_policy("check.auto_ignore_flags", false)`.
- Replaced the `on_load`/`package:add(...)` block with the top-level `add_cxxflags(...)` form.
- Added an `on_check` guard for GCC. In the current xmake package environment, the C++ GCC tool is named `gxx`; checking for `gcc` incorrectly rejects the working GCC toolchain.
- Tested `set_languages("c++26")`, but it is not available in the current xrepo package-script environment and raises an undefined-global error, so `-std=c++26` remains an explicit forced package flag.
- `check_cxxsnippets` also needs `configs.force.cxflags` to pass the experimental `-freflection` flag through the package test.

The official test script now passes in Fedora WSL with xmake `v3.0.9+HEAD.2b184e1` and GCC 16.1.1: all template tests passed and the `structural_interface` package test passed. xmake still emits its warning for the experimental reflection flag, but the test exits successfully.

waruqi:
We should let users configure the C++26 flags themselves to avoid conflicts with other packages.

PR #10440, opened 2026-07-16
libui: update to 2025.3.15

SFGrenade:
and make it buildable again

### DIFF
--- a/packages/g/gdk-pixbuf/xmake.lua
+++ b/packages/g/gdk-pixbuf/xmake.lua
@@ -65,6 +65,13 @@ package("gdk-pixbuf")
         io.gsub("meson.build", "subdir%('fuzzing'%)", "")
         io.gsub("meson.build", "subdir%('docs'%)", "")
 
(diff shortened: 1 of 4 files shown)


luadebug:
```xmake
        -- get links and ensure link order
        import("core.base.graph")
        local dag = graph.new(true)
        local pkgconfigdir = package:installdir("lib", "pkgconfig")
        for _, pcfile in ipairs(os.files(path.join(pkgconfigdir, "*.pc"))) do
            local link = path.basename(pcfile)
            local content = io.readfile(pcfile)
            for _, line in ipairs(content:split("\n")) do
                if line:startswith("Requires: ") then
                    local requires = line:sub(10):split(",")
                    for _, dep in ipairs(requires) do
                        dep = dep:split("=")[1]:trim()
                        dag:add_edge(link, dep)
                    end
                end
            end
        end
        local links = dag:topological_sort()
        package:add("links", links)
```

It made to lose Windows static library built links for some reason... 💫 😵 
Well nevermind, I dont understand how to use that. :D

Issue #10470, opened 2026-07-20
glib 2.89.2 update incompatibility for glibmm


EZ4Stephen:
Since the glib update to 2.89.2, glibmm 2.66.8 failed to build. Apparently it has something to do with `G_DECLARE_FINAL_TYPE`, and gio/gdbusactiongroup.h and gio/gemblem.h.

Exact errors:
```
[74/280] Compiling C++ object gio/giomm/giomm-2.4-1.dll.p/.._.._untracked_gio_giomm_wrap_init.cc.obj
FAILED: [code=2] gio/giomm/giomm-2.4-1.dll.p/.._.._untracked_gio_giomm_wrap_init.cc.obj 
[...]

PR #10471, opened 2026-07-20
glibmm: cap glib dependency version to <2.89.2

EZ4Stephen:
#10470 .


EZ4Stephen:
macOS x86_64 static needs a rerun, freebsd shared might be stuck. I assume the PR can be merged anyways, but if not, reruns are needed.

EZ4Stephen:
The two still need reruns...

PR #10480, opened 2026-07-22
Update on_install function for macOS arm64 support

pixcai:
Add arm64 architecture support option


SirLynix:
merged despite failing test because they're unrelated to macOS

Issue #9841, opened 2026-04-24
[New Package]: boost_uuid


EZ4Stephen:
### Action Type

New Package

### Package Name

boost_uuid

### Version(s)

latest

### Project Repository

https://github.com/boostorg/uuid

### License

BSL-1.0

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Android
- [ ] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD
- [ ] Other (specify in Additional Context)

### Build System & Dependencies

Upstream uses b2/CMake
Dependencies: boost_assert, boost_cmake, boost_config, boost_headers, boost_throw_exception, boost_type_traits

kekopom:
@EZ4Stephen merged can you check if it fits your needs plz?

kekopom:
you can use this config in xmake.lua:

```lua
add_requires("boost", {config=uuid})
add_rules("mode.debug", "mode.release")

target("your_target")
    add_packages("boost")
    set_kind("binary")
    add_files("src/*.cpp")
```

EZ4Stephen:
I mean, it does build uuid, but I wish there was a way to build just uuid and its dependencies. Currently add_requires("boost", {config=uuid}) does install a lot (all?) of headers as well.

Maybe you can refer to weights or levels in https://pdimov.github.io/boostdep-report/.
(Also contrary to what my open boost issues may imply, I think the logic staying inside packages/b/boost/ would be ideal, to get this right, rather than separate boost_* packages.)

My test file which helps see what includes and libs get installed:

```
add_requires("boost", {config=uuid})

target("collect_deps")
    set_kind("phony")
    after_load(function()
        import("core.project.project")
        import("core.project.config")
        local dest = path.join(os.projectdir(), config.mode() or "release")
        local subdirs = {"include", "lib"}
        if is_plat("windows") then table.insert(subdirs, "bin") end
        for _, pkg in pairs(project.required_packages()) do
            local dir = pkg:installdir()
            if dir and os.isdir(dir) then
                for _, sub in ipairs(subdirs) do
                    local src = path.join(dir, sub)
                    if os.isdir(src) then
                        for _, fp in ipairs(os.files(path.join(src, "**"))) do
                            local tgt = path.join(dest, sub, path.relative(fp, src))
                            os.mkdir(path.directory(tgt))
                            os.cp(fp, tgt)
                        end
                    end
                end
            end
        end
    end)
[...]

PR #10483, opened 2026-07-23
add boost uuid

kekopom:

------

https://github.com/xmake-io/xmake-repo/issues/9841


packages/b/boost/libs.lua
kekopom, line 174: shall we add config and uuid as header only buildable?

I would say so since both don't have src/ folder in their github

kekopom:
I'm not 100% sure the test I wrote is actually ran? anyone knows if we can trigg it in ci ?

function _uuid(package, snippets)

packages/b/boost/libs.lua
```diff
@@ -162,7 +164,11 @@ local libs_dep = {
   },
   spirit = {
     "thread",
-  }
+  },
+  uuid = {
+    "config"
+  },
```
kekopom, line 170: according to https://github.com/boostorg/uuid/blob/develop/CMakeLists.txt#L15 there might be more deps to add

kekopom:
been able to test locally with 
xmake l scripts/test.lua --shallow -vD -f "uuid=true" boost

packages/b/boost/libs.lua
kekopom: Not needded

kekopom:
@waruqi hi to me this is ready to review/merge if you could have a look plz? :)

packages/b/boost_config/xmake.lua
```diff
@@ -0,0 +1,23 @@
...
+    add_versions("1.89.0", "a16a53f8f0dde2a9ff485fb134a069e54aca738009d825837dac9a35cd4afe0a")
+
+    on_install(function (package)
+        os.cp("include/boost", package:installdir("include"))
+    end)
+
+    on_test(function (package)
+        assert(package:check_cxxsnippets({test = [[
+            #include <boost/config.hpp>
+            void test() { }
```
waruqi: empty implmention? why? please add some test codes

kekopom: added test code

PR #10532, opened 2026-07-29
Add h2o package

Persson-dev:
For some reasons, the package was forced to be built with `-g3` flag even in release, and only `-O2` optimization.
Should i keep the patch as it is, or just let xmake decide for the appropriate flags to use ?


Persson-dev:
The builds seems to fail because of : `undefined reference to "yaml_event_delete"` but I'm unsure of what to do ...

Persson-dev:
For some reason, the link error only happens when the package is built in shared and debug mode

PR #10544, opened 2026-07-31
Update cairo to 1.18.4

EZ4Stephen:


EZ4Stephen:
freebsd static run is stuck.

PR #10552, opened 2026-08-02
pcre2: Update to 10.47, update links

EZ4Stephen:


EZ4Stephen:
Doing this to indirectly fix building debug/shared version of pcre2 with MSVC.

PR #10584, opened 2026-08-06
bgfx: update to 9392 and support MD runtime

FFFelian:
\* update to 9392 and replace` bgfx::init() `with `bgfx::init(init)` since the default argument was removed in 9392 .
 \* add support to MD/MDd runtime by passing `--with-dynamic-runtime`.


FFFelian:
我看最新genie 1204已经可以处理vs2026了, 改成`add_deps("genie 1204")`, 去掉`if tonumber(vs) >= 2022`是不是也行

packages/b/bgfx/xmake.lua
```diff
@@ -72,15 +77,15 @@ package("bgfx")
...
             msbuild.build(package, configs)
 
-            os.trycp("../../win*_vs*/bin/*.lib|*example*", package:installdir("lib"))
-            os.trycp("../../win*_vs*/bin/*.dll", package:installdir("lib"))
-            os.trycp("../../win*_vs*/bin/*.lib", package:installdir("lib"))
-            os.trycp("../../win*_vs*/bin/*.exe", package:installdir("bin"))
+            os.trycp("../../*_vs*/bin/*.lib|*example*", package:installdir("lib"))
+            os.trycp("../../*_vs*/bin/*.dll", package:installdir("lib"))
+            os.trycp("../../*_vs*/bin/*.lib", package:installdir("lib"))
+            os.trycp("../../*_vs*/bin/*.exe", package:installdir("bin"))
```
luadebug, line 97: Is this going to work with old versions of bgfx?

packages/b/bgfx/xmake.lua
```diff
@@ -111,7 +117,21 @@ package("bgfx")
             local envs = make.buildenvs(package)
             envs.BX_DIR = bxdir
             envs.BIMG_DIR = bimgdir
+            
+            if package:is_plat("windows") and package:is_arch("arm64") then
+                os.rm("3rdparty/glsl-optimizer/include/c99/stdint.h")
+            end
+            if package:version() and package:version():ge("9392") and package:is_plat("iphoneos") then
+                io.replace("scripts/bgfx.lua", 'configuration { "osx*" }', 'configuration { "ios*" }\n\t\tbuildoptions { "-x objective-c++" }\n\tconfiguration { "osx*" }', {plain = true})
+            end
```
luadebug, line 132: Try delete it maybe it would work without this, since we updated genie.

packages/b/bgfx/xmake.lua
```diff
@@ -9,40 +8,43 @@ package("bgfx")
...
     elseif is_plat("iphoneos") then
-        add_frameworks("OpenGLES", "CoreGraphics", "Metal", "QuartzCore", "UIKit")
+        add_frameworks("OpenGLES", "CoreGraphics", "Metal", "QuartzCore", "UIKit", "VideoToolbox", "CoreMedia", "CoreVideo")
     elseif is_plat("linux") then
         add_deps("libx11")
         add_syslinks("GL", "pthread", "dl")
     end
 
-    add_deps("genie")
+    add_deps("genie 1204")
```
luadebug: I think we can keep it as is "genie" it should acquire newest?

packages/b/bgfx/xmake.lua
```diff
@@ -72,15 +77,15 @@ package("bgfx")
 
             local configs = {}
             table.insert(configs, "/p:Configuration=" .. mode)
-            table.insert(configs, "/p:Platform=" .. (package:is_arch("x64") and "x64" or "Win32"))
-            table.insert(configs, "bgfx.sln")
+            table.insert(configs, "/p:Platform=" .. (package:is_arch("x64") and "x64" or (package:is_arch("arm64") and "ARM64" or (package:is_arch("arm") and "ARM" or "Win32"))))
+            table.insert(configs, tonumber(msvc:config("vs")) >= 2026 and "bgfx.slnx" or "bgfx.sln")
```
luadebug: Is it going to work with old versions of bgfx? As I'm not sure old ones provide **.slnx**

packages/b/bgfx/xmake.lua
```diff
@@ -111,7 +117,21 @@ package("bgfx")
             local envs = make.buildenvs(package)
             envs.BX_DIR = bxdir
             envs.BIMG_DIR = bimgdir
+            
+            if package:is_plat("windows") and package:is_arch("arm64") then
+                os.rm("3rdparty/glsl-optimizer/include/c99/stdint.h")
```
luadebug: Maybe it's better to place **os.tryrm**

packages/b/bgfx/xmake.lua
FFFelian: 旧版本bgfx也正常兼容的。bx的scripts/toolchain.lua用Win32_/win64_/arm64_拼接action名，至少自从7816版本开始便已经这种做法了。

luadebug (changes requested):
Good. But we need place on_check("linux") so this would check if version equal to newest and arm cpu build so it would assert that newest version of bgfx does not support Linux arm.

luadebug (approved):

waruqi:
linux arm fails, you can fix or ignore it in on_install

PR #10596, opened 2026-08-09
imgui: update to v1.92.9b and add several new options

Northn:
- Added ImGui v1.92.8, v1.92.9, and v1.92.9b.
- Added the corresponding docking variants.
- Added the following optional configurations, all disabled by default:
  - `no_demo_windows`
  - `no_debug_tools`
  - `no_obsolete_functions`
  - `builtin_math_operations`

### DIFF
--- a/packages/i/imgui/xmake.lua
+++ b/packages/i/imgui/xmake.lua
@@ -7,6 +7,9 @@ package("imgui")
     add_urls("https://github.com/ocornut/imgui.git", {alias = "git"})
 
     -- don't forget to add the docking versions as well
+    add_versions("v1.92.9b", "21d8a0a565e85dce943e375db00812c2f3f0ab21f3f0f7964e364a63422d7f99")
+    add_versions("v1.92.9", "af97ed649182c39314320514a672b82008ab462b9293fe23d37b30bfa5d05519")
+    add_versions("v1.92.8", "fecb33d33930e12ff53a34064e9d3a06c8f7c3e04408f14cd36c80e3faac863b")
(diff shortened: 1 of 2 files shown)


SirLynix (approved):

packages/i/imgui/xmake.lua
```diff
@@ -7,6 +7,9 @@ package("imgui")
     add_urls("https://github.com/ocornut/imgui.git", {alias = "git"})
 
     -- don't forget to add the docking versions as well
+    add_versions("v1.92.9b", "21d8a0a565e85dce943e375db00812c2f3f0ab21f3f0f7964e364a63422d7f99")
```
waruqi, line 10: this is not a valid semver format; xmake will only parse it as 1.92.9

https://github.com/xmake-io/xmake/issues/7748#issuecomment-5512273660

packages/i/imgui/xmake.lua
```diff
@@ -59,6 +62,9 @@ package("imgui")
     add_versions("v1.76",   "e482dda81330d38c87bd81597cacaa89f05e20ed2c4c4a93a64322e97565f6dc")
     add_versions("v1.75",   "1023227fae4cf9c8032f56afcaea8902e9bfaad6d9094d6e48fb8f3903c7b866")
 
+    add_versions("v1.92.9b-docking", "90ded916bd57db2e0e171b6b098940a47c6f5042725dcdc67fb19940ca8bfdcc")
```
waruqi, line 65: and here

PR #10598, opened 2026-08-10
lief: add 1.0.0 and fix on_test link failure

luiox:
LIEF 1.0.0 was released on 2026-07-12 but the autoupdate bot PR #10461 fails CI on x86_64 / i686 platforms. This PR adds 1.0.0 and the two missing 0.17.x point releases (0.17.5, 0.17.4), and fixes the underlying CI failures + the pre-existing mbedtls/arm inconsistencies.

## Root cause

The autoupdate PR only adds `add_versions("1.0.0", ...)`. CI then fails during `on_test`, not `on_install`:

```
> checking for c snippet(find_package/spdlog)
checkinfo: /usr/bin/ld: cannot find -lspdlog: No such file or directory
...
> checking for c snippet(find_package/nlohmann_json)
checkinfo: /usr/bin/ld: cannot find -lnlohmann_json: No such file or directory
```

`package:has_cfuncs()` calls `fetch_librarydeps()` which propagates the links of **all** library deps into the test link command. The header-only / build-time deps (`nlohmann_json`, `tl_expected`, `utfcpp`, `tcb-span`, `frozen`) have no real `.lib`, so xmake falls back to a bogus `-l<name>` and the link step fails — even though the actual cmake build succeeds. `spdlog` is built `header_only=false`, so its `libspdlog` legitimately links.

Note: all ARM/android jobs on #10461 show SUCCESS only because the old `on_check` `raise()` on arm skipped the package entirely — they never reached the build/test step.

## Changes

- **add 1.0.0** (sha256 of the release source tarball).
- **add 0.17.5 and 0.17.4** — the two missing 0.17.x point releases.
- **header-only deps → `{private = true}`**: `nlohmann_json`, `tl_expected`, `utfcpp`, `tcb-span`, `frozen`. Their links are no longer propagated to `on_test` (or downstream consumers), matching the pattern used by `ktx` and `nzsl` (which has the same `frozen` + `nlohmann_json` deps). `spdlog` and `mbedtls` stay public since `libLIEF` genuinely links them.
- **mbedtls version-gated** (per review): 1.0.0 needs `mbedtls >=4.0.0` (`find_package(MbedTLS 4.0.0)` + `MbedTLS::tfpsacrypto`). Pre-1.0.0 actually supports the whole **3.6.x LTS** line (3.6.0 .. 3.6.7), so the constraint is now `>=3.6.0 <3.7.0` instead of the old `<3.6.0`. Resolved via `on_load` so all version lines keep working.
- **drop the arm guard** (per review): the `on_check` `raise()` on arm had no real basis — mbedtls supports arm — and only existed to paper over the old `<3.6.0` constraint. Now that pre-1.0.0 uses the 3.6.x LTS line and 1.0.0 uses 4.x, both of which support arm, the guard is removed so arm builds are no longer skipped. (mbedtls 4.x prerequisite from #10590 is merged.)
- **utf8cpp rewrite skipped for `>= 1.0.0`**: 1.0.0 already links `utf8cpp::utf8cpp` natively, so the `io.replace` is guarded by a version check. The two `lief_spdlog` rewrites are kept since 1.0.0 still inlines `lief_spdlog`.

## Local validation

Loaded the package from the local repo with xmake:

- `add_requires("lief 1.0.0")` → deps resolve to `mbedtls >=4.0.0` ✓
- `add_requires("lief 0.17.6")` / `0.17.5` / `0.17.4` → deps resolve to `mbedtls >=3.6.0 <3.7.0` ✓

Refs #10461.

## Portability fixes for the remaining CI platforms (5a1fe5e6)

- **android r22 (6 jobs)**: NDK r22 ships clang 11.0.5, which CMake rejects `cxx_std_23` for (`target_compile_features The compiler feature "cxx_std_23" is not known to CXX compiler ... version 11.0.5`, from `src/runtime/CMakeLists.txt:22`). With `LIEF_RUNTIME=OFF` (the default) only the C++17 runtime stubs + `osx|windows/common.cpp` are compiled there (no C++20/23 constructs — `LIB_LIEF` itself is `CXX_STANDARD 17`), so `runtime-cxx-std.patch` requires `cxx_std_23` only when `LIEF_RUNTIME` is enabled, and `cxx_std_17` otherwise. r27 (clang 18) keeps working.
- **wasm shared (1 job)**: `wasm-ld: error: unknown argument: --exclude-libs` — LIEF adds `-Wl,--exclude-libs,ALL` for `UNIX AND NOT APPLE`, which is also true under emscripten. `emscripten-link-options.patch` keeps only `-Wl,--gc-sections` (supported by wasm-ld) when `EMSCRIPTEN` is set.
- **freebsd (1 job + fail-fast cancel)**: FreeBSD's `sys/types.h` defines `major(d)`/`minor(d)` macros which mangle `version_t`'s `major(major)`/`minor(minor)` member initializers in `include/LIEF/runtime/{osx,windows}/Host.hpp`. `freebsd-major-minor.patch` `#undef`s both (guarded by `__FreeBSD__`) in those headers.

**Update**: two follow-up fixes after running CI on a fork (#1 on luiox/xmake-repo, all 17 workflows green at 4fe52fe7):

- the same `-Wl,--exclude-libs,ALL` link flags and the POST_BUILD `${CMAKE_STRIP} --strip-all` rule also break under emscripten for the shared build (`wasm-ld: unknown argument: --exclude-libs`; host strip reports `file format not recognized` on the wasm `libLIEF.so`). Both are now skipped when `EMSCRIPTEN` is set — for 1.0.0 (`target_link_options`) and for the newly added 0.17.5/0.17.4 (same flags set via `LINK_FLAGS`).


packages/l/lief/xmake.lua
```diff
@@ -46,8 +49,19 @@ package("lief")
 
     add_deps("cmake")
     add_deps("spdlog", {configs = {header_only = false, noexcept = true}})
-    add_deps("nlohmann_json", {configs = {cmake = true}})
-    add_deps("tl_expected", "utfcpp", "mbedtls <3.6.0", "tcb-span", "frozen")
+    add_deps("nlohmann_json", {configs = {cmake = true}, private = true})
+    add_deps("tl_expected", "utfcpp", "tcb-span", "frozen", {private = true})
+
+    -- mbedtls: 1.0.0 requires mbedtls >= 4.0.0 (PSA Crypto / tf-psa-crypto target);
+    -- older versions still need mbedtls < 3.6.0 (mbedcrypto target). The latter does
```
Northn: previous versions actually support all 3.6.* versions (3.6.4 offially listed); and the latest mbedtls version is 3.6.7. better to set >= 3.6.0 <3.7.0 (at least for 0.17.x builds)

packages/l/lief/xmake.lua
```diff
@@ -56,10 +70,10 @@ package("lief")
                 if vs_toolset then
                     local vs_toolset_ver = import("core.base.semver").new(vs_toolset)
                     local minor = vs_toolset_ver:minor()
-                    assert(minor and minor >= 30, "package(lief) require vs_toolset >= 14.3")
+                    assert(minor and minor >= 30, "package(lief) requires vs_toolset >= 14.3")
                 end
             end
-            if package:is_arch("arm.*") then
+            if package:is_arch("arm.*") and package:version() and package:version():lt("1.0.0") then
                 raise("package(lief) dep(mbedtls <3.6.0) unsupported arm arch")
```
Northn: please take a look at this also: https://github.com/xmake-io/xmake-repo/pull/10461#issuecomment-5216761519

packages/l/lief/xmake.lua
luiox: Good catch, thanks. Widened the pre-1.0.0 constraint to `mbedtls >=3.6.0 <3.7.0` (the 3.6.x LTS line, up to 3.6.7) in da6eacf7 — verified locally that 0.17.6/0.17.5/0.17.4 now resolve to `mbedtls >=3.6.0 <3.7.0` while 1.0.0 still resolves to `>=4.0.0`.

packages/l/lief/xmake.lua
luiox: Agreed — the arm guard had no real basis since mbedtls supports arm. Dropped the `raise()` entirely in da6eacf7, so arm builds are no longer skipped. (The mbedtls 4.x prerequisite from #10590 is already merged.)

PR #10638, opened 2026-08-17
Add mdspan package

homeomorphisms:
Add kokkos/mdspan.

### DIFF
--- a/packages/m/mdspan/xmake.lua
+++ b/packages/m/mdspan/xmake.lua
@@ -1,0 +1,38 @@
+package("mdspan")
+    set_kind("library", {headeronly = true})
+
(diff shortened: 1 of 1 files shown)


luadebug:
It seems github rate limits when it tried to handle just six package tags at once. I don't know if curl that was before aria as download tool was either good for this case. Maybe sending headers of curl to aria tool...

homeomorphisms:
I believe these failures were caused by the GitHub outage on Aug 17. Maybe re-run the failed jobs? Test on my local machine looks good.

luadebug:
OK. It is working now. You can drop windows if on_load defines doesn't help.

PR #10701, opened 2026-08-26
sqlgen: fix on_check

Persson-dev:
Changes :
- Add C++20 check
- Allow build for linux arm

### DIFF
--- a/packages/s/sqlgen/xmake.lua
+++ b/packages/s/sqlgen/xmake.lua
@@ -39,13 +39,25 @@ package("sqlgen")
     end)
 
     on_check(function (package)
-        if package:config("postgres") then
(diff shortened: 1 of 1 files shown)


packages/s/sqlgen/xmake.lua
```diff
@@ -39,13 +39,25 @@ package("sqlgen")
     end)
 
     on_check(function (package)
-        if package:config("postgres") then
+        if package:config("postgres") and is_plat("windows") then
```
waruqi: package:is_plat

Persson-dev: oops, my bad

Persson-dev: should be fixed now

PR #10729, opened 2026-08-31
blackbone: add package

DesolateVE:
Blackbone does not currently have an xmake-repo package. This adds its Windows user-mode static library so projects can use `add_requires("blackbone")` and `add_packages("blackbone")`.

Following the review, the recipe removes `src/3rd_party` before building and uses repository dependencies instead of compiling or installing the bundled libraries. The upstream `DIA` binary directory is also removed.

Changes:
- Pin Blackbone `2023.07.17` to upstream commit `5ede6ce50cd8ad34178bfa6cae05768ff6b3859b`; support Windows x86/x64 with MSVC.
- Add AsmJit `2014.12.01` (`48da90ded775fa2ba0fd3f15522890ad631ad6de`) to the existing package. This matches Blackbone's old `X86Assembler` API. Enable both x86/x64 code generators, accommodate CMake 4, and test this API separately from newer versions. Blackbone pins this exact version rather than assuming compatibility across other old snapshots.
- Use the existing `rewolf-wow64ext 2022.09.26` package on x86. Enable static builds using the consumer's CRT and initialization without `DllMain`, propagate the static header define, fix the SDK `_CLIENT_ID` collision, and retain the LGPL license text. The package now uses the normal static default; explicit `shared=true` remains supported. Shared resources use the SDK's `winres.h` instead of requiring MFC's `afxres.h`.
- Exclude the WOW64-only subsystem on native x64. On x86, use WOW64Ext's public `GetModuleHandle64` instead of the bundled fork's `getNTDLL64` helper.
- Use system `diasdk` for headers and GUID linking. Fix its include directory and add x86 lookup, including compatibility with older xmake SDK detection. Do not redistribute Microsoft's GUID library or runtime DLLs.
- Use the existing `beaengine` package for TraceHook so its bundled headers can also be removed.
- Relocate Blackbone's own `VersionApi.h` into `BlackBone/Include`; it implements Blackbone-specific version helpers and is not an external library.
- Install public headers, export `BLACKBONE_STATIC`, propagate dependency/system links, and retain the LDasm license notice. No driver is built or installed; only the driver-definition header required by user-mode headers is installed.

Validation on Windows 11 x64, MSVC 14.51, xmake 3.0.9:
- The official package-test entry point passed all eight Blackbone combinations after debundling: x86/x64, Release/Debug, MT/MD (MTd/MDd in Debug).
- Separate x86 and x64 consumers passed in Release and Debug using installed packages: read a test DLL into memory, map it into the current process without CreateLdrRef, resolve/call `add(12, 30)`, verify 42, check absence from the native loader list, and unmap.
- Consumer import-table checks show no BlackBone, AsmJit, WOW64Ext, or test-DLL imports.
- WOW64Ext `2022.09.26` and `v1.0.0+9` each passed static/shared x86 Release tests, including a consumer that includes `winternl.h` before `wow64ext.h`.
- AsmJit `2014.12.01` additionally passed an independent x64 shared build/link test.
- `git diff --check` passed.

Example:
```powershell
xmake l scripts/test_packages.lua --shallow -vD -p windows -a x64 -m release -k static --runtimes=MT blackbone
```

The unified `scripts/test.lua` imports `core.package.addon`, which is unavailable in my installed xmake 3.0.9. I used its official package-test entry point directly without modifying repository test scripts. Other operating systems/toolchains have not been tested locally.

Building Blackbone requires the MSVC C++ ATL component and DIA SDK. Consumers must provide the DIA runtime and, when needed for symbol-server access, SymSrv from their Microsoft SDK installation. Runtime initialization can access Microsoft's symbol server; the package's `on_test` only compiles and links, without process attachment or symbol downloads.


luadebug:
Why not to debundle asmjit. Just add_version for asmjit package (`packages/a/asmjit/xmake.lua`) and later on use it for blackbone package by specifying asmjit package version `"asmjit <=x.y.z")`?
Why not use existing package for `rewolf-wow64ext` (`packages/r/rewolf-wow64ext/xmake.lua`)?
It would be perfect to just `os.rm("src/3rd_party"` and use `diasdk` package (`packages/d/diasdk/xmake.lua`) so everything would get de-bundled.

luadebug:
Diasdk maybe can be resolved via `package:targetarch("...")` or `package:is_targetarch("...")`?
The windows runner that fails looks like x64 windows that runs something like this `xmake l scripts/test.lua -vD -a arm64 --shallow diasdk`

packages/r/rewolf-wow64ext/xmake.lua
```diff
@@ -55,12 +88,13 @@ package("rewolf-wow64ext")
                 end
         ]])
         import("package.tools.xmake").install(package)
+        os.cp("lgpl-3.0.txt", package:installdir("licenses"))
```
luadebug, line 91: I am unsure that any xmake package ever installs license file.

waruqi: xmake doesn't standardize the license installation directory, and it's not actually used. You only need to specify `set_license`.

packages/b/blackbone/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+        io.replace("src/BlackBone/Process/ProcessCore.cpp", "if (wowSrc == TRUE)", "#ifdef USE32\n        if (wowSrc == TRUE)", {plain = true})
+        io.replace("src/BlackBone/Process/ProcessCore.cpp", "else\n            _native = std::make_unique<Native>",
+                   "else\n#endif\n            _native = std::make_unique<Native>", {plain = true})
+        os.rm("src/3rd_party")
+        os.rm("DIA")
+        os.cp(path.join(package:scriptdir(), "port", "xmake.lua"), "xmake.lua")
+        import("package.tools.xmake").install(package)
+
+        -- LDasm is part of the Blackbone sources and retains its own license.
+        os.cp("src/BlackBone/Asm/LDasm.c", package:installdir("licenses"))
```
luadebug: Same thing

packages/b/blackbone/xmake.lua
```diff
@@ -0,0 +1,56 @@
...
+                io.replace(file, "<3rd_party/DIA/dia2.h>", "<dia2.h>", {plain = true})
+                io.replace(file, "<3rd_party/BeaEngine/headers/BeaEngine.h>", "<beaengine/BeaEngine.h>", {plain = true})
+                io.replace(file, "<3rd_party/rewolf-wow64ext/src/wow64ext.h>", "<wow64ext.h>", {plain = true})
+            end
+        end
+        io.replace("src/BlackBone/Subsystem/Wow64Subsystem.cpp", "getNTDLL64()", 'GetModuleHandle64(L"ntdll.dll")', {plain = true})
+        -- Native x64 processes never use the WOW64-only subsystem.
+        io.replace("src/BlackBone/Process/ProcessCore.cpp", "if (wowSrc == TRUE)", "#ifdef USE32\n        if (wowSrc == TRUE)", {plain = true})
+        io.replace("src/BlackBone/Process/ProcessCore.cpp", "else\n            _native = std::make_unique<Native>",
+                   "else\n#endif\n            _native = std::make_unique<Native>", {plain = true})
```
luadebug: Should this part be sent as PR to blackbone repo itself?

PR #10762, opened 2026-09-06
fix dead download urls (giflib, iso-codes, ncurses, gmp, cppunit, itstool, mpich)

gistrec:
Now that repology tracks xrepo (https://repology.org/repository/xrepo/problems), it flagged several dead/redirecting download URLs. All checked live today; checksums verified unchanged where the URL changed.

- **giflib**: SourceForge moved files into subdirectories — the flat `/project/giflib/giflib-X.tar.gz` path now 404s for both 5.2.1 and 5.2.2. Switched to `/project/giflib/giflib-5.x/...`; sha256 of both tarballs verified identical. Note: the xmake-mirror fallback (https://github.com/xmake-mirror/giflib/releases) only has a 5.2.1 asset, so before this fix **giflib 5.2.2 was not installable at all** — uploading giflib-5.2.2.tar.gz there would be a nice extra safety net.
- **iso-codes**: `deb.debian.org` pool only keeps current versions — 4.16.0 (and 4.8.0) orig tarballs are gone, so the latest listed version was not installable. Added 4.20.1 (current in pool, sha256 computed from the downloaded tarball).
- **ncurses, gmp**: dropped the `ftpmirror.gnu.org` URL — the GNU mirror redirector currently returns 502 for every request (confirmed from two vantage points); `ftp.gnu.org` and the existing mirrors remain.
- **cppunit, itstool, mpich**: http → https (servers permanently redirect; avoids one hop and clears repology's permanent-redirect warnings).


gistrec:
Pushed fixes for the CI failures (00a3685):

- **iso-codes**: 4.20.1 turned out to be meson-only (upstream dropped autotools), which broke the package's `autoconf.install`. Replaced with **4.18.0** — the newest autotools-based version still in the Debian pool (sha256 computed from the downloaded tarball).
- **mpich**: macOS runners got 403 from www.mpich.org (bot protection on GET). Added the official GitHub release asset (https://github.com/pmodels/mpich/releases) as the primary URL — sha256 verified identical — keeping mpich.org as fallback.
- **giflib**: put the (fixed) SourceForge URL first — the xmake-mirror release only has a 5.2.1 asset, so for 5.2.2 it always 404s before falling back.
- **cppunit**: restricted `on_install` to `!wasm` — its `DllPlugInTester` fails to link under emscripten (undefined libc++ symbols), so the package was never installable on wasm anyway.

Two remaining failures are **pre-existing and unrelated to the URL changes** (they show up because touching a package makes CI rebuild it):
- **ncurses 6.6 on MinGW**: `curses.priv.h:2300: conflicting types for '_nc_wctomb'` — upstream 6.6 vs mingw-w64 headers issue; the downloaded tarball hash is unchanged and correct.
- The earlier MinGW "unmatched checksum" lines were truncated downloads from invisible-mirror.net (the file itself is correct — verified sha256 matches from a clean download); the GNU fallback then succeeded.

waruqi:
@copilot resolve the merge conflicts in this pull request

waruqi:
Modifying too many packages at once makes it difficult to pass CI tests. It's recommended to do one patch per package.

gistrec:
Resolved the conflicts by merging dev — and since @luadebug kindly upstreamed the mpich (#10765) and ncurses (#10764) parts with proper build fixes on top, this PR is now down to five packages: giflib, iso-codes, gmp, cppunit, itstool. Happy to split further if you'd still prefer one package per PR.

PR #10763, opened 2026-09-06
update homepage urls flagged by repology (https, canonical locations)

gistrec:
Repology flagged these homepages as permanent redirects (https://repology.org/repository/xrepo/problems). This updates 44 packages' `set_homepage` to the final destination — checked live today, every target returns HTTP 200.

Most are plain http → https on the same host. Four moved further along the redirect chain to their canonical location:

- **udis86 / libudis86**: udis86.sourceforge.net → https://sourceforge.net/projects/udis86/ (the old project subdomain now redirects there)
- **wineditline**: mingweditline.sourceforge.net → https://sourceforge.net/projects/mingweditline/
- **usd**: www.openusd.org → https://openusd.org/
- **websocketpp**: www.zaphoyd.com/websocketpp → https://www.zaphoyd.com/projects/websocketpp/

Note: https://www.mysql.com/ returns 403 to plain curl (bot protection) but is fine in a browser — kept as-is.

No functional changes — `set_homepage` only, one line per package.


waruqi:
@copilot resolve the merge conflicts in this pull request

gistrec:
Fair point, thanks — closing this in favour of small batches. Batch 1 (#10775, nine packages that all built green in this PR's CI) is open now; I'll send the rest as ~9-package batches one at a time as each lands, keeping packages with pre-existing build issues (e.g. e2fsprogs on macOS, pqp vs modern GCC, the heavy mysql/usd downloads) isolated in their own batches so they don't block the clean ones. Shout if you'd prefer strictly one package per PR.

PR #10775, opened 2026-09-07
update homepage urls flagged by repology (batch 1: avisynthplus, bento4, cello, ck, libraw, libuv, lua, mujs, pcg-cpp)

gistrec:
Follow-up to #10763, split into small batches as requested there. This is batch 1 of ~5: nine packages, one `set_homepage` line each — http → https, targets verified live (HTTP 200). All nine packages built successfully in the CI runs of #10763.

Next batches will follow one at a time as the previous one lands, keeping known-problematic packages (pre-existing build failures like e2fsprogs on macOS or pqp with modern GCC) isolated in their own batches.

### DIFF
--- a/packages/a/avisynthplus/xmake.lua
+++ b/packages/a/avisynthplus/xmake.lua
@@ -1,5 +1,5 @@
 package("avisynthplus")
-    set_homepage("http://avs-plus.net")
+    set_homepage("https://avs-plus.net/")
     set_description("AviSynth with improvements")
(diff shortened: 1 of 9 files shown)


luadebug:
Thanks for a batch. Feel free to fetch fixes for pcg-cpp.

gistrec:
Done — merged dev to pick up your #10776, thanks! The arm64 jobs should be green on the rerun.

luadebug (approved):

PR #10808, opened 2026-09-09
Update imguizmo to 1.91.3+wip.0110 to adapt to imgui >=1.92.8 ImDrawL…

### LINKED ISSUE #10807: [Update Package]: imguizmo
### Action Type

Update Package

### Package Name

imguizmo

### Version(s)

1.9,1.10

### Project Repository

https://github.com/CedricGuillemet/ImGuizmo

### Project Homepage

_No response_

### License

MIT

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [x] Android
- [x] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD

The upstream version uses CMake as build system

### Additional Context

- From imgui version 1.92.8 onwards, the order of the parameters in `ImDrawList` was swapped, causing imguizmo to report an error following an imgui upgrade
- From version 1.9 of https://github.com/CedricGuillemet/ImGuizmo, the issue with the swapped ImDrawList parameters was fixed; however, the repository file structure was also modified, with both header and cpp files now located in the `src` directory.

gsw945:
fix #10807

### DIFF
--- a/packages/i/imguizmo/xmake.lua
+++ b/packages/i/imguizmo/xmake.lua
@@ -4,13 +4,23 @@ package("imguizmo")
     set_license("MIT")
 
     add_urls("https://github.com/CedricGuillemet/ImGuizmo.git")
-    add_versions("1.83", "664cf2d73864a36b2a8b5091d33fc4578c885eca")
-    add_versions("1.89+wip", "82e2465b8d029e2d85002905cc4ed5087e2119fe")
+
+    add_versions("1.91.3+wip.0110", "b796ac3b861afc6e91ca74e4611effd9c9527367") -- v1.10
+    add_versions("1.91.3+wip.0109", "9e437c6ce80453bd45a5674ac2d81ee754cee95c") -- v1.9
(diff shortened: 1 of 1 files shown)


SirLynix:
I think we should follow ImGuizmo versions. Remove 1.83/1.91 and replace them by 1.9 and 1.10.

That means people with existing ImGuizmo packages won't have it automatically upgraded though (since semver 1.83>=1.10).

What do you think @waruqi?

waruqi:
> I think we should follow ImGuizmo versions. Remove 1.83/1.91 and replace them by 1.9 and 1.10.
> 
> That means people with existing ImGuizmo packages won't have it automatically upgraded though (since semver 1.83>=1.10).
> 
> What do you think @waruqi?

That is possible; this is a case of version breakage caused by the upstream source, and we currently cannot ensure good compatibility. Perhaps we could add some warnings in `on_load`.

SirLynix:
Can we detect an existing imguizmo installation in version 1.89/1.91 and issue a warning from the package itself?

waruqi:
> Can we detect an existing imguizmo installation in version 1.89/1.91 and issue a warning from the package itself?

package:version() in on_load?

PR #10830, opened 2026-09-10
Add raptor2, lrdf

### LINKED ISSUE #9850: [New Package]: lrdf
### Action Type

New Package

### Package Name

lrdf

### Version(s)

latest

### Project Repository

https://github.com/swh/LRDF

### Project Homepage

_No response_

### License

GPL-2.0

### Supported Platforms

- [x] Windows
- [x] Linux
- [x] macOS
- [ ] Android
- [ ] iOS
- [ ] WebAssembly (Wasm)
- [ ] FreeBSD

Upstream uses Autoconf.
Dependency: raptor2.

### Additional Context

LRDF is stable software that works reliably since 15 years, and doesn't seem to need a new commit.

EZ4Stephen:
Adds raptor2 2.0.16 and lrdf 0.6.1. Makes use of custom xmake build scripts, patches both to build on windows.
The .git link for raptor is https://github.com/dajobe/raptor , but would've required additional build tools, so it wasn't added.

Closes #9850 .

### DIFF
--- a/packages/l/lrdf/port/xmake.lua
+++ b/packages/l/lrdf/port/xmake.lua
@@ -16,7 +16,7 @@ target("lrdf")
     end
 
     if is_plat("mingw") and is_kind("shared") then
-        add_cxflags("-fvisibility=default")
+        add_shflags("-Wl,--export-all-symbols")
     end
 
     before_build(function(target)
--- a/packages/l/lrdf/port/xmake.lua
+++ b/packages/l/lrdf/port/xmake.lua
@@ -4,6 +4,10 @@ add_requires("raptor2")
 target("lrdf")
     set_kind("$(kind)")
 
+    if not is_plat("windows") then
+        add_defines("HAVE_OPENSSL")
+    end
+
     if is_plat("windows") then
         add_cflags("/D_CRT_SECURE_NO_WARNINGS")
         if is_kind("shared") then
@@ -92,7 +96,7 @@ target("lrdf")
         end
     end)
 
-    add_files("src/lrdf.c", "src/lrdf_multi.c")
+    add_files("src/lrdf.c", "src/lrdf_multi.c", "src/md5.c")
     add_includedirs(".", "src")
     add_headerfiles("lrdf.h", "lrdf_types.h")
 
--- a/packages/l/lrdf/port/xmake.lua
+++ b/packages/l/lrdf/port/xmake.lua
@@ -9,8 +9,8 @@ target("lrdf")
     end
 
     if is_plat("windows", "mingw") then
-        add_cflags("/D_CRT_SECURE_NO_WARNINGS")
-        if is_kind("shared") then
+        add_defines("_CRT_SECURE_NO_WARNINGS")
+        if is_plat("windows") and is_kind("shared") then
             add_rules("utils.symbols.export_all")
         end
     end
@@ -20,8 +20,8 @@ target("lrdf")
     end
 
     before_build(function(target)
-        local srctypes = path.join("lrdf_types.h")
-        local srcmain = path.join("lrdf.h")
+        local srctypes = "lrdf_types.h"
+        local srcmain = "lrdf.h"
         local srclrdf = path.join("src", "lrdf.c")
         local srclmulti = path.join("src", "lrdf_multi.c")
 
--- a/packages/l/lrdf/port/xmake.lua
+++ b/packages/l/lrdf/port/xmake.lua
@@ -4,7 +4,7 @@ add_requires("raptor2")
 target("lrdf")
     set_kind("$(kind)")
 
-    if not is_plat("windows", "mingw") then
+    if not is_plat("windows", "mingw", "wasm") then
         add_defines("HAVE_OPENSSL")
     end
 
@@ -13,6 +13,10 @@ target("lrdf")
         if is_kind("shared") then
             add_rules("utils.symbols.export_all")
         end
+    end
+
+    if is_plat("mingw") and is_kind("shared") then
+        add_cxflags("-fvisibility=default")
     end
 
     before_build(function(target)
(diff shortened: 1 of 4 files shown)


EZ4Stephen:
I am a little surprised that mingw got an error of an undefined symbol despite add_rules("utils.symbols.export_all") in lrdf. I'll assume that's intentional.

EZ4Stephen:
I'd appreciate any help on fixing the mingw errors.

EZ4Stephen:
Looks like this is good now. One check could be rerun.

luadebug (approved):
Thanks.

packages/l/lrdf/port/xmake.lua
```diff
@@ -0,0 +1,107 @@
...
+
+target("lrdf")
+    set_kind("$(kind)")
+
+    if not is_plat("windows", "mingw", "wasm") then
+        add_defines("HAVE_OPENSSL")
+    end
+
+    if is_plat("windows", "mingw") then
+        add_cflags("/D_CRT_SECURE_NO_WARNINGS")
```
luadebug: Why not `add_defines`?
```lua
option("tests", {default = false, description = "Enable Tests"})

target("foo")
    set_kind("binary")
    add_files("src/*.cpp")
    if has_config("tests") then
        add_defines("TESTS")
    end
```

packages/l/lrdf/port/xmake.lua
```diff
@@ -0,0 +1,107 @@
...
+        if is_kind("shared") then
+            add_rules("utils.symbols.export_all")
+        end
```
luadebug: This usually wrapped only over `windows` but not for `mingw`?
```lua
    if is_plat("windows") and is_kind("shared") then
        add_rules("utils.symbols.export_all")
    end
```

Besides `mingw` seems down below already exporting them via `add_shflags("-Wl,--export-all-symbols")`?

packages/l/lrdf/port/xmake.lua
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    if is_plat("mingw") and is_kind("shared") then
+        add_shflags("-Wl,--export-all-symbols")
+    end
+
+    before_build(function(target)
+        local srctypes = path.join("lrdf_types.h")
+        local srcmain = path.join("lrdf.h")
```
luadebug: Do we really need do `path.join` over one arg?

packages/r/raptor2/port/xmake.lua
```diff
@@ -0,0 +1,163 @@
...
+add_requires("libxml2", "libcurl")
+if has_config("libxslt") then
+    add_requires("libxslt")
+end
+
+set_configvar("RAPTOR_VERSION_DECIMAL", 20016)
+set_configvar("RAPTOR_VERSION_MAJOR", 2)
+set_configvar("RAPTOR_VERSION_MINOR", 0)
+set_configvar("RAPTOR_VERSION_RELEASE", 16)
+set_configvar("VERSION", "2.0.16")
```
luadebug, line 14: Maybe good to use something like shown in `angelscript` at this case.
```lua
option("vers", {description = "Set the version"})

target("angelscript")
    set_version("$(vers)")
```

luadebug (commented):
Very good.

packages/r/raptor2/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+
+    add_urls("https://download.librdf.org/source/raptor2-$(version).tar.gz")
+
+    add_versions("2.0.16", "089db78d7ac982354bdbf39d973baf09581e6904ac4c92a98c5caadb3de44680")
+
+    add_configs("libxslt", {description = "Enable GRDDL parser", default = true, type = "boolean"})
+
+    add_deps("libxml2", "libcurl")
+    add_links("raptor2")
+    add_includedirs("include/raptor2", {public = true})
```
waruqi: remove `public = true`, it's ununsed

packages/r/raptor2/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+    set_license("LGPL-2.1-or-later")
+
+    add_urls("https://download.librdf.org/source/raptor2-$(version).tar.gz")
+
+    add_versions("2.0.16", "089db78d7ac982354bdbf39d973baf09581e6904ac4c92a98c5caadb3de44680")
+
+    add_configs("libxslt", {description = "Enable GRDDL parser", default = true, type = "boolean"})
+
+    add_deps("libxml2", "libcurl")
+    add_links("raptor2")
```
waruqi: remove it

packages/r/raptor2/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+            package:add("deps", "libxslt")
+        end
+        if not package:config("shared") then
+            package:add("defines", "RAPTOR_STATIC", {public = true})
+        end
+    end)
+
+    on_install("!iphoneos", function(package)
+        local configs = {}
+        configs.kind = package:config("shared") and "shared" or "static"
```
waruqi: remove it, xmake will do it.

packages/l/lrdf/xmake.lua
```diff
@@ -0,0 +1,22 @@
...
+    add_urls("https://github.com/swh/LRDF/archive/refs/tags/v$(version).tar.gz",
+             "https://github.com/swh/LRDF.git")
+
+    add_versions("0.6.1", "d579417c477ac3635844cd1b94f273ee2529a8c3b6b21f9b09d15f462b89b1ef")
+
+    add_deps("raptor2")
+
+    on_install("!iphoneos", function(package)
+        local configs = {}
+        configs.kind = package:config("shared") and "shared" or "static"
```
waruqi: and here

packages/r/raptor2/port/xmake.lua
EZ4Stephen: I think that's different; that's for the package version but the cmake template needed this instead.

EZ4Stephen:
Is this good now?

PR #10837, opened 2026-09-12
update homepage urls flagged by repology (batch 2: itstool, krb5, libyaml, mpc, rply, soci, websocketpp, wren)

gistrec:
Batch 2 of the homepage cleanup started in #10775 (split from #10763 as requested). Eight packages, one `set_homepage` line each — http → https, targets verified live. Two notes:

- **websocketpp**: also follows the moved path — https://www.zaphoyd.com/projects/websocketpp/
- **wren**: adds the canonical trailing slash — https://wren.io/

libudis86 from the original list is gone — already handled by removing the duplicate package in #10817. Remaining after this batch: the riskier builds (open3d, qjson, mysql-build-tools, strawberry-perl, viennacl, wineditline, libpfm) in a final batch.

### DIFF
--- a/packages/i/itstool/xmake.lua
+++ b/packages/i/itstool/xmake.lua
@@ -1,6 +1,6 @@
 package("itstool")
     set_kind("binary")
-    set_homepage("http://itstool.org/")
+    set_homepage("https://itstool.org/")
     set_description("ITS Tool allows you to translate your XML documents with PO files")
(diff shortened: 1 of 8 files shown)


luadebug (approved):

luadebug:
https://repology.org/repository/xrepo/problems Thanks this should get us close to 98 problems remaining.

PR #10842, opened 2026-09-13
fix urls flagged by repology (alsa-lib, libspectre, libuuid, polyclipping, rapidxml)

gistrec:
Small follow-up in the repology cleanup series (#10775, #10837). Five one-line fixes:

- **alsa-lib**, **libspectre**: download URLs http → https (servers permanently redirect; https verified live)
- **libuuid**, **polyclipping**, **rapidxml**: trailing slash on the SourceForge homepage (repology's canonical form for project pages)

Note: the `download_dead` entries repology currently shows for gnu-sed and lunzip were a temporary GNU infrastructure outage (ftpmirror.gnu.org and download.savannah.gnu.org returned 502 for a few days) — both are back up, so those will clear on their own.


luadebug:
Either change sourceforge download url to github download url or wait for https://github.com/xmake-io/xmake/pull/7758 ?
It seems reason why it was working before is that curl was used by default without aria2 useragent and multithreaded download.

luadebug:
Seems it has been merged but xmake that being acquired by github CI is old commit [2c1fd](https://github.com/xmake-io/xmake/commit/2c1fdfbea285c696350e795af7c4d45a2000acf1) I guess.
I will try re-open pr but not sure it would help acquire required commit of xmake.

luadebug:
Alright now time to fetch fix for ghostscript and PR should finally pass CI checks.

PR #10843, opened 2026-09-13
replace dead homepage urls (iowow, libisl, metis, patch, tinyxml2, szip)

gistrec:
Continuing the repology cleanup series (#10775, #10837, #10842) - these homepages are genuinely gone (checked live), replaced with the project's current home:

- **libisl**: isl.gforge.inria.fr - INRIA GForge was shut down -> https://libisl.sourceforge.io/
- **metis**: glaros.dtc.umn.edu now redirects to a 404 -> https://github.com/KarypisLab/METIS
- **tinyxml2**: grinninglizard.com/tinyxml2 404s -> https://github.com/leethomason/tinyxml2
- **iowow**: iowow.softmotions.com no longer resolves -> https://github.com/Softmotions/iowow
- **patch**: gnu.org/software/patch/patch.html now redirects to the Savannah project page -> https://savannah.gnu.org/projects/patch/
- **szip**: the old doc_resource path 404s -> archived docs at https://docs.hdfgroup.org/archive/support/doc_resource/SZIP/index.html

One set_homepage line per package, no functional changes. gmsh was dropped from this PR: #10851 already switched its homepage to https together with the gcc16 fix. A second small batch (field3d, openscenegraph, wt - heavier builds) will follow.


luadebug:
It seems `gmsh` is very large maybe better to open PR just for gmsh and if it requires some patch just patch or just go with new version?

luadebug:
Feel free to fetch gmsh fixes.

PR #10844, opened 2026-09-13
add package keynub_licdongle 1.1.1

ab-tools:
Adds `keynub_licdongle` 1.1.1, the SDK for the KeyNub USB-C license dongle (https://www.keynub.com/): the C API, a header-only C++11 wrapper and a flat companion API.

The library ships prebuilt in the upstream repository (`natives/<os>-<arch>/`), so the package installs the headers, the shared library and its import library for the requested platform and does not build anything; `shared` is therefore readonly. Supported: windows x64/x86/arm64, linux x86_64/arm64, macosx x86_64/arm64.

Tested locally with `xmake l scripts/test.lua -vD keynub_licdongle` on windows x64 (MSVC): both `licd_version` and `licdf_version` link.


ab-tools:
Hello @luadebug,

first thanks for reviewing the PR so quickly!

But I'm afraid you broke the build with your changes. ;-)

Anything else I can still support with to get the PR merged?

Best regards and thanks
Andreas

